PR #10, opened 2019-11-27
Cloudinary plugin implementation

JacoboGallardo:
Adds cloudinary plugin based on official Cloudinary Media Library widget.


plugins/cloudinary/jest.config.js
```diff
@@ -0,0 +1,22 @@
...
+  coverageThreshold: {
+    global: {
+      branches: 80,
+      functions: 80,
+      lines: 80,
+      statements: 80
+    }
+  },
+  collectCoverageFrom: ['src/**']
+}
```
JacoboGallardo: I thought it was a good idea to get the jest configuration out of the package.json file.

plugins/cloudinary/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@builder.io/plugin-cloudinary",
-  "version": "0.0.2-12",
+  "version": "0.1.0",
```
JacoboGallardo: Not sure what version convention is more convenient. I guess we can promote it to v1.0.0 if it's okay for you?

plugins/cloudinary/package.json
```diff
@@ -24,9 +24,9 @@
     "prebuild": "rimraf dist",
     "build": "tsc --module commonjs && rollup -c rollup.config.ts",
     "release:dev": "npm run build && npm version prerelease --no-git-tag-version && npm publish --tag dev",
-    "start": "SERVE=true rollup -c rollup.config.ts -w",
-    "test": "jest --coverage",
-    "test:watch": "jest --coverage --watch",
+    "start": "cross-env SERVE=true rollup -c rollup.config.ts -w",
```
JacoboGallardo: This makes sure environment variable is set regardless of the OS (windows, linux, etc...)

plugins/cloudinary/src/CloudinaryImageEditor.tsx
```diff
@@ -0,0 +1,162 @@
...
+    this.props.onChange(image)
+  }
+
+  componentDidMount() {
+    this.appendMediaLibraryScriptToPlugin()
+  }
+
+  render() {
+    const shouldRequestCloudinarySettings = this.shouldRequestCloudinaryCredentials()
+    const buttonStyle = { width: '45%', margin: '5px' }
```
JacoboGallardo: This is an arbritrary UI decision... 2 buttons with same width within the same row. One to pick up the image, one to (re)set the Cloudinary credentials.

plugins/cloudinary/src/CloudinaryMediaLibraryDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+interface CloudinaryMediaLibraryDialogProps {
+  openDialog: boolean
+  closeDialog(): void
+  selectImage(image: CloudinaryImage): void
+  apiKey: string | undefined
+  cloudName: string | undefined
+}
+
+export interface CloudinaryImage {
```
JacoboGallardo: This honors the data that comes from the Cloudinary Media Library widget.

plugins/cloudinary/src/CloudinaryMediaLibraryDialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+  private openCloudinaryMediaLibrary() {
+    const mediaLibrary = this.generateNewMediaLibrary()
+    this.showMediaLibrary(mediaLibrary)
+  }
+
+  private showMediaLibrary(mediaLibrary: any) {
+    if (mediaLibrary) {
+      mediaLibrary.show({
+        multiple: false,
+        max_files: 1
```
JacoboGallardo: We are limiting the selection to one image only - on purpose.

plugins/cloudinary/src/setupEnzyme.ts
```diff
@@ -0,0 +1,4 @@
+import { configure } from 'enzyme'
```
JacoboGallardo: We need this file so Enzyme, jest and typescript play well together.

plugins/cloudinary/src/CloudinaryCredentialsDialog.tsx
```diff
@@ -0,0 +1,82 @@
...
+  cloudName: string
+}
+
+export default class CloudinaryCredentialsDialog extends React.Component<
+  CloudinaryCredentialsDialogProps,
+  CloudinaryCredentialsDialogState
+> {
+  state = {
+    apiKey: '',
+    cloudName: ''
```
steve8708: it would be great to set these to their current values if already set, that way on subsequent opens it won't confuse the user into thinking they haven't set these if they display blank but the settings do actually already exist

plugins/cloudinary/src/CloudinaryImageEditor.tsx
```diff
@@ -0,0 +1,162 @@
...
+              this.setState({
+                showDialog: !this.state.showDialog
+              })
+            }}
+          >
+            Choose Image
+          </Button>
+
+          <Button
+            variant="contained"
```
steve8708: it would be great to change this to be `variant="text"` if credentials are set already, to take the focus off of this, but make it "contained" if credentials are not set so to make it the focus in that case

plugins/cloudinary/src/CloudinaryImageEditor.tsx
```diff
@@ -0,0 +1,162 @@
...
+            selectImage={this.selectImage.bind(this)}
+            apiKey={this.state.apiKey}
+            cloudName={this.state.cloudName}
+          />
+        )}
+
+        <div>
+          <Button
+            css={buttonStyle}
+            color="primary"
```
steve8708: would be great to make this `disabled` if credentials are not yet set to drive focus to set credentials first

plugins/cloudinary/src/CloudinaryCredentialsDialog.tsx
JacoboGallardo: ok!

JacoboGallardo:
I think I have addressed of the useful UI changes that you have pointed out.

Thanks for the feedback!

PR #11, opened 2019-11-29
Show image public id and customize button caption when image is selected

JacoboGallardo:
This PR adds a label so the user can see the `public id` property of the selected image.

Additionally, changes the pick up image button text to `UPDATE IMAGE` if there is an image previously selected.


steve8708:
Looks great 👍

PR #12, opened 2019-11-29
fix(common): fix for angular example: add options variable

lyczos:
In `app.component.ts`, `options` variable was missing.


steve8708:
Great, thanks for the fix!

PR #14, opened 2019-12-20
fix(react-amp): don't wrap img with <picture> in amp mode

ca136:


packages/react/src/blocks/Image.tsx
```diff
@@ -99,65 +99,71 @@ class ImageComponent extends React.Component<any> {
...
+                      this.props.image.split('?')[0]) ||
+                    undefined
+                  : undefined
+              }
+              // height={
+              //   this.props.height || (aspectRatio ? Math.round(aspectRatio * 1000) : undefined)
+              // }
+              // width={
+              //   this.props.width || (aspectRatio ? Math.round(1000 / aspectRatio) : undefined)
+              // }
```
steve8708: comments here can be removed

packages/react/src/blocks/Image.tsx
```diff
@@ -99,65 +99,71 @@ class ImageComponent extends React.Component<any> {
...
-                    }}
-                    className="builder-image"
-                    src={this.props.image}
-                    // TODO: memoize on image on client
-                    srcSet={srcset}
-                    sizes={this.props.sizes}
-                  />
-                )}
-              </picture>
+              { amp ?
```
steve8708: this is backwards (amp should not show picture)

packages/react/src/blocks/Image.tsx
```diff
@@ -99,65 +99,71 @@ class ImageComponent extends React.Component<any> {
         {value => {
           const amp = value.ampMode
           const Tag: 'img' = amp ? ('amp-img' as any) : 'img'
+
+          const ImageContents = (!lazy || this.state.load) && (
```
steve8708: this should be lowercase `imageContents`

packages/react/src/blocks/Image.tsx
```diff
@@ -99,65 +99,71 @@ class ImageComponent extends React.Component<any> {
...
+                  position: 'absolute',
+                  height: '100%',
+                  width: '100%',
+                  left: 0,
+                  top: 0
+                }),
+                ...(amp && {
+                  ['& img']: {
+                    objectFit: this.props.backgroundSize,
+                    OObjectPosition: this.props.backgroundPosition
```
steve8708: oh this is actually a typo from the original code - should be `objectPosition: ...`

packages/react/src/blocks/Image.tsx
```diff
@@ -99,65 +99,64 @@ class ImageComponent extends React.Component<any> {
...
+                ...(amp && {
+                  ['& img']: {
+                    objectFit: this.props.backgroundSize,
+                    ObjectPosition: this.props.backgroundPosition
+                  }
+                })
+              }}
+              className="builder-image"
+              src={this.props.image}
+              // TODO: memoize on image on client
```
ca136: @steve8708 you want this comment removed as well?

ca136:
@steve8708 updated, didn't add the unit test yet, will follow-up with a new PR for that and fix the tests for Reach SDK (breaking due to:
```
Cannot find module 'lib/on-change.js' from 'builder-page.component.js'
```

packages/react/src/blocks/Image.tsx
steve8708: this one we can keep actually, would be good to add at some point but not urgent

steve8708:
ah yes I need to remove that file from being git ignored, will add to todo list :)

PR looks great!

PR #31, opened 2020-01-06
Remove height property due to Safari layout issues

ca136:


steve8708:
oops, thanks for sending just did too so closing so we don't have merge conclit

PR #35, opened 2020-01-09
adds a key for every imageEditor instance, to avoid images override o…

donnieblando:
By adding an editor key, specific for every cloudinary image editor, it prevents the cloudinary modal dialog to override all the images existing in the page.

Fixes bug on multiple adding of the same script to the document.head


steve8708:
Looks great, thanks @donnieblando !

PR #37, opened 2020-01-11
fresh build with Node 12, clean package-lock, ts issues, etc

ca136:


packages/react/src/components/builder-simple.component.tsx
```diff
@@ -95,11 +97,13 @@ export class BuilderSimpleComponent extends React.Component<BuilderSimpleCompone
     }
 
     if (!builder.apiKey) {
-      const subscription = builder['apiKey$'].subscribe((key?: string) => {
-        if (key) {
-          this.getContent()
+      const subscription = builder['apiKey$'].subscribe(
+        (key?: string | null) => {
```
ca136: added null, was causing an error with null vs undefined

PR #38, opened 2020-01-13
start upgrade to angular 9 and ivy

ca136:


steve8708:
@ca136 is this no longer needed?

Issue #59, opened 2020-02-04
Homepage is not displaying on IE11 - wrong userAttributes


lyczos:
Whenever we are trying to display any content on IE11 on home page we are getting an empty response.
It is because of wrong user attribute URL Path:

On Chrome etc. builder is passing:
`userAttributes.urlPath: '/'`

but on IE11 builder passing empty string:
`userAttributes.urlPath: ''`

ca136:
@lyczos I'm hoping this is fixed in: @builder.io/angular@1.0.83-alpha.6

PR #60, opened 2020-02-04
fix ie11 empty path root

ca136:
fixes https://github.com/BuilderIO/builder/issues/59


packages/core/src/builder.class.ts
```diff
@@ -29,9 +29,14 @@ const urlParser = {
 
     // IE 11 pathname handling workaround
     // (IE omits preceeding '/', unlike other browsers)
-    if (out.pathname && typeof out.pathname === 'string' && out.pathname.indexOf('/') !== 0) {
```
ca136: so if `out.pathname` is empty string, then the conditional would be false and this wouldn't add the `/`

PR #63, opened 2020-02-05
Restrict env overrides

ca136:


packages/core/src/builder.class.ts
```diff
@@ -14,6 +14,7 @@ import serverOnlyRequire from './functions/server-only-require.function';
 
 export type Url = any;
 export const isReactNative = typeof navigator === 'object' && navigator.product === 'ReactNative';
+export const validEnvList = ['production', 'qa', 'development', 'dev', 'cdn-qa'];
```
steve8708: we actually have a couple more of this via https://github.com/BuilderIO/builder/blob/master/packages/core/src/builder.class.ts#L1448:L1448

`cloud` `fast` `cdn2`

PR #64, opened 2020-02-05
update polyfill for wc and fix react includes

ca136:


ca136:
@lyczos I deployed the angular example app to firebase so it's easier to test:
https://angular-basic.firebaseapp.com/?builder.wcVersion=1.0.134-alpha.10

PR #69, opened 2020-02-12
 add input for tab alignment

ca136:


packages/widgets/src/components/Tabs.tsx
```diff
@@ -233,6 +238,19 @@ export const Tabs = withBuilder(TabsComponent, {
...
+      name: 'tabHeaderLayout',
+      type: 'enum',
+      helperText: 'Change the layout of the tab headers (uses justify-content)',
+      defaultValue: 'flex-start',
+      enum: [
+        'center',
+        'space-between',
+        'space-around',
+        'flex-start',
+        'flex-end'
```
steve8708: since some of these names won't be intuitive to people, let's use the alt format that takes a `{ label: '...', value: '...' }` so we can name then more friendly things, like

`{ label: 'Center', value: 'center' }, { label: 'Right', value: 'flex-end' }` etc

ca136:
👍 updated

packages/widgets/src/components/Tabs.tsx
```diff
@@ -243,13 +243,13 @@ export const Tabs = withBuilder(TabsComponent, {
       name: 'tabHeaderLayout',
       type: 'enum',
       helperText: 'Change the layout of the tab headers (uses justify-content)',
-      defaultValue: 'flex-start',
+      defaultValue: { label: 'left', value: 'flex-start' },
```
steve8708: default value has to be a string - `'flex-start'`

PR #70, opened 2020-02-12
Dropdown plugin

fasaas:


plugins/dropdown/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Creating a new plugin from this example
+
+### Install
+
+```bash
+git clone https://github.com/BuilderIO/builder.git
+cp -r plugins/example plugins/your-plugin-name
+cd plugins/your-plugin-name
+npm install
```
steve8708: we can remove the clone / getting started instructions here and just say `npm install` to install

plugins/dropdown/README.md
```diff
@@ -0,0 +1,98 @@
+# Builder.io example plugin
```
steve8708: Let's call this pugin the "dynamic dropdown" plugin (as you can already make dropdowns with text fields, but this one is specifically special because the options can be dynamic from an API)

so let's update the other names, like the foldername to `dynamic-dropdown` etc

open to other suggestions as well like `remote-dropdown` `api-dropdown` etc too

plugins/dropdown/package.json
```diff
@@ -0,0 +1,125 @@
...
+  "keywords": [],
+  "main": "dist/builder-plugin-dropdown.umd.js",
+  "module": "dist/builder-plugin-dropdown.es5.js",
+  "unpkg": "dist/builder-plugin-dropdown.system.js",
+  "typings": "dist/types/builder-plugin-dropdown.d.ts",
+  "files": [
+    "dist"
+  ],
+  "author": {
+    "name": "Jacobo Gallardo",
```
steve8708: This should be Asier :)

plugins/dropdown/package.json
```diff
@@ -0,0 +1,125 @@
+{
+  "name": "@builder.io/plugin-dropdown",
```
steve8708: `@builder.io/plugin-dynamic-dropdown`

plugins/dropdown/src/dropdownPropsExtractor.ts
```diff
@@ -0,0 +1,28 @@
+import Mustache from 'mustache'
+
+const getMassagedProps = (props: any): any => {
+  const { url, mapper } = props.field.options || ({} as any)
```
steve8708: this needs to be `props.field.options.toJSON()` I believe, is it working without that?

steve8708: it's a `Map` type

plugins/dropdown/src/mapperEvaluator.ts
```diff
@@ -0,0 +1,13 @@
+// @ts-ignore
+import SES from 'ses'
+
+const safeEvaluate = (code: String, context: any = {}) => {
+  const realm = SES.makeSESRootRealm({
+    consoleMode: 'allow'
+  })
+
+  const result = realm.evaluate(code, context)
```
steve8708: I think you had mentioned this isn't working, giving errors? I remember when I first integrated it was working and tests passing, but then my tests didn't cover all use cases

can you remind me what the error was, I think it has to do with `Array`?

plugins/dropdown/src/dropdown.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const onSelectChange = (event: React.ChangeEvent<{ value: unknown }>) => {
+    const selectedValue = event.target.value ?? null
+    props.onChange(selectedValue)
+  }
+
+  const getSelections = async (url: any, mapper: String) => {
+    try {
+      const data = await executeGet(url)
+      const mappedSelections = safeEvaluate(mapper, { data })
+      if (mappedSelections.sort) mappedSelections.sort(comparer)
```
steve8708: maybe the mapper should handle the sorting? in case people want other sorting than alphabetical

e.g. a list of products returned may want to leave the sort the same, since it may be by relevance or such

plugins/dropdown/README.md
fasaas: 👍

plugins/dropdown/README.md
fasaas: Dynamic dropdown actually (IMHO) is the most accurate description

plugins/dropdown/package.json
fasaas: 😅

plugins/dropdown/src/dropdownPropsExtractor.ts
fasaas: I had an error stating the `options` didn't have a `toJSON()` method, I could add it back, however I'm not confident it would work

plugins/dropdown/src/dropdown.tsx
fasaas: Definitely agree!

plugins/dropdown/src/mapperEvaluator.ts
fasaas: Funnily, the SES code is actually working now! On the other hand, when selecting a new selection, I am not capable of seeing the selection as selected in the dropdown.

Also, the tests for SES throw an exception

plugins/dynamic-dropdown/README.md
```diff
@@ -0,0 +1,89 @@
+# Builder.io dynamic dropdown
+
+See [here](src/dropdown.tsx) for the React component that powers this plugin
```
steve8708: I should have mentioned sooner, but would be great to describe what this plugin does/why and how to use it (e.g. with simple example here)

plugins/dynamic-dropdown/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Status
+
+Builder plugins are in beta. If you run into any issues or have questions please
+contact help@builder.io for help
```
steve8708: also should have caught this sooner too - but this status sooner can be removed

plugins/dropdown/src/mapperEvaluator.ts
steve8708: interesting, and what is the exception? the `Array` one?

plugins/dropdown/src/dropdownPropsExtractor.ts
steve8708: was that when running in the app or in tests? e.g. we may just need to make sure the tests use a Map instead of an object when passing to the editor like should be the behavior at runtime

fasaas: That was when running the app, not the tests

plugins/dropdown/src/mapperEvaluator.ts
fasaas: `Uncaught Error: [mobx-state-tree] No matching type for union(null | number | string | boolean | Map...`

steve8708: oh great, that should have nothing to do with SES and be easy to fix. I'll pull down your code shortly and see if we can get that error resolved

plugins/dropdown/src/dropdownPropsExtractor.ts
steve8708: cool, I'll pull this down and run it in the app and try it out and see if i can replicate

plugins/dropdown/src/mapperEvaluator.ts
steve8708: sent this PR to address the outstanding issues, once merged I think we can get this PR merged as well! https://github.com/fasaas/builder/pull/3

plugins/dynamic-dropdown/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+Now as you develop you can restart Builder to see the latest version of your plugin.
+
+To uninstall your plugin run
+
+```js
+// Removes all plugins
+builder.plugins.replace([])
+// Saves for all in your organization and reloads the browser
+builder.savePlugins.then(() => location.reload())
```
steve8708: very minor, but: `savePlugins().then`

PR #71, opened 2020-02-12
add the option to autofocus on open

teleaziz:


packages/react/src/blocks/Text.tsx
```diff
@@ -168,6 +168,7 @@ export const Text = withBuilder(TextComponent, {
       name: 'text',
       type: 'html',
       required: true,
+      autofocus: true,
```
steve8708: let's capitalize the `F` to mirror other react `autoFocus` like in materialUI

PR #76, opened 2020-02-21
Gatsby builder plugin

teleaziz:


examples/gatsby-using-plugin/.eslintrc
```diff
@@ -0,0 +1,27 @@
+{
```
steve8708: I think we should just update the default example to use the plugin, rather than have two

examples/gatsby-using-plugin/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2018 Builder.io, Inc
```
steve8708: 2020

examples/gatsby-using-plugin/src/templates/page.tsx
```diff
@@ -0,0 +1,49 @@
...
+    allBuilderModels {
+      docsHeader(limit: 1) {
+        everything
+      }
+      docsNav(limit: 1) {
+        everything
+      }
+      docsContent(target: { urlPath: $path }, limit: 1) {
+        everything
+      }
```
steve8708: similar comment as before, our simple examples should show standard usage rather than usage specific to us. things like the default "page" model

steve8708:
So main piece of feedback here -

If we want to have a gatsby plugin our repo, it should be in `packages/` and we can `npm publish` it from there, but I don't think we want to have one just hard coded in an example since people clone the examples and I don't think we want people cloning the plugin as our canonical use case example, but instead have them `npm install` it like other examples. 

I think the right approach is to wait for the plugin to be merged into the gatsby repo, then make a small update to our existing example to use the plugin, and also link to our open source gatsby docs site as a more fleeshed out example, and likewise update that to use the official pugin once merged and published by gatsby

teleaziz:
Closing in favor of #78

PR #78, opened 2020-02-24
add @builder.io/gatsby-plugin

teleaziz:
## Description

Plugin for integrating [Builder.io](https://www.builder.io) with Gatsby.

### Documentation
it puts the Builder.io schema under an `allBuilderModels` field of the Gatsby GraphQL query, If a templates map in options, this plugin will create pages dynamically for each model entry on the path specified.

#### Install

`npm install --save @builder.io/gatsby-plugin`

#### How to use

```javascript
// In your gatsby-config.js
const path = require('path')
module.exports = {
  plugins: [
    {
      resolve: "@builder.io/gatsby-plugin",
      options: {
        // public API Key
        publicAPIKey: "MY_PUBLIC_API_KEY",
        templates: {
          myPageModel: path.resolve('templates/my-page.tsx')
        }
      },
    },
  ]
}

```

#### How to Query
For an up-to-date complete examples check out the examples on [BuilderIo/SDK](https://github.com/BuilderIO/builder/tree/master/examples)

```graphql
{
  allBuilderModels {
    myPageModel {
      everything
    }
  }
}
```


packages/gatsby-plugin/README.md
```diff
@@ -0,0 +1,42 @@
...
+const path = require("path")
+module.exports = {
+  plugins: [
+    {
+      resolve: "@builder.io/gatsby-plugin",
+      options: {
+        // public API Key
+        publicAPIKey: "MY_PUBLIC_API_KEY",
+        templates: {
+          myPageModel: path.resolve("templates/my-page.tsx"),
```
steve8708: let's call this `page:` since that's the default, and have a comment that this is the name of your page model

packages/gatsby-plugin/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## How to Query
+
+For an up-to-date complete examples check out the examples on [BuilderIo/SDK](https://github.com/BuilderIO/builder/tree/master/examples)
+
+```graphql
+{
+  allBuilderModels {
+    myPageModel {
+      everything
```
steve8708: I created a new alias called `content` we should use now instead of `everything`. `everything` is still there but the name really runs counter to the point of graphql "only get what you need", and the `content` alias we can keep precise to that (return the needed content, but not any additional metadata etc that could be fetched but isn't required for rendering components and pages)

packages/gatsby-plugin/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## How to use
+
+```javascript
+// In your gatsby-config.js
+const path = require("path")
+module.exports = {
+  plugins: [
+    {
+      resolve: "@builder.io/gatsby-plugin",
```
steve8708: let's just call this `@builder.io/gatsby`

packages/gatsby-plugin/README.md
```diff
@@ -0,0 +1,42 @@
...
+        },
+      },
+    },
+  ],
+}
+```
+
+## How to Query
+
+For an up-to-date complete examples check out the examples on [BuilderIo/SDK](https://github.com/BuilderIO/builder/tree/master/examples)
```
steve8708: let's call this `BuilderIO/examples/gatsby` and link direct to the gatsby example

packages/gatsby-plugin/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@builder.io/gatsby-plugin",
```
steve8708: let's just call it `@builder.io/gatsby`

packages/gatsby-plugin/src/builder-config.js
```diff
@@ -0,0 +1,25 @@
...
+}
+
+module.exports.getGQLOptions = options => {
+  const config = {
+    ...module.exports.defaultOptions,
+    ...options,
+  }
+  invariant(
+    config.publicAPIKey && config.publicAPIKey.length > 0,
+    `gatsby-plugin-builder requires a public API Key`
```
steve8708: if our plugin is called `@builder.io/gatsby` should all of these `gatsby-plugin-builder` be replaced with `@builder.io/gatsby`?

packages/gatsby-plugin/src/gatsby-node.js
```diff
@@ -0,0 +1,137 @@
...
+  }
+  const { createPage } = actions
+  const models = Object.keys(config.templates)
+  const result = await graphql(`
+    query {
+      ${config.fieldName} {
+        ${models
+          .map(
+            model => `${model} {
+            everything
```
steve8708: let's replace all `everything` with `content`

packages/gatsby-plugin/src/gatsby-node.js
```diff
@@ -0,0 +1,137 @@
...
+}
+
+exports.createPages = async ({ graphql, actions }, options) => {
+  const config = {
+    ...defaultOptions,
+    ...options,
+  }
+  const { createPage } = actions
+  const models = Object.keys(config.templates)
+  const result = await graphql(`
```
steve8708: we need to paginate this right?

steve8708:
beautiful!

PR #80, opened 2020-02-25
fix a bug where an incremental patch is reverting other props to old …

teleaziz:
…values


steve8708:
I'm nervous this will cause problems in certain situations, as the `block={...}` prop can change outside of an inner state change (i.e. the outter data can change passing in the new block)

Is there a way to accomplish the same goal without an additional wrapping? 

In order to help here, can you give me more context as to what exactly is causing this issue?

 I may be able to suggest a similar solution that wouldn't cause possible other issues

teleaziz:
Closing this as this introduces another set of issues

PR #81, opened 2020-02-27
typo in 404 link

muescha:


muescha:
i would also suggest a link to the docs, because this link is behind a login and not visible for first time devs

steve8708:
Great idea, will add the link to the docs as you suggest too, thanks @muescha !

PR #86, opened 2020-03-02
fix gatsby example

teleaziz:


examples/gatsby/README.md
```diff
@@ -33,3 +33,40 @@ See:
...
+import { graphql } from 'gatsby';
+import { BuilderComponent } from '@builder.io/react';
+
+
+export default class ExamplePage extends React.Component<any> {
+  render() {
+    const { header, page } = this.props.data;
+    return (
+      <div>
+        <BuilderComponent name="header" content={header[0].content} />
```
steve8708: Would be good to also mention this can be your `<Header />` instead, or just default to that, bc that's what 95+% of people do (use builder pages within their react headers and footers)

examples/gatsby/src/templates/page.tsx
```diff
@@ -1,7 +1,11 @@
 import * as React from 'react';
 import { graphql } from 'gatsby';
 import { BuilderComponent } from '@builder.io/react';
-import '../components/CodeBlock';
+/**
+ * Import all your custom components here If you're using any with builder.io,
```
steve8708: would be good to have an example here instead as most people won't understand this

e.g. add ../components/custom-block-example.ts

and in that example link to the docs for custom react components and the design system example

examples/gatsby/README.md
```diff
@@ -33,3 +33,40 @@ See:
 <img width="796" alt="Screen Shot 2020-02-18 at 9 48 51 AM" src="https://user-images.githubusercontent.com/5093430/74763082-f5457100-5233-11ea-870b-a1b17c7f99fe.png">
 
 When you deploy this to a live or staging environment, you can change the preview URL for your model globally from [builder.io/models](https://builder.io/models) (see more about models [here](https://builder.io/c/docs/guides/getting-started-with-models) and preview urls [here](https://builder.io/c/docs/guides/preview-url))
+
+This example create pages dynamically based on the url you add to your entries on [Builder.io](https://www.builder.io), if you want to create a page manually, do not include the model in your `tempaltes` config as above, add a file under the `pages` folder and query all the entries your page needs from Builder.io, for example:
+
+```ts
+import React from 'react';
```
steve8708: this is a great addition to the docs 👍

examples/gatsby/src/components/CodeBlock.tsx
```diff
@@ -0,0 +1,43 @@
...
+import javascript from 'react-syntax-highlighter/dist/cjs/languages/hljs/javascript';
+import githubGist from 'react-syntax-highlighter/dist/cjs/styles/hljs/github-gist';
+import { Builder } from '@builder.io/react';
+
+SyntaxHighlighter.registerLanguage('javascript', javascript);
+
+// Example of a custom component you can use in the Builder visual editor
+// See here https://github.com/BuilderIO/builder/tree/master/examples/react-design-system
+// for many more custom component examples
+function CodeBlockComponent({ language, code }: any /* TODO: types */) {
```
steve8708: so this code block component I think is not the best example for us to use, it's pretty particular to our use case and i think all the stuff about "code" and enter code can cause confusion

something that will be more relatable like a simple button or hero, e.g. just strait copy and pasted from our design system example, will be a much more relatable and understandable example for people

steve8708:
👍 👍 👍

PR #89, opened 2020-03-04
do not show 'undefined' for custom css

ca136:
this was causing an issue where ie11 stopped parsing the rest of the css in the block.


steve8708:
great 👍

PR #91, opened 2020-03-05
add the option to render a custom link component

teleaziz:


packages/react/src/components/Link.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { BuilderStoreContext } from '../store/builder-store'
+
+export const Link: React.SFC<React.AnchorHTMLAttributes<
+  HTMLAnchorElement
+>> = props => (
```
ca136: was this formatted by prettier like this? haven't seen wrapping like this before

packages/react/src/components/builder-block.component.tsx
```diff
@@ -450,7 +451,7 @@ export class BuilderBlock extends React.Component<
   getElement(index = 0, state = this.privateState.state): React.ReactNode {
     const { child, fieldName } = this.props
     const block = this.block
-    let TagName = (block.tagName || 'div').toLowerCase()
+    let TagName: string | typeof Link = (block.tagName || 'div').toLowerCase()
```
ca136: Is it a common pattern to use `typeof` for getting the 'type' of React components?

packages/react/src/components/Link.tsx
teleaziz: yes , I ran prettier on both changed files in this PR

packages/react/src/components/builder-block.component.tsx
teleaziz: Not sure, but it's less verbose than writing React.SFC<React.AnchorHTMLAttributes<HTMLAnchorElement>>

teleaziz:
@ca136 I just read your feedback comment above, I'll include an explanation on the link component next pr

PR #94, opened 2020-03-08
@builder.io/storybook

teleaziz:
# @builder.io/storybook

Addon for integrating [Builder.io](https://www.builder.io) to allow drag and drop page building inside storybook. 

example

## Install

`npm install @builder.io/storybook`

## How to use

Make a free account over at [Builder.io](https://www.builder.io/) and grab your public API key from your [organization page](https://builder.io/account/organization)

## Using your components in the editor

See this [design systems example](/examples/react-design-system) for lots of examples using your deisgn system + custom components + storybook

👉**Tip: want to limit page building to only your components? Try [components only mode](https://builder.io/c/docs/guides/components-only-mode)**

Register a component

```tsx
import { Builder } from '@builder.io/react';

class SimpleText extends React.Component {
  render() {
    return <h1>{this.props.text}</h1>;
  }
}

Builder.registerComponent(SimpleText, {
  name: 'Simple Text',
  inputs: [{ name: 'text', type: 'string' }],
})

```

in `.storybook/preview.js`
Add `builderDecorator` as a global decorator for your storybook and pass `BuilderComponent` as a parameter 
```tsx
import { addParameters, addDecorator } from '@storybook/react';
import { BuilderComponent } from '@builder.io/react';
import { builderDecorator } from '@builder.io/storybook'
// builder-settings is where you configure your builder instance: init with api key, add custom menus ..
import '../src/builder-settings';

// add global decorator
addDecorator(builderDecorator)

// pass preview component
addParameters({
  builder: {
    component: BuilderComponent
  },
});

```

## Learn more

- [Design system example](/examples/react-design-system)
- [Official docs](https://www.builder.io/c/docs/getting-started)


examples/react-design-system/src/builder-menus.js
```diff
@@ -0,0 +1,85 @@
+export const menus = [
```
steve8708: why this change? i try to remove non-critical abstractions from examples so people can more clearly see things right inline and not have to chase down references and understand how used. it also allows easier type completion to leave as it was

`Builder.register('insertMenu', { ... }`

examples/react-design-system/src/builder-settings.js
```diff
@@ -11,86 +29,4 @@ if (OVERRIDE_INSERT_MENU) {
...
-          options: {
-            text: 'I am am h3',
-            type: 'h3',
-          },
-        },
-      },
-    },
-  ],
-});
+menus.forEach(menu => Builder.register('insertMenu', menu));
```
steve8708: if the goal of separating the menu configs was just to clean up this file, perhaps we just make a separate file for the menu settings that does the `Builder.register(...)`?

packages/storybook/src/Tab.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { RenderOptions } from '@storybook/addons';
+import { useStorybookState, useStorybookApi } from '@storybook/api';
+import { config } from './config';
+import { AddonPanel } from '@storybook/components';
+
+const style = {
+  display: 'block',
+  height: '90vh',
```
steve8708: 90vh doesn't ever seem to fit the window right, can this be `calc(100vh - _px)` instead?

packages/storybook/src/Tab.tsx
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  return (
+    <AddonPanel {...props}>
+      {storyId ? (
+        <builder-editor
+          style={style}
+          options={`{
+        "floatingLeftSidebar": false,
+        "rtlMode": false,
```
steve8708: `rtlMode` and `floadtingLeftSidebar` isn't needed here

something I think we should do tho is add a couple options (or one simple `storybookMode`) option where we can do some handy things like allowing saving pages made to files and/or react code, etc

packages/storybook/src/util.ts
```diff
@@ -0,0 +1,17 @@
+export const transformConfigToProps = (config: { inputs: any[] }) => {
```
steve8708: should we call this `getDefaultProps` instead as it seems that is what this does - grabs default props object from inputs?

packages/storybook/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+  "compilerOptions": {
+    "declaration": true,
+    "lib": ["es2017", "dom"],
+    "jsx": "react",
+    "moduleResolution": "node",
+    // "allowSyntheticDefaultImports": true,
```
steve8708: remove?

examples/react-design-system/src/builder-menus.js
teleaziz: left over from a former config where it required the menus to be passed down, I'll revert

packages/storybook/src/Tab.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const api = useStorybookApi()
+  requestAnimationFrame(() => api.toggleFullscreen(props.active))
+  const storyId = state.storyId;
+  React.useEffect(() => {
+    const script = document.getElementById(config.addonId);
+    if (script) {
+      return;
+    }
+    const builderScript = document.createElement('script');
+    builderScript.src = 'https://cdn-qa.builder.io/js/editor';
```
steve8708: why `cdn-qa`? this can/should just be `cdn.builder.io/js/editor`

steve8708:
This is great. Last feedback is just about how to make explaining how to use this as clear as possible to new people so they don't get stuck/confused/lost/frustrated. Some areas I got stuck at first that could use some pointing out

- We point to the design system example from storybook but don't mention how to run the storybook server. Our design system example readme would benefit from a section on storybook on how to run it, add a new story, etc
- How exactly do I add a new component for Builder.io and/or storybook. E.g. it takes adding/updating a few files - explaining which for what is helpeful (e.g. a "how to add a component" section with subsection how to builder-enable it (add the config, add the import) and how to storybook enable it too
- Connection issues. This is tricky because if everything isn't setup fully right (and given how things are currently this will be the case by default) users get these "can't connect" issues. This has led to a lot of user confusion and dropoff. Coming up with something here (like a way to use https magically somehow by default) would be huge. My other hypothesis is we launch a modal with helpful info when the connection doesn't work but people dismiss it without reading it. We should add something in the editor that instead of hiding it completely, minimizes it to some kind of warning bar that makes it clear to click it to reopen the modal to troubleshoot how to connect

examples/react-design-system/package.json
```diff
@@ -20,6 +20,7 @@
   "main": "src/index.js",
   "dependencies": {
     "@builder.io/react": "^1.1.11-1",
+    "@builder.io/storybook": "file:../../packages/storybook",
```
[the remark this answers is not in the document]
teleaziz: good point! I was planning to update this as soon the package is merged and published

examples/react-design-system/.storybook/preview.js
```diff
@@ -8,5 +8,7 @@ addDecorator(builderDecorator)
 addParameters({
   builder: {
     component: BuilderComponent,
+    // for local development
+    // isDev: true,
```
steve8708: prob remove? I think this is only for _us_ to develop locally tho

packages/storybook/README.md
```diff
@@ -0,0 +1,72 @@
...
+}
+
+Builder.registerComponent(SimpleText, {
+  name: 'Simple Text',
+  inputs: [{ name: 'text', type: 'string' }],
+})
+
+```
+
+in `.storybook/preview.js`
```
steve8708: It would be good to separate this as it's not required, e.g. 

### Storybook (optional)

To add a story for this component ...

packages/storybook/README.md
```diff
@@ -0,0 +1,72 @@
...
+  builder: {
+    component: BuilderComponent
+  },
+});
+
+```
+
+### Mixed Content errors
+Our embedded editor is on https and it'll try to load your current setup from your localhost, which will be blocked by the browser when serving from http, you can fix this by serving storybook on https (ie. `start-storybook -p 9009 -s public --https --ssl-cert [path-to.crt] -ssl-key [path-to.key]`)
+
```
steve8708: it would be good to clarify that the above workaround is only needed if you _don't_ do the SSL bit above

packages/storybook/README.md
```diff
@@ -0,0 +1,72 @@
+# @builder.io/storybook
+
+Addon for integrating [Builder.io](https://www.builder.io) to allow drag and drop page building inside storybook.
```
steve8708: We should host the storybook build somewhere so we can link people to be able to just click to try it out

packages/storybook/README.md
steve8708: oops ignore this, I thought this was the design system readme, since this is the storybook readme yeah this makes more sense as it is required in this context

packages/storybook/README.md
```diff
@@ -0,0 +1,72 @@
+# @builder.io/storybook
+
+Addon for integrating [Builder.io](https://www.builder.io) to allow drag and drop page building inside storybook. 
+
+<img src="https://user-images.githubusercontent.com/5093430/76154244-ebbe6480-608d-11ea-9dc9-08a59eda220e.gif" alt="example" />
```
steve8708: we should also update the react-esign-system readme to mention this shows storybook too (linking to this readme and to the hosted storybook for this we should add)

steve8708:
Going to merge this so I can publish to NPM - afterward can you follow up replacing the @buidler.io/storybook in the design system example to use the live package and also with the feedback

PR #99, opened 2020-03-13
update readmes and isDev

teleaziz:


examples/react-design-system/README.md
```diff
@@ -1,13 +1,20 @@
-## Builder.io example using a strict custom design system in React
+## Builder.io example using a strict custom design system in React and integrating builder.io with storybook
```
teleaziz: This title makes me think maybe we should separate the examples for clariity

steve8708: yeah I thought that at a point but now I'm torn. Let's leave for now and just call this 

## Builder.io custom design system example with React and storybook

PR #105, opened 2020-03-16
feedback, and add config for linking stories to builder on dblclick

teleaziz:


examples/react-design-system/src/builder-settings.js
```diff
@@ -1,4 +1,5 @@
 import { Builder, builder } from '@builder.io/react';
+import '@builder.io/widgets';
```
steve8708: if we want this in we should add a comment here that adding builder.io/widgets adds a carousel, tabs, and slider component for use in the editor

teleaziz: I ended up removing it

steve8708: great

PR #107, opened 2020-03-17
remove global from window to fix realm context error

teleaziz:


plugins/dynamic-dropdown/src/mapperEvaluator.ts
```diff
@@ -2,6 +2,13 @@
 import SES from 'ses'
 
 const safeEvaluate = (code: String, context: any = {}) => {
+  if (typeof window !== 'undefined') {
+    // workaround realm-shems thinking it's a node context
+    // Remove once this line updated
+    // testing isNode = (this===global)
+    // https://github.com/Agoric/realms-shim/blob/master/src/unsafeRec.js#L28
+    (window as any).global = null
```
steve8708: since this code runs in our actual app, I'm concerned it could cause other issues of our code or libraries being confused about the environment by adding a `global` variable like nodeJS has

teleaziz:
working on updating ses and using the new syntax, will reopen

PR #109, opened 2020-03-18
update deps on @builder.io/react so lerna builds in correct order

ca136:


packages/react/package.json
```diff
@@ -24,6 +24,7 @@
     "_unlink": "npm unlink *",
     "_install": "npm install",
     "lint": "tslint -t codeFrame '../**/*.ts' 'test/**/*.ts'",
+    "prebuild": "rimraf dist",
```
ca136: @steve8708 added this back ^

Issue #111, opened 2020-03-22
(packages/angular) When prerender is set to false then handleRouting flag is ignored


lyczos:
When I've set `prerender` flag to `false` then the angular router is not handling routing (pages are fully reloaded).
When I set this flag back again to `true` everything is working as expected

PR #112, opened 2020-03-22
fix(angular): handle clicks also when prerender is set to false

lyczos:
fix for: https://github.com/BuilderIO/builder/issues/111


steve8708:
great catch, thanks @lyczos !

lyczos:
@steve8708 when it will be included in the release? it's affecting our production :(

steve8708:
@lyczos just pushed @builder.io/angular@1.1.3-alpha.1 with your update

PR #113, opened 2020-03-23
override custom 404 in development mode to allow dynamic page creatin…

teleaziz:
…g and editing


packages/gatsby/src/gatsby-node.js
```diff
@@ -92,60 +82,79 @@ function createSchemaNode({ id, typeName, fieldName, createContentDigest }) {
...
-  })
+    });
+  });
   if (hasMore) {
     await createPagesAsync(config, createPage, graphql, models, offsets);
   }
-}
+};
+
+exports.onCreatePage = ({ page, actions }, options) => {
```
teleaziz: This is the only change in this file, rest is prettier

PR #114, opened 2020-03-23
fix(angular): add control for navigationTimeout

lyczos:
Give a possibility to disable/change the timeout delay. Sometimes on the target route, users can have resolver and it can increase the time needed for all checks/data fetch. 
In my opinion, this should be disabled by default but i don't want to affect other users that are using it right now.

`navigate` and `naviagteByUrl` in Angular rejecting promise on fail


steve8708:
this is great, thanks @lyczos

steve8708:
published to @builder.io/angular@1.1.3-alpha.2

PR #116, opened 2020-03-26
fix(angular): fix for custom-elements-es5-adapter

### LINKED ISSUE #115: Issue with 3rd party libs which using webcomponents (elevio)
When we are trying to use Elevio we are getting error.
image
Elevio code where issue appears:
image

And because of this error all angular components are gone (we are not seeing them on the page).

 Elevio also using web components and it seems that it is an issue. 

Repro steps:

1. Create page in bulder.io and use on it Angular component
2. Add elevio scirpt
3. Navigate to another route (using angular router)
4. Go back to page with builder.io template that has angular component

lyczos:
fixes #115


steve8708:
Great, thanks @lyczos !

steve8708:
Published @builder.io/angular@1.1.3-alpha.4 with this fix

PR #133, opened 2020-04-01
release recent updates as dev version

ca136:


ca136:
Not sure why the package-lock files had tabs before this...

PR #143, opened 2020-04-14
fix duplicate content when the same data is passed down on two or mor…

teleaziz:
…e page components


packages/react/src/components/builder-page.component.tsx
```diff
@@ -781,7 +781,7 @@ export class BuilderPage extends React.Component<
     const dataString =
       this.props.data && size(this.props.data) && hash(this.props.data)
     let key = Builder.isEditing ? this.name : this.props.entry
-    if (!Builder.isEditing && dataString && dataString.length < 300) {
+    if (key && !Builder.isEditing && dataString && dataString.length < 300) {
```
teleaziz, line 784: `undefined:${hashOfObject}` this was causing two components of with the same url but different models to show the same content

steve8708:
released with @builder.io/react@1.1.13-11

Issue #144, opened 2020-04-15
Changing attributes e.g `aspect ratio` breaks images in amp-mode.


homanp:
Steps to reproduce: 

1) Create an AMP page inside the editor.
2) Add an image with desired ratio and other attributes.
3) Save page. 
4) Edit image attributes such as aspect ratio.

No images will be visible and the following error shows up in the console: 

```
Uncaught Error: The last option must not have a media condition​​​
    at Za.f.assert (log.js:429)
    at D (log.js:995)
    at new mi (size-list.js:167)
    at li (size-list.js:136)
    at HTMLElement.Di.b.applySizesAndMediaQuery (custom-element.js:680)
    at ci.f.applySizesAndMediaQuery (resource.js:374)
    at h (resources-impl.js:1173)
    at yj.setState (finite-state-machine.js:72)
    at Cj.f.doPass (resources-impl.js:743)
    at ji.Ug (resources-impl.js:134)
```

steve8708:
Oh interesting, @teleaziz I think this may be per your latest update where you remove the duplicates (which removes the srcset url without the size aka "media condition")

homanp:
@teleaziz any news on this? Still seeing broken images with the above error.

mrkoreye:
@homanp Sorry it is still an issue for you!

I am trying to reproduce on my end and I am not able to. Can you provide any more context on how to reproduce the error?

I will let you know if I find and fix any other issues that might be causing or are related this one.

homanp:
Hey! 

Well You can see the errors here https://hs.mersenne.now.sh/. All images Have that issue.

And We are experiencing it on other AMP pages as well. 

Is there anything else you need?

homanp:
Just to clarify, this only happens on models with the type AMP and started appearing after your latest image fix.

mrkoreye:
Thanks, that should help! I will let you know if I need anything else to investigate.

homanp:
@mrkoreye so I figured out that this only happens in amp-mode when you put an image inside a box. Not otherwise.

PR #147, opened 2020-04-22
Added sizes logic, ensure lazy loading, fixed typo

mrkoreye:
This PR fixes a couple of things
- it ensures that lazy loading of images never occurs when in `ampMode`
- it ensures that the final entry in the image `sizes` string does not have a media query. This should fix https://github.com/BuilderIO/builder/issues/144 
- fixes what looks to be a typo (`date-source` -> `data-source`) that was causing the preview in the Builder editor to fail

### DIFF
--- a/packages/react/src/blocks/Image.tsx
+++ b/packages/react/src/blocks/Image.tsx
@@ -109,11 +109,35 @@ class ImageComponent extends React.Component<any> {
     return getSrcSet(url)
   }
 
+  getSizes(sizes: string) {
+    if (!sizes) {
+      return ''
+    }
+
+    const splitSizes = sizes.split(',')
+    const sizesLength = splitSizes.length
+    return splitSizes
+      .map((size: string, index) => {
+        if (sizesLength === index + 1) {
+          // If it is the last size in the array, then we want to strip out
+          // any media query information. According to the img spec, the last
+          // value for sizes cannot have a media query. If there is a media
+          // query at the end it breaks AMP mode rendering
+          // https://github.com/ampproject/amphtml/blob/b6313e372fdd1298928e2417dcc616b03288e051/src/size-list.js#L169
+          return size.replace(/\([\s\S]*?\)/g, '').trim()
+        } else {
+          return size
+        }
+      })
+      .join(', ')
+  }
+
   render() {
     const { aspectRatio, lazy } = this.props
     const children = this.props.builderBlock && this.props.builderBlock.children
 
     let srcset = this.props.srcset
+    const sizes = this.getSizes(this.props.sizes)
     const image = this.props.image
 
     if (srcset && image && image.includes('builder.io/api/v1/image')) {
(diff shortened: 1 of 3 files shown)


packages/react/src/blocks/Image.tsx
```diff
@@ -109,11 +109,35 @@ class ImageComponent extends React.Component<any> {
...
+          // If it is the last size in the array, then we want to strip out
+          // any media query information. According to the img spec, the last
+          // value for sizes cannot have a media query. If there is a media
+          // query at the end it breaks AMP mode rendering
+          // https://github.com/ampproject/amphtml/blob/b6313e372fdd1298928e2417dcc616b03288e051/src/size-list.js#L169
+          return size.replace(/\([\s\S]*?\)/g, '').trim()
+        } else {
+          return size
+        }
+      })
```
steve8708, line 131: idk if we want to just remove the size, perhaps duplicate the largetst instead

e.g. `sizes="(max-width: 998px) 33vw, (max-width: 490px) 28vw"`

I don't think we want to just remove that media query,  think it would be safer to be smart about what the default is, or just duplicate the latest size so that its' 

`sizes="(max-width: 998px) 33vw, (max-width: 490px) 28vw, 28vw"`

I'm not sure what the right logic is, maybe the final one is the default? so maybe it should represent what the largest max-width query gives if the final one has a query??

mrkoreye: The browser evaluates the list of size conditions in order, until it matches one. We put them in ascending `max-width` in the code that generates the sizes list, meaning that `sizes="(max-width: 490px) 28vw, (max-width: 998px) 33vw, 33vw"` would be the same thing as `sizes="(max-width: 490px) 28vw, 33vw"` as far as I can tell. The reason I did not just append the last width as the default was that I did not want to cause any other duplicate size issues like before. However, now that I am looking at it again I don't think that will be an issue.

Want me to just change the logic to append whatever the last width is to the list of sizes, instead of removing the media query?

packages/react/src/blocks/Image.tsx
```diff
@@ -109,11 +109,35 @@ class ImageComponent extends React.Component<any> {
     return getSrcSet(url)
   }
 
+  getSizes(sizes: string) {
```
teleaziz, line 112: since this doesn't rely on any instance property, what do you think of hoisting it up like the `getSrcSet` function

packages/react/src/blocks/Image.tsx
steve8708: hm yea ideally then the biggest one we would append to the end. how confident are you that we _always_ put the biggest one last?

mrkoreye: Ha, yeah I am not 100% confident we _always_ put the biggest one last, but looking at our code we should be. Also, looking at the AMP implementation as far as I can tell when they parse the sizes string, it is in order from left to right (so whatever order we give them). Then when they actually select it they just iterate over that list in order until they find a matching media query: https://github.com/ampproject/amphtml/blob/b6313e372fdd1298928e2417dcc616b03288e051/src/size-list.js#L183

So if we are not putting it in the right order then the optimization will not really be working correctly anyway. I am cool with either approach here (append an additional width that is the last in the list, or keep it how I have it). Which do you prefer?

steve8708: oh shit I see what you are saying now re: left to right. ok cool we can go with this

mrkoreye:
@steve8708 updated

PR #148, opened 2020-04-22
Block direct access to publish

ca136:
image


steve8708:
does this also run if you publish from the project root (e.g. `npm run release:patch`)?

also does this cause issues when we use scripts for build && patch && release? or maybe that's ok bc publish is (I think) always the last step?

ca136:
Yeah you always have to add the env var if you want to be able to publish. We can add it to the other scripts, wanted to make sure this was the direction we wanted to go before making too many of those changes.

steve8708:
ah sorry I was still thinking this was prompting us. cool works for me

steve8708:
only one last comment - let's get these env vars into the project root release scripts too

packages/webcomponents/tools/confirm-publish.ts
```diff
@@ -0,0 +1,6 @@
+if (!process.env.BYPASS) {
```
mrkoreye, line 1: Maybe change to something more specific, like `BYPASS_BUILDER_PUBLISH`? Not really a big deal, but just in case someone has an env variable that is `BYPASS` for some other reason

steve8708: oh yeah or just like ALLOW_PUBLISH or BLOCK_PUBLISH === 'false' or something

package-lock.json
```diff
@@ -2,39 +2,6 @@
   "requires": true,
   "lockfileVersion": 1,
   "dependencies": {
-    "@babel/runtime": {
```
ca136, line 5: not sure why all of these are removed, I just installed a new dev dependency at the root of the repo

package.json
```diff
@@ -20,9 +21,10 @@
...
-    "release:patch": "npm run verify-git-not-dirty && npm run build && lerna version patch --yes && lerna run build:unpkg:dist --no-private && lerna publish from-package --yes",
-    "release:minor": "npm run verify-git-not-dirty && npm run build && lerna version minor --yes && lerna run build:unpkg:dist --no-private && lerna publish from-package --yes",
-    "release:prerelease:dev": "npm run build && lerna version prerelease --yes && lerna run build:unpkg:dist --no-private && lerna publish from-package --dist-tag dev --yes",
-    "release:dev": "npm run release:prerelease:dev"
+    "release:patch": "npm run verify-git-not-dirty && npm run build && lerna version patch --yes && lerna run build:unpkg:dist --no-private && ALLOW_PUBLISH=true lerna publish from-package --yes",
+    "release:minor": "npm run verify-git-not-dirty && npm run build && lerna version minor --yes && lerna run build:unpkg:dist --no-private && ALLOW_PUBLISH=true lerna publish from-package --yes",
+    "release:prerelease:dev": "npm run build && lerna version prerelease --yes && lerna run build:unpkg:dist --no-private && ALLOW_PUBLISH=true lerna publish from-package --dist-tag dev --yes",
```
ca136, line 26: I haven't been able to fully test this one due to memsource error:
```
@builder.io/memsource-connector: > webpack --mode production
@builder.io/memsource-connector: sh: webpack: command not found
```

package.json
```diff
@@ -1,6 +1,7 @@
 {
   "private": true,
   "devDependencies": {
+    "block-publish": "^1.1.1",
```
steve8708, line 4: neat

next time we should put in in the @builder.io org and publish there, maybe people will use this and discover us!

ca136:
published this so we didn't have to duplicate the little script: https://www.npmjs.com/package/block-publish

package.json
steve8708: maybe that repo just needs an `npm install webpack --dev`?

also we can have these scripts exclude memsource-connector with a lerna filter

package.json
ca136: updated, going to make a little repo for it too

package.json
ca136: did that, should we have lerna ignore all plugins? I removed them from the lerna.json file but can add back and manually ignore this one

PR #149, opened 2020-04-22
include custom event polyfill in core

teleaziz:


teleaziz:
closing for testing, will reopen later

packages/core/src/builder.class.ts
```diff
@@ -1,3 +1,4 @@
+import 'custom-event-polyfill';
```
steve8708, line 1: I actually vaguely recall us using this then removing it due to a problem. can you check the source code to see if this is identical to what we used in the webcomponents? or possibly check the git history as to when we changed this over to be manually copy + pasted and edited

my vague recollection was it defining the CustomEvent even if it already was defined or something. or not being compatible server side (not checking for window existing) first

teleaziz:
This lib is similar to the polyfill we have in,  except it handles preventing default on IE:
https://github.com/kumarharsh/custom-event-polyfill/blob/813029a466f0a0c7174db86656634ea2c74b1011/polyfill.js#L16-L18

checking history now

packages/core/src/builder.class.ts
teleaziz: ok in this case will use the old polyfill

packages/core/src/polyfills/custom-event-polyfill.js
```diff
@@ -0,0 +1,12 @@
+;(function() {
```
steve8708, line 1: we will need to check that window exists too - webcomponents only loads in browser but core loads everywhere (browser, server, native, possibly even webworker)

so just a `if (typeof window === 'undefined') { return }` should suffice

teleaziz: it's in the last commit, but will update to use this syntax

steve8708: oh oops I missed that (clicked via email), what you have totally works

Issue #154, opened 2020-04-29
amp-img sizes attribute sizes images in half on mobile viewports.


homanp:
The `sizes` attribute on `amp-img`-tags sizes images to half the viewport event though the parent element has a 100% width. The amp-img nodes also has a redundant CSS class with `position: absolute`

Any suggestions on how to resolve this issue?

<img width="1792" alt="Screenshot 2020-04-29 at 08 44 25" src="https://user-images.githubusercontent.com/2464556/80567915-47560000-89f6-11ea-8519-9c6f62fd9c79.png">

<img width="1792" alt="Screenshot 2020-04-29 at 08 49 52" src="https://user-images.githubusercontent.com/2464556/80567996-77050800-89f6-11ea-817f-3fd9daf71203.png">

kevva:
This is related to https://github.com/BuilderIO/builder/commit/734de9b524412beee8123f14dd25146b017c93b7. Without the sizer that sets padding, the image needs an explicit height since it's absolutely positioned.

AMP maintains aspect ratio for the images already, as long as you set `width` and `height`, so you should probably remove these lines as well https://github.com/BuilderIO/builder/blob/master/packages/react/src/blocks/Image.tsx#L191-L197.

mrkoreye:
Thanks for the tip, @kevva! I will take a look today

kevva:
It works as far as I can see now. @homanp, can you confirm?

homanp:
Works like a charm @mrkoreye. Closing :)

PR #155, opened 2020-04-29
Removed sizes attribute for amp mode

mrkoreye:
According to the amp docs, we don't need to add `sizes` if we are already including `srcset`. If we include `sizes`, then amp will add an inline style to the image to make it whatever width we set in the sizes attribute (https://amp.dev/documentation/components/amp-img/). I believe this is causing the issue here https://github.com/BuilderIO/builder/issues/154. I also removed the absolute positioning css that gets added for elements with `aspectRatio` since amp will handle that.


steve8708:
released in `@builder.io/react@1.1.16-alpha.8`

PR #158, opened 2020-05-04
Update Next.js example to use `getServerSideProps`

kevva:
It's also possible to use `getStaticProps`, but then you'd need to build the site every time you update something in Builder so I went with `getServerSideProps`.

https://nextjs.org/docs/basic-features/data-fetching

### DIFF
--- a/examples/next-js/pages/[...slug].js
+++ b/examples/next-js/pages/[...slug].js
@@ -8,43 +8,47 @@ import Nav from '../components/nav';
 const BUILDER_API_KEY = 'YOUR_KEY';
 builder.init(BUILDER_API_KEY);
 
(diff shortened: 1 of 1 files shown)


examples/next-js/pages/[...slug].js
```diff
@@ -8,43 +8,47 @@ import Nav from '../components/nav';
...
-        </div>
-      </>
-    );
-  }
-}
+  return {
+    props: {
+      builderPage: {
+        ...page,
+        testVariationName: page?.testVariationName || null
```
Comment (kevva) [inline examples/next-js/pages/[...slug].js:29]: Needed because if `undefined` it cannot be serialized.

image

steve8708:
thanks @kevva !

PR #163, opened 2020-05-06
update @builder.io/storybook in design system example

teleaziz:


examples/react-design-system/package.json
```diff
@@ -20,6 +20,7 @@
   "main": "src/index.js",
   "dependencies": {
     "@builder.io/react": "^1.1.11-1",
+    "@builder.io/storybook": "0.0.2",
```
teleaziz, line 23: oops, should be dev, updating

PR #171, opened 2020-05-15
add proof of concept for rich text plugin

ca136:


plugins/rich-text/README.md
```diff
@@ -0,0 +1,98 @@
+# Builder.io example plugin
```
steve8708, line 1: would be good to update the reamde to explain what this is. Does need a lot of updates, just mention this is an example rich text plugin replacing the built in rich text editor, and update any references to `'file'` below as the original example plugin replaces the file picker

steve8708:
Going to merge so we can get this in but would be great to follow up with the readme updates

PR #173, opened 2020-05-20
Fix for using inline symbols

mrkoreye:
We should show the inline content of a symbol in the editor, not only on render. The previous logic caused the editor to display the same content for repeated symbols that rely on inline content. In this case, one user made a "container" symbol and used inline content to change the contents of each one. Here is a screenshot of the issue this fixes.

Before:
Screen Shot 2020-05-20 at 1 56 15 PM

After:
Screen Shot 2020-05-20 at 2 00 45 PM

### DIFF
--- a/packages/react/src/components/builder-page.component.tsx
+++ b/packages/react/src/components/builder-page.component.tsx
@@ -573,6 +573,11 @@ export class BuilderPage extends React.Component<
         )
       })
     }
+
+    console.log({
+      propsContent: this.props.content,
+      content: this.content
+    })
   }
 
   updateState = (fn?: (state: any) => void) => {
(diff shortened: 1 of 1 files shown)


packages/react/src/components/builder-page.component.tsx
```diff
@@ -853,8 +853,8 @@ export class BuilderPage extends React.Component<
                         ...this.props.options
                       }}
                       inline={
-                        !Builder.isEditing &&
-                        (this.props.inlineContent || 'content' in this.props)
+                        this.props.inlineContent ||
+                        (!Builder.isEditing && 'content' in this.props)
```
ca136, line 857: great find, how did you know to look here?

mrkoreye: I did a diff of the last web components version that worked with the current one and went through the code to see what could have done it. Plus I was familiar because I fixed a different bug that had to do with `inline`, and I introduced this particular bug with my last fix 😄

PR #178, opened 2020-06-04
add shopify WrapperTag component

teleaziz:


packages/shopify/react/components/WrapperTag.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const head = tags[0];
+    let node = head;
+    let i = 0;
+    while (node) {
+      i++;
+      if (tags[i]) {
+        node.children = [tags[i]];
+      }
+      node.tagName = node.component?.options.tag;
+      // opencondtag, perhaps remove it after second traverse on import
```
mrkoreye, line 48: Can you expand on this comment or remove it? I think it will be confusing when looking at this code in the future.

packages/shopify/react/components/WrapperTag.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const tags = fastClone(validTags) as BuilderElement[];
+
+    if (tags.length === 0) {
+      return <BuilderBlockComponent block={builderBlock} />;
+    }
+
+    const head = tags[0];
+    let node = head;
+    let i = 0;
+    while (node) {
```
mrkoreye, line 42: Any reason to use `while` here instead of a `for` loop or something else?

teleaziz: easier to reason about, since I'm creating a chain of children here

packages/shopify/react/components/WrapperTag.tsx
```diff
@@ -0,0 +1,63 @@
...
+  builderBlock: BuilderElement;
+  conditionalTags: ConditionalTag[];
+  builderState: BuilderStore;
+}
+
+function fastClone<T>(obj: T): T {
+  return JSON.parse(JSON.stringify(obj));
+}
+
+export const WrapperTag: React.FC<WrapperTagProps> = ({
```
steve8708, line 23: would be good to add a quick comment about what this is/does

PR #182, opened 2020-06-09
Add form support

mrkoreye:
This PR adds support for the React SDK to support Shopify `form` tag builder blocks. The docs on the form tag can be found [here](https://shopify.dev/docs/themes/liquid/reference/tags/theme-tags#form). Note: there is some funky logic in the component that looks at a global window variable in order to get the data from the shopify `form` [object](https://shopify.dev/docs/themes/liquid/reference/objects/form). The export step of liquid code generation on the Builder side of things adds a script tag that gets the data from the form object and writes it to the global variable.

### DIFF
--- a/packages/shopify/react/components/form.tsx
+++ b/packages/shopify/react/components/form.tsx
@@ -1,0 +1,146 @@
+import { withBuilder, BuilderStoreContext, BuilderBlockComponent } from '@builder.io/react';
+import * as React from 'react';
+import { FormBlockProps } from '../interfaces/component-props';
+
+export function FormBlock(props: FormBlockProps) {
+  const { type, builderState, parameter, customAttributes, builderBlock, showErrors } = props;
+  let form: any = {};
+  let formAttributes: any = {
+    method: 'post',
+    acceptCharset: 'UTF-8',
+  };
+
+  // This looks for a global variable that we have server rendered by Shopify.
+  // The goal is to get the Shopify form object and allow anything that needs data from
+  // that form object to get it from the Builder state. https://shopify.dev/docs/themes/liquid/reference/objects/form
+  if (window && (window as any).BuilderShopifyData && builderBlock?.id) {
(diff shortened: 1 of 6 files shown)


packages/shopify/react/components/form.tsx
```diff
@@ -0,0 +1,146 @@
...
+      } else if (key && !useValue) {
+        // This is an attribute that does not have a value but should be added to the attribute list (e.g. boolean attribute)
+        formAttributes[key] = true;
+      }
+    });
+  }
+
+  return (
+    <BuilderStoreContext.Consumer>
+      {store => (
```
teleaziz, line 96: Using 
```tsx
<StateProvider state={{form}} > <...></StateProvider>
```
should do the same as this

mrkoreye: Oh, does it work with jsx like the `form` stuff I am doing? Looking at the component I thought it takes a `builderBlock` that has children as a prop and then renders them, but not children elements:
```js
class StateProviderComponent extends React.Component<StateProviderProps> {
  render() {
    return (
      <BuilderStoreContext.Consumer>
        {state => (
          <BuilderStoreContext.Provider
            value={{
              ...state,
              state: {
                ...state.state,
                ...this.props.state
              },
              context: {
                ...state.context,
                ...this.props.context
              }
            }}
          >
            {this.props.builderBlock &&
              this.props.builderBlock.children &&
              this.props.builderBlock.children.map((block, index) => (
                <BuilderBlockComponent
                  block={block}
                  key={block.id}
                  index={index}
                  child={true} 
                />
              ))}
            {this.props.children}
          </BuilderStoreContext.Provider>
        )}
      </BuilderStoreContext.Consumer>
    )
  }
}
```

teleaziz: it doesn't look like it, what you have should work fine

PR #183, opened 2020-06-10
ensure that webcomponents file loads on older browsers

ca136:


steve8708:
I think we need to update how we load s.min.js too right, since it has es6 syntax in it?

steve8708:
(this) https://github.com/BuilderIO/builder/pull/183/files#diff-ff410b3eed8e07bb426eeaa9f35e062fL6

PR #184, opened 2020-06-12
update systemjs

teleaziz:


packages/webcomponents/package-lock.json
```diff
@@ -263,148 +263,6 @@
 			"integrity": "sha512-u5J+m2GSGGGQDKzIBNjaHPLEHJIYDA05kyzWHBxq0qcZfbbVtMVH8x+P2B36YYyYOBWQIIsJjYmNJ05mbmqeWA==",
 			"dev": true
 		},
```
teleaziz, line 265: Does those changes make sense?

steve8708: oh hm, try running `npm run bootstrap` from the root of this repo, it may revert some of this back

also were you able to confirm that the latest version of s.min.js is using es5 with this update?

teleaziz: Yes confirmed

teleaziz: running bootstrap doesn't seem to affect this file.

PR #185, opened 2020-06-17
add variants provider, template and calculate variants

teleaziz:


packages/react/src/components/variants-provider.component.tsx
```diff
@@ -0,0 +1,137 @@
...
+        ></script>
+      </React.Fragment>
+    )
+  }
+
+  const cookieName = `builder.tests.${initialContent.id}`
+
+  const variantId = builder.getCookie(cookieName) || initialContent?.id
+
+  // Q: Calculate variants and set cookie on client side?
```
teleaziz, line 134: otherwise client side routes will have no variant until a full refresh.

PR #186, opened 2020-06-17
Add variants provider, template and calculate variants

teleaziz:


packages/react/src/components/variants-provider.component.tsx
```diff
@@ -0,0 +1,137 @@
...
+        ></script>
+      </React.Fragment>
+    )
+  }
+
+  const cookieName = `builder.tests.${initialContent.id}`
+
+  const variantId = builder.getCookie(cookieName) || initialContent?.id
+
+  // Q: Calculate variants and set cookie on client side?
```
teleaziz, line 134: This is pending, I'll follow up with an update

packages/react/src/components/builder-content.component.tsx
```diff
@@ -227,21 +229,43 @@ export class BuilderContent<
...
-        builder-content-id={useData && useData.id}
-        builder-model={this.props.modelName}
-      >
-        {this.props.children(
-          useData && useData.data,
-          this.props.inline ? false : loading,
-          useData
-        )}
-      </TagName>
+      <VariantsProvider isStatic={this.props.isStatic} initialContent={useData}>
```
teleaziz, line 232: looking on at the tracking calls here, I think I should've added the variants provider on top of the builderContent in BuilderPage... I'll update

packages/react/src/components/variants-provider.component.tsx
```diff
@@ -0,0 +1,159 @@
...
+  delete newData.blocksString
+  return newData
+}
+
+const variantsScript = (variantsString: string, contentId: string) => `
+(function() {
+  if (typeof document.createElement("template").content === 'undefined') {
+    // IE 11
+    return ;
+  }
```
steve8708, line 28: maybe in this case just delete the templates and always show the default? 

otherwise if `template` means nothing they'd see every variation until JS loads right? and then the hydration from react could be broken here too

packages/react/src/components/variants-provider.component.tsx
```diff
@@ -0,0 +1,159 @@
...
+    if (winningTemplate) {
+      var parentNode = winningTemplate.parentNode;
+      while (parentNode.firstChild) {
+        parentNode.firstChild.remove();
+      }
+      parentNode.appendChild(winningTemplate.content.firstChild);
+    }
+  }
+})()
+`
```
steve8708, line 83: for basic compression, may be good to do 

`.replace(/\s+, ' ')` here just to trim newlines and whitespace and save some unnecessary chars

packages/react/src/components/variants-provider.component.tsx
```diff
@@ -77,9 +83,10 @@ const variantsScript = (variantsString: string, contentId: string) => `
       newParent.appendChild(winningTemplate.content.firstChild)
       parentNode.parentNode.replaceChild(newParent, parentNode);
     }
+  } else if (variants.length > 0) {
+    removeVariants()
   }
-})()
-`
+})()`.replace(/\s+/, ' ')
```
steve8708, line 89: mistake in my last comment - we'll want this to have `/g` so it matches all whitespace

```
})()`.replace(/\s+/g, ' ')
```

teleaziz:
server rendering seems busted on the next js example, still trying to find out the reason.

packages/react/src/components/builder-content.component.tsx
```diff
@@ -227,21 +233,49 @@ export class BuilderContent<
...
+                          !this.props.dataOnly && {
+                            ref: (ref: any) => (this.ref = ref)
+                          })}
+                        className="builder-content"
+                        onClick={this.onClick}
+                        builder-content-id={content?.id}
+                        builder-model={this.props.modelName}
+                      >
+                        {this.props.children(
+                          // whhat's going on
```
steve8708, line 265: remove :D

packages/core/src/builder.class.ts
```diff
@@ -1554,10 +1557,20 @@ export class Builder {
...
+          return match
+        }
+
         const matchData = match && match.data;
         if (!matchData) {
           return null;
         }
+
+        if (typeof matchData.blocksString !== 'undefined') {
+          matchData.blocks = JSON.parse(matchData.blocksString);
```
steve8708, line 1570: oof, this should actually be fixed on our backend, will make a note

packages/react/src/components/variants-provider.component.tsx
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  function setCookie(name,value,days) {
+    var expires = "";
+    if (days) {
+        var date = new Date();
+        date.setTime(date.getTime() + (days*24*60*60*1000));
+        expires = "; expires=" + date.toUTCString();
+    }
+    document.cookie = name + "=" + (value || "")  + expires + "; path=/";
```
steve8708, line 45: we should add `; Secure; SamSite=None` here

packages/react/src/components/variants-provider.component.tsx
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  function setCookie(name,value,days) {
+    var expires = "";
+    if (days) {
+        var date = new Date();
+        date.setTime(date.getTime() + (days*24*60*60*1000));
+        expires = "; expires=" + date.toUTCString();
+    }
+    document.cookie = name + "=" + (value || "")  + expires + "; path=/" + "; Secure; SamSite=None";
```
steve8708, line 45: SamSite :D

PR #190, opened 2020-06-26
share state down to symbols

teleaziz:


steve8708:
if I understand this update correctly, we don't want this by design - just like how react components don't auto inherit from parents like ng-controller used to

do you have a use case where you can't pass the state you want as inputs?

teleaziz:
My use case was copying and pasting from page to inside an inline symbol and wanted the bindings to continue to work, this breaks current behavior, closing to think of an alternative

PR #192, opened 2020-06-26
move builder-block templates out of inner html

ca136:


steve8708:
@ca136 remind me - is this still WIP?

ca136:
Yes, there are some issues with the Angular web component bindings I need to resolve before we can merge.

PR #199, opened 2020-07-06
Add filter img url and Srcset logic

mrkoreye:
This PR adds the `img_url` filter to our version of liquidjs, as well as refactors how those filters are added.

Edit: I added in extra support for dynamically generating `srcset` values for shopify urls, as well as support to add `sizes` if none are set and we find a width set on the element.

### DIFF
--- a/packages/shopify/js.js
+++ b/packages/shopify/js.js
@@ -1,1 +1,1 @@
-export * from './js/index.js'
+export * from './js/index.js';
--- a/packages/shopify/js.js
+++ b/packages/shopify/js.js
@@ -1,1 +1,1 @@
-export * from './js/index.js
+export * from './js/index.js'
--- a/packages/shopify/js.js
+++ b/packages/shopify/js.js
@@ -1,1 +1,1 @@
-export * from './js/index.js';
+export * from './js/index.js
--- a/packages/shopify/js/utils/register-liquid-filters.ts
+++ b/packages/shopify/js/utils/register-liquid-filters.ts
@@ -1,0 +1,52 @@
+import { Liquid } from 'liquidjs';
+
+const filters = [
+  {
+    name: 'money',
+    action(value: any) {
+      const str = String(value);
+      return ('$' + str.slice(0, -2) + '.' + str.slice(-2)).replace('..', '.');
+    },
+  },
+  {
+    name: 'handle',
+    action(item: any) {
+      return item
+        .toLowerCase()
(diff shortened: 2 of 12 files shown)


packages/shopify/js/utils/register-liquid-filters.ts
```diff
@@ -0,0 +1,73 @@
...
+    name: 'strip_html',
+    action(item: any) {
+      return item.replace(/<[^>]+>/g, '');
+    },
+  },
+  {
+    name: 'img_url',
+    action(src: any, size: any) {
+      // Taken from the shopify theme script repo
+      // https://github.com/Shopify/theme-scripts/blob/bcfb471f2a57d439e2f964a1bb65b67708cc90c3/packages/theme-images/images.js#L59
```
steve8708, line 30: great find. are these sizes dynamic or set? I would love to update our image component to automatically use srcset based on dynamic or known sizes that shopify cdn creates for images to much better optimize all of our shopify images

will really help our performance which is a big thing we sell on https://github.com/BuilderIO/builder/blob/master/packages/react/src/blocks/Image.tsx#L23:L23

e.g. do you think we can just do url.match(/cdn\.shopify\.com/) and generate srcset with sizes? would be huge if so

mrkoreye: These are set in the template, so they are not dynamic. An example is something like `` which would give you a URL for the image that has a max width of 2048px and max height of 2048px. In theory yes we could just add the shopify cdn and generate the srcset  in the Image.tsx component. I just did some experimenting and it looks like it is dynamic and will allow you to enter arbitrary dimensions (try changing the 250x here to other values: https://cdn.shopify.com/s/files/1/0374/6457/2041/products/valerie-elash-o1Ic6JdypmA-unsplash_250x.jpg?v=1592506853). Want me to add that in to this PR?

steve8708: ooo yes please!

packages/shopify/js.js
```diff
@@ -1 +1 @@
-export * from './js/index.js'
\ No newline at end of file
+export * from './js/index.js
```
mrkoreye, line 1:
```suggestion
export * from './js/index.js'
```

mrkoreye:
@steve8708 updated with changes to allow for shopify `srcset`

steve8708:
killer

PR #200, opened 2020-07-07
pass args to api call, rerender on args change

teleaziz:


packages/shopify/react/functions/serialize-liquid-args.ts
```diff
@@ -0,0 +1,16 @@
+export function serializeLiquidArgs(data?: Record<string, any>) {
+  const argStrings: string[] = [];
+  if (data) {
+    for (const key in data) {
+      const value = data[key];
+      // Key must be just letters, numbers, _
+      if (/^[a-z_0-9]+$/i.test(key)) {
+        // For now just support boolean, number, string
+        if (['boolean', 'number', 'string'].includes(typeof value)) {
+          argStrings.push(`${key}: ${JSON.stringify(value)}`);
```
steve8708, line 10: something I should have done before but only realized from another thread, it would be great to do 

```ts
JSON.stringify(typeof value === 'string' ? value.replace(/"/g, '&quot;') : value)
```

PR #204, opened 2020-07-13
Fix proper fallback when no image size can be inferred from a Shopify image URL

mandx:


steve8708:
Since this is a pretty critical fix and I want to get some QA on it for release in the AM - merging eagerly, but @mrkoreye please don't hesitate to still send feedback we can followup with updates on

mrkoreye:
Yes, so this fix is great. We definitely do not want to be writing the url as `null`. Looking into the issue more, it looks like because the regex we use to determine if we can modify the size is specific and only allows one query param (`v`), it fails when the image url has another query param (e.g. `https://cdn.shopify.com/s/files/1/0061/0608/7522/products/DSC_5084.jpg?v=1593441233&width=500`). I got the regex from shopify code (https://github.com/Shopify/theme-scripts/blob/bcfb471f2a57d439e2f964a1bb65b67708cc90c3/packages/theme-images/images.js#L68), so I am not sure we want to modify, but if we remove the `$` we should be able to account for this case as well, allowing us to render the proper sizes for urls like this. 

So the question is, do we want to add that fix in as well?

steve8708:
oh funny, that's likely because in this case we are rendering the URLs and maybe I added something one day that adds that param tho I think it is a no-op

tbh I think removing the $ works and is a good idea. most people probably render shopify URLs exactly as provided, with no extra query params as they support none, but I think the $ in this case is excessively restrictive albeit it may be very rare anyone is adding query params to URLs

for superkin tho, it would be nice to optimize their images despite the noop query param, so I am in favor of just removing the $ form the URL. E.g. by checking cdn.shopify.com in there too (if not already checked higher up in the logic)

```ts
src.match(/cdn\.shopify\.com.+?\.(jpg|jpeg|gif|png|bmp|bitmap|tiff|tif)(\?v=\d+)?$/i)
```

mrkoreye:
`cdn.shopify.com` is already checked above, so we should be good with that. Just checking, but in your regex, it still has `$` at the end. You are down with removing it, correct? I can send follow up PR

steve8708:
oops haha yeah I'm up for removing that despite my example

steve8708:
I have to make some other local changes so I can add this right now real quick

mandx:
If we are going to modify the regex to guess image sizes, I would strongly suggest to first properly parse the URL, and use regexes on the `path` component only. Using regexes on the entire URL would be extremely brittle, while operating only on the path should be straightforward. Also, the querystring would be parsed as well (as an object I guess) we can easily inspect that as well if needed.

mrkoreye:
@mandx totally fair point. Do you want to send a follow up PR to change that behavior?

steve8708:
Since this code runs client side on render, performance is a very high priority, and the regex is going to be a lot faster than a url parse and regex ([almost a 10x difference](https://jsperf.com/url-parse-and-regex-vs-just-regex)). `new URL` also doesn't have full browser support so could depend on loading a library if we don't already

In the other PR where this code runs at compile time I totally agree with this comment, but in the case of at runtime I'm more concerned about performance than URLs that contain 'cdn.shopify.com' and '.jpg' in places other than their host and path, which seems like a very unlikely event I'd hate to add runtime cost for when I can't conceive of a common enough case this would happen

PR #207, opened 2020-07-17
Update to allow repeats on custom code blocks

mrkoreye:
This change makes it so we can use a custom code block with `replaceNodes` set to true within a repeated element. An example is putting in judge.me reviews on product cells in a collection page:

Screen Shot 2020-07-17 at 4 39 26 PM


steve8708:
killer

PR #210, opened 2020-07-20
staleCacheSeconds param for getting content

steve8708:


packages/core/src/builder.class.ts
```diff
@@ -1825,6 +1826,10 @@ export class Builder {
           queryParams.cacheSeconds = options.cacheSeconds;
         }
 
+        if (typeof options.staleCacheSeconds === 'number') {
```
mandx: Could we also check for valid number values? Like, `isNaN`, `>= 0`, etc.

steve8708:
@mandx updated

PR #218, opened 2020-07-21
parse page param to fix next urls for shopify

ca136:
_Screenshot_
image


packages/shopify/react/components/Paginate.tsx
```diff
@@ -31,7 +31,7 @@ export function PaginateBlock(props: AssignBlockProps) {
     let currentPage = 1;
     if (builderState.state.location.query.page) {
       // this is undefined, should probably add support to query params on location
-      currentPage = builderState.state.location.query.page;
+      currentPage = parseInt(builderState.state.location.query.page, 10);
```
mrkoreye, line 34: If the query param is not a number, `currentPage` will be `NaN`. Maybe change this to:
```suggestion
      currentPage = parseInt(builderState.state.location.query.page, 10) || currentPage;
```

PR #223, opened 2020-07-24
Lazy load videos

steve8708:


packages/react/src/blocks/Image.tsx
```diff
@@ -160,6 +160,15 @@ class ImageComponent extends React.Component<any> {
...
   pictureRef: HTMLPictureElement | null = null;
 
+  scrollListener: null | ((e: Event) => void) = null;
+
+  componentWillUnmount() {
+    if (Builder.isBrowser && this.scrollListener) {
+      window.removeEventListener('scroll', this.scrollListener);
+      this.scrollListener = null;
+    }
+  }
```
steve8708, line 170: this now also ensures we cleanup image scroll listeners on unmount too

packages/react/src/blocks/Video.tsx
```diff
@@ -37,25 +52,62 @@ class VideoComponent extends React.Component<{
...
   componentDidMount() {
     this.updateVideo();
+
+    if (this.lazyLoad && Builder.isBrowser) {
+      // TODO: have a way to consolidate all listeners into one timer
+      // to avoid excessive reflows
+      const listener = throttle(
+        (event: Event) => {
+          if (this.containerRef) {
+            const rect = this.containerRef.getBoundingClientRect();
```
mrkoreye, line 62: Do we support IE11? If not, we could use `IntersectionObserver` for this sort of logic: https://caniuse.com/#feat=intersectionobserver

steve8708: great idea

steve8708: I was hating the fact of how many timers we are creating that on fire trigger a reflow each so this would be a huge improvement

PR #230, opened 2020-08-03
ensure we fetch shopify sections if missing contents

ca136:
Not sure if this is the best fix, but the issue is that the SDK is rendering empty divs for shopify sections, and the `SectionRef` component was seeing those divs and deciding not to fetch the contents:
image

fixes: https://app.asana.com/0/1166097513754419/1186760367446164

### DIFF
--- a/packages/shopify/react/components/SectionRef.tsx
+++ b/packages/shopify/react/components/SectionRef.tsx
@@ -12,7 +12,9 @@ const cache: { [key: string]: string } = {};
 const refs =
   Builder.isBrowser &&
   Array.from(document.querySelectorAll('[builder-shopify-section]')).reduce((memo, item) => {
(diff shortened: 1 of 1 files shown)


packages/shopify/react/components/SectionRef.tsx
```diff
@@ -43,15 +45,19 @@ export const SectionRef = (props: SectionRefProps) => {
...
         )}`
       )
         .then(res => res.text())
         .then(text => {
-          cache[sectionName] = text;
-          setHtml(text);
+          // to ensure we don't keep fetching content for empty sections, we need to insert something
+          const safeText = text || '<span></span>';
+          cache[sectionName] = safeText;
+          setHtml(safeText);
```
ca136, line 60: @mrkoreye @steve8708 one minor update here to handle empty sections

PR #231, opened 2020-08-04
wrap builder element in a div before rendering React

steve8708:
this is so the <builder-component> tag is not lost, causing various bugs


.prettierignore
```diff
@@ -0,0 +1,7 @@
+coverage
+node_modules
+dist
+.next
+.cache-loader
+.cache
+docs
```
teleaziz, line 7: eof

Issue #234, opened 2020-08-10
[angular] Could not make an angular element - multiple tries to create custom element


lyczos:
Inside `BuilderModule` constructor - there is a code which trying to create angular element even if it already exists, that's why we are getting that issue:

![image](https://user-images.githubusercontent.com/22659410/89790225-ea540500-db21-11ea-875c-5382a91996c4.png)

Whole logic should be moved to another place (`forRoot` method). And fired only once on first init.
Right now, wherever we try to use `BuilderModule` with lazy loaded modules we will end up with this error.

I will add hot-fix as a part of today PR but I think it should be refactored and moved to separate place - outside `constructor`

steve8708:
Thanks @lyczos ! Closing via #235 being merged

PR #235, opened 2020-08-10
fix(angular): fix multiple tries to create custom element

lyczos:
## Description
fixing #234

_Screenshot_
image


packages/angular/src/app/modules/builder/builder.module.ts
```diff
@@ -29,7 +29,10 @@ export class BuilderModule {
       // This cannot use a normal import, via https://github.com/angular/angular/issues/24551
       const { createCustomElement } = require('@angular/elements');
```
lyczos, line 30: this whole logic should be moved to `forRoot` method to avoid multiple creation

steve8708:
Thanks @lyczos! Published as `@builder.io/angular@1.1.15-alpha.1`

steve8708:
actually @lyczos use this version, has one more bug fix in it :) `@builder.io/angular@1.1.15-alpha.2`

PR #236, opened 2020-08-10
remove subscription to user attributes

teleaziz:
This was caused by fixing subscription first run, and now it's causing a max stack exception


packages/angular/src/app/modules/builder/components/builder-component/builder-component.component.ts
```diff
@@ -149,7 +149,6 @@ export class BuilderComponentComponent implements OnDestroy, OnInit {
         builder.apiKey = this.builderService.apiKey;
         builder.canTrack = this.builderService.canTrack;
         builder.setUserAttributes(omit(this.builderService.getUserAttributes(), 'urlPath'));
-        builder.userAttributesChanged.subscribe((attrs) => builder.setUserAttributes(attrs));
```
steve8708, line 152: oops, this is supposed to be here, but should be

```ts
        this.builderService.userAttributesChanged.subscribe((attrs) => builder.setUserAttributes(attrs));	
```

teleaziz: that did fix the exception

PR #237, opened 2020-08-10
include cookies in subdomains

teleaziz:


packages/core/src/builder.class.ts
```diff
@@ -91,12 +91,14 @@ function setCookie(name: string, value: string, expires?: Date) {
     }
 
     const secure = isBrowser ? location.protocol === 'https:' : true;
+    const nakedDomain = location.hostname.split('.').slice(-2).join('.');
```
steve8708, line 94: this will break for `.co.uk` etc I think?

teleaziz: true, I'll go the other way about it, omitting only the first sub-domain

mrkoreye: Do we also need to modify the `this.setCookie` method to include domain for when the cookie is set via the `Cookies` class on the server? https://github.com/BuilderIO/builder/blob/master/packages/core/src/builder.class.ts#L2021

teleaziz: yes @mrkoreye , already modified here https://github.com/BuilderIO/builder/pull/237/files#diff-5ed22976e6209bba5c2b4da86850f86dR59

mrkoreye: oh lol how did i miss that 😆

PR #239, opened 2020-08-12
rely on updated_at when tracking shopify conversion

teleaziz:


packages/shopify/track/track.ts
```diff
@@ -40,9 +40,9 @@ if (!_window[TRACKED_KEY]) {
   } else if (!Shopify) {
     console.debug('No Shopify object');
   } else if (Shopify.checkout?.order_id) {
-    const ordderCreatedDate = new Date(Shopify.checkout.created_at);
+    const orderUpdatedAt = new Date(Shopify.checkout.updated_at);
 
-    const orderCreatedMinutesAgo = (Date.now() - ordderCreatedDate.getTime()) / 1000 / 60;
+    const orderCreatedMinutesAgo = (Date.now() - orderUpdatedAt.getTime()) / 1000 / 60;
```
mrkoreye, line 45: since this is not the order created time but now updated time can we change the variable name to `orderUpdatedMinutesAgo`?

PR #243, opened 2020-08-18
Also use server-side rendered state when rendering symbols

mandx:
When rendering on the server, new state generated by running symbol's JS on the server ends up in a different location. This changes makes `Symbol` blocks to _also_ use the state data in that location when rendering.

Currently this data takes precedence, that is, it will override existing keys (for example, input data). I believe this should be fine, as we want the symbol JS code to have the last word on the shape of the symbol's state.

### DIFF
--- a/packages/react/src/blocks/Symbol.tsx
+++ b/packages/react/src/blocks/Symbol.tsx
@@ -136,7 +136,11 @@ class SymbolComponent extends React.Component<SymbolProps> {
                   context={{ ...state.context }}
                   modelName={model}
                   entry={entry}
-                  data={this.props.inheritState ? { ...data, ...state.state } : data}
+                  data={{
+                    ...data,
+                    ...(!!this.props.inheritState && state),
+                    ...this.props.builderBlock?.component?.options?.symbol?.content?.data?.state,
+                  }}
                   inlineContent={symbol?.inline}
                   {...(content && { content })}
                   options={{ key }}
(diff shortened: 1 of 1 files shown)


packages/react/src/blocks/Symbol.tsx
```diff
@@ -136,7 +136,11 @@ class SymbolComponent extends React.Component<SymbolProps> {
                   context={{ ...state.context }}
                   modelName={model}
                   entry={entry}
-                  data={this.props.inheritState ? { ...data, ...state.state } : data}
+                  data={{
+                    ...data,
+                    ...(!!this.props.inheritState && state),
```
steve8708, line 141: I think this needs to be `state.state`

packages/react/src/blocks/Symbol.tsx
```diff
@@ -136,7 +136,11 @@ class SymbolComponent extends React.Component<SymbolProps> {
...
+                    ...this.props.builderBlock?.component?.options?.symbol?.content?.data?.state,
```
steve8708, line 142: 👍

packages/react/src/blocks/Symbol.tsx
steve8708: (bad name I know, since `state` refers to the react internal context state, which has `context` which is the Builder context, and `state` which is the Builder state (and other stuff like `content` that is Builder content, etc))

PR #250, opened 2020-08-27
getAll method for SDK

steve8708:


packages/core/src/builder.class.ts
```diff
@@ -2033,6 +2033,10 @@ export class Builder {
     if (!this.apiKey) {
       throw new Error('Builder needs to be initialized with an API key!');
     }
-    return this.queueGetContent(modelName);
+    return this.queueGetContent(modelName, options);
+  }
+
+  getAll(modelName: string, options: GetContentOptions = {}) {
```
mandx, line 2039: Does the user still needs to add a `.promise()` to this call to be used with `await`?

mandx: Also how about pagination?

steve8708: ah that's a good idea to .promise() by default for this

for pagination this supports the `offset` and `limit` params

mandx: Since this is public API, can I suggest to add return types to this method?

steve8708: donzeo

PR #251, opened 2020-08-27
pass content options in the getContent method

ca136:
We've had a few people request a method in the SDK to fetch all pages for a model. The current `builder.get` method only fetches the first match. You can currently use something like `builder.queueGetContent('page', {limit: 50})`, but this change fixes the `getContent` method to pass through the options. It's slightly more developer friendly than `queueGetContent`.


ca136:
fixed here: https://github.com/BuilderIO/builder/pull/250

PR #263, opened 2020-09-12
support `getAll` server side

steve8708:


packages/core/src/builder.class.ts
```diff
@@ -2077,10 +2077,28 @@ export class Builder {
...
+      res?: ServerResponse;
+      apiKey?: string;
+    } = {}
+  ): Promise<BuilderContent[]> {
+    let instance: Builder = this;
+    if (!Builder.isBrowser && (options.req || options.res)) {
+      instance = new Builder(options.apiKey || this.apiKey, options.req, options.res);
+    } else {
+      if (options.apiKey && !this.apiKey) {
+        this.apiKey = options.apiKey;
```
mandx, line 2093: Shouldn't be
```suggestion
        instance.apiKey = options.apiKey;
```

steve8708: when the above condition isn't met (`!Builder.isBrowser ... `) `instance === this`

PR #264, opened 2020-09-15
More complete ecom design system demo

steve8708:
This is a more complete example demo for our react-design-system focused on ecomerce and featuring landing pages as well as components like collection page heros, cart upsells, announcement bars, product info, etc

Also breaks the storybook stuff into it's own example folder and removes it from the design system example


steve8708:
this is purely example code and not SDK / runtime code, so merging to add some follow updates, happy to come back with updates based on feedback as well

PR #265, opened 2020-09-15
fix: support ssr rendering

ocombe:
## Description

This block of code is throwing an error on the server side with universal because it says that you cannot redefine `HTMLElement.prototype` as it is read only.
Adding the try catch allows me to use BuilderIo on the server side (by using `prerender` on your servers).


steve8708:
this is great, thanks @ocombe !

PR #268, opened 2020-09-15
fix replaceState typo

steve8708:


ca136 (approved):

ca136:
Looks like it needs a prettier update

steve8708:
weird, for some reason we circle seems to just ignore some PRs, will be great to move to github actions here sometime when time permits

steve8708:
merging for now since we need this in and the CI output is not making any sense and is just lint related anyway. will come back to the lint issue

PR #274, opened 2020-09-22
fix impression tracking

steve8708:
I believe our issues was accidentally sending `context` which includes the entire `content` document (huge JSON tree) as `customProperties` (which go into our query string, logs, and DB) causing some area to fail for tracking - either at the S3 level (failure to log the entire request being too large) or at the parsing level (long request maybe gets truncated and events JSON could not be parsed) or at the saving level (bigQuery string field max length exceeded)


ca136 (approved):
Oh great find, I noticed the content object in the request and thought it was intentional.

PR #277, opened 2020-09-26
fix(sdk): fix statement in getLocation()

lyczos:
## Issue
When the app is rendered on serverside the `urlPath` in API request is wrong - always `/` (SSR with mocked window/location)

## Description
When using SSR there is no need to check the `location` object 'cause if someone will mock `window`/`location` on the server side then `parsedLocation` will be overwritten by the second statement - which IMO shouldn't happen.

_Screenshot_
Behavior on SSR before the change(when  `window`/`location` are mocked):
image

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -1501,10 +1501,7 @@ export class Builder {
     // in ssr mode
     if (this.request) {
       parsedLocation = parse(this.request.url);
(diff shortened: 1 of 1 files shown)


ocombe (approved):

steve8708:
published the above with `@builder.io/angular@1.1.20-alpha.0`

PR #281, opened 2020-09-29
unique cache keys for getAll with different options passed

steve8708:
to avoid bugs of getting the wrong results

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -2106,7 +2106,7 @@ export class Builder {
           options.key ||
           // Make the key include all options so we don't reuse cache for the same conent fetched
           // with different options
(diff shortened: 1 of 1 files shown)


mrkoreye (approved):

packages/core/src/builder.class.ts
```diff
@@ -2101,6 +2102,12 @@ export class Builder {
       .getContent(modelName, {
         limit: 30,
         ...options,
+        key:
+          options.key ||
+          // Make the key include all options so we don't reuse cache for the same conent fetched
+          // with different options
+          (Builder.isBrowser && hash(omit(options, 'initialContent', 'req', 'req'))) ||
```
mandx, line 2109:
```suggestion
          (Builder.isBrowser && hash(omit(options, 'initialContent', 'req', 'res'))) ||
```

steve8708: 🙏

PR #285, opened 2020-10-05
allow extending of trusted hosts

teleaziz:


packages/core/src/builder.class.ts
```diff
@@ -1263,15 +1271,11 @@ export class Builder {
         const url = parse(event.origin);
         const isRestricted =
           ['builder.register', 'builder.registerComponent'].indexOf(event.data?.type) === -1;
-        if (
-          isRestricted &&
-          !(
-            url.hostname &&
-            (url.hostname === 'builder.io' ||
-              url.hostname.endsWith('.builder.io') ||
```
steve8708, line 1271: this update negates this piece of code - we want to support **.builder.io, not just builder.io

packages/core/src/builder.class.ts
```diff
@@ -422,6 +422,7 @@ export class Builder {
   authToken = '';
 
   static editors: any[] = [];
+  static trustedHosts: string[] = ['builder.io', 'localhost:1234'];
```
steve8708, line 425:
```suggestion
  static trustedHosts: string[] = ['builder.io', 'localhost'];
```
this check is against `hostname` which excludes the port (ant not `host` which includes it)

packages/core/src/builder.class.ts
teleaziz: updated to use `isTrusted(host)` which only check if the `builder.io` is in the host string

packages/core/src/builder.class.ts
```diff
@@ -489,6 +491,14 @@ export class Builder {
     this.actions.push(action);
   }
 
+  static registerTrustedHost(host: string) {
+    this.trustedHosts.push(host);
+  }
+
+  static isTrusted(host: string) {
```
steve8708, line 498:
```suggestion
  static isTrustedHost(host: string) {
```

nit pick, just a little more consistent and clear imo

packages/core/src/builder.class.ts
steve8708: 👍

packages/core/src/builder.class.ts
```diff
@@ -489,6 +491,14 @@ export class Builder {
...
+    return this.trustedHosts.findIndex(allowedHost => host.indexOf(allowedHost) > -1) > -1;
```
steve8708, line 499: theres still an issue here, someone could register

maliciousbuilder.io and it works

or builder.io.malicious.com

it's important for host matching that the host is either exact, or `*.${host}`

teleaziz: make sense, I'll update 👍

steve8708 (approved):

PR #286, opened 2020-10-08
fix re-assigning of default test group

teleaziz:


packages/react/src/components/variants-provider.component.tsx
```diff
@@ -56,7 +56,7 @@ const variantsScript = (variantsString: string, contentId: string) =>
   }
   var cookieName = 'builder.tests.${contentId}';
   var variantInCookie = getCookie(cookieName);
-  var availableIDs = variants.map(function(vr) { return vr.id });
+  var availableIDs = variants.map(function(vr) { return vr.id }).concat("${contentId}");
```
ca136, line 59: nit: looks like we use single quotes in most places, is prettier not formatting this?
```suggestion
  var availableIDs = variants.map(function(vr) { return vr.id }).concat('${contentId}');
```

ca136 (approved):

packages/react/src/components/variants-provider.component.tsx
teleaziz: This is an interpolated string that we print in ssr

PR #287, opened 2020-10-09
allow overriding of session id through url

teleaziz:


packages/shopify/track/track.ts
```diff
@@ -35,7 +35,9 @@ setTimeout(() => {
 
     // Allow passing a session ID to the tracking script to support cross domain converison tracking
     // AKA: <script src="https://cdn.builder.io/js/shopify/track?apiKey=YOUR_KEY&sessionId={{checkout.attributes._builderSessionId}}">
-    const sessionIdParam = getQueryParam(
+    // OR: append session ID to your checkout URL
+    const overridSessionId = getQueryParam(location.href, 'sessionId');
```
steve8708, line 39: while I'm not 100% sure this will work as expected (may still just not be knowing the full flow), if this is read from the page URL I would at least namesapce it

```suggestion
    const overridSessionId = getQueryParam(location.href, 'builderSessionId');
```

steve8708 (approved):

PR #288, opened 2020-10-09
fix running scripts of application/json type

teleaziz:


steve8708 (approved):
good catches, which issues does this fix?

PR #289, opened 2020-10-12
add "builder:load" and "connect" events to webcomponents

steve8708:


packages/webcomponents/src/elements.ts
```diff
@@ -254,6 +260,10 @@ if (Builder.isBrowser && !customElements.get(componentName)) {
       }
       this.connected = true;
 
+      this.dispatchEvent(
+        new CustomEvent('connected', { detail: { builder, Builder }, bubbles: true })
```
teleaziz, line 264: should this be `builder:connected`?

steve8708: I was torn on that. leaned toward calling the per-element event 'connected' since we also fire 'load' and 'htmlload' events w/o namespace, and just saving the namespace for events we fire globally (e.g. the new one on window)

so if you do addEventListener('connected') you can always check that e.target.tagName.toLowerCase() === 'builder-component' if you are listening on a parent

teleaziz (approved):

PR #309, opened 2020-11-19
AOT compilation (codegen) support for react using jsx-lite

steve8708:


packages/react/src/blocks/Columns.tsx
```diff
@@ -152,6 +154,20 @@ class ColumnsComponent extends React.Component<any> {
   }
 }
 
+const Column = (props: any) => (
+  <div
+    {...props.attributes}
+    style={{ width: '50%', ...props.attributes?.style, ...props.style }}
```
steve8708, line 160:
```suggestion
    style={{ width: `${props.width}%`, ...props.attributes?.style, ...props.style }}
```

steve8708:
moving to #344

PR #317, opened 2020-12-15
Cleanup warnings & update method to fetch the slug

saurabhbhatia:
## Description

This pull request updates 2 things in the Next JS example - 

1. Cleans up warning for the prefetch syntax update.

```
event - compiled successfully
Next.js auto-prefetches automatically based on viewport. The prefetch attribute is no longer needed. More: https://err.sh/vercel/next.js/prefetch-true-deprecated
Next.js auto-prefetches automatically based on viewport. The prefetch attribute is no longer needed. More: https://err.sh/vercel/next.js/prefetch-true-deprecated
```

2. Update how we fetch slug from the dynamic route generated by Next JS. The newer versions of next js make the dynamic slug available as a param attribute in the `getServerSideProps`. The updated code, makes use of this as opposed to reading from the url. Following is what we get when we to fetch `params` in `getServerSideProps` :

```
{ slug: [ 'dynamic-page-slug' ] }
```

### DIFF
--- a/examples/next-js/components/nav.js
+++ b/examples/next-js/components/nav.js
@@ -13,7 +13,7 @@ const Nav = () => (
   <nav>
     <ul>
       <li>
-        <Link prefetch href="/">
+        <Link href="/">
           <a>Home</a>
         </Link>
       </li>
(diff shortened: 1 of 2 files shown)


teleaziz (approved):
LGTM, thanks for the PR @saurabhbhatia

PR #322, opened 2020-12-29
use post on admin graphql client

teleaziz:


packages/cli/src/admin-sdk.ts
```diff
@@ -22,7 +22,9 @@ const root = 'https://qa.builder.io';
 const createGraphqlClient = (privateKey: string) =>
   createClient({
     fetcher: ({ query, variables }, fetch, qs) =>
-      fetch(`${root}/api/v2/admin?${qs.stringify({ query, variables })}`, {
+      fetch(`${root}/api/v2/admin`, {
```
mandx, line 25: Does the endpoint still accept `GET` requests or is that being removed?

mandx (approved):

packages/cli/src/admin-sdk.ts
teleaziz: Yes it does accept GET, no restrictions, the reason for this change is the URL was getting huge when sending mutations

PR #324, opened 2021-01-04
Update React@17.0.1

antonhalim:
Hi Builder Team 👋🏼 

## Description

React recently upgrade to the latest [17.0.1](https://github.com/facebook/react/blob/master/CHANGELOG.md#1701-october-22-2020), despite the major version, this seems to be a safe upgrade since it doesn't introduce any difficult breaking changes.

I'd like to install and use a single version of React if possible, and builder.io/react is one of our sub-dependency that's is on 16.x.

First time doing this here, so lmk if there is much more process here, thanks!

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -110,7 +110,7 @@
     "lint-staged": "^7.0.0",
     "prettier": "^1.4.4",
     "prompt": "^1.0.0",
-    "react-test-renderer": "^16.13.1",
+    "react-test-renderer": "^16.13 || ^17.0.1",
     "replace-in-file": "^3.0.0-beta.2",
     "rimraf": "^2.6.1",
     "rollup": "^1.32.1",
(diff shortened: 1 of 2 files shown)


packages/react/package.json
```diff
@@ -110,7 +110,7 @@
     "lint-staged": "^7.0.0",
     "prettier": "^1.4.4",
     "prompt": "^1.0.0",
-    "react-test-renderer": "^16.13.1",
+    "react-test-renderer": "^17.0.1",
```
mandx, line 113: Hi @antonhalim, sorry for the late reply! Just one thing, what do you think of relaxing the version a bit, like, allowing React v16 _or_ v17? Something like:
```suggestion
    "react-test-renderer": "^16.13 || ^17.0.1",
```
My point is that we don't want to force folks to upgrade to React 17 just to get features/fixes they need after this change is made.

antonhalim: Oh yeah, for sure, thanks!

antonhalim:
@mandx thanks, whenever you're ready

mandx (dismissed):

antonhalim:
@mandx sounds good, didnt know it exists

mandx (approved):

PR #329, opened 2021-01-15
utils: fix build and add a method to target specific props and extend

teleaziz:


packages/utils/README.md
```diff
@@ -50,4 +50,100 @@ export async function getStaticProps(context) {
...
+```
+This component will update to fetch product data browser side by query chosen in the Builder editor. But if you want this to render server side you need `props.products` to be passed with the needed data on render.
+
+To accomplish this, you can use `extendAsyncProps`. 
+
+Here is a usage example with Next.js:
+
+```tsx
+import { Builder, builder } from '@builder.io/react';
+import { extendAsyncProps } from '@builder.io/utils';
```
mandx, line 130: In this example, this is imported, but not used?

packages/utils/README.md
```diff
@@ -50,4 +50,100 @@ export async function getStaticProps(context) {
...
+
+export default function MyPage(props) {
+  return <BuilderComponent model="page" content={props.content} />
+}
+
+export async function getStaticProps(context) {
+  const content = await builder.get('page', { url: context.resolvedUrl }).promise();
+  
+  await getAsyncProps(content, {
```
mandx, line 139:
```suggestion
  await extendAsyncProps(content, {
```
Is this what you mean?

packages/utils/README.md
```diff
@@ -50,4 +50,100 @@ export async function getStaticProps(context) {
 
   return { props: { content } }
 }
+```
+
+## extendAsyncProps
+
+`extendAsyncProps` is similar to `getAsyncProps` except it look up the inputs by name across components and map it's props to the result of the asynchronus mapper function passed.
```
mandx, line 57:
```suggestion
`extendAsyncProps` is similar to `getAsyncProps` except it looks up the inputs by name across components and map its props to the result of the asynchronus mapper function passed.
```
Also, maybe reword this in simpler terms?

> map its props to the result of the asynchronus mapper function passed.

It might takes a few re-reads to understand properly...

packages/utils/README.md
```diff
@@ -50,4 +50,100 @@ export async function getStaticProps(context) {
...
\ No newline at end of file
+```
+
+## extendAsyncProps
+
+`extendAsyncProps` is similar to `getAsyncProps` except it looks up the inputs by name across components and map its props to the result of the asynchronus mapper function passed.
+
+### Example
+
+Say you have multiple components that takes a query input and you want to resolve this query asynchronusly before rendering on the server
```
mrkoreye, line 61:
```suggestion
Say you have multiple components that take a query input and you want to resolve this query asynchronusly before rendering on the server
```

mrkoreye (approved):

PR #331, opened 2021-01-20
Allow the usage of targeting attributes in the url

tonytamps:
## Description

I wanted to use my targeting attributes in the url of the `dynamic-dropdown` but they weren't available.
image

I've stumbled my way through an implementation that looks like it does what I want and was going to publish my own plugin but realised this probably doesn't warrant a standalone plugin but would be better to enhance the existing one.

This allows me to use targeting properties in the `dynamic-dropdown` url like this:
```js
{
    name: "dropdown",
    type: "dynamic-dropdown",
    options: {
        url: `https://example.com/api/v1/users?locale={{targeting.locale.0}}`,
        mapper: mapper.toString(),
    },
},
```

I'd be keen to hear any feedback you've got to improve this.

_Screenshot_
(no visual changes)


teleaziz (approved):
LGTM, thanks for the PR @tonytamps

PR #335, opened 2021-01-22
Builder.registerComponent -> Builder.registerBLock

steve8708:
I'm still torn on this one, but want to provide us with the option

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -635,7 +635,10 @@ export class Builder {
     };
   }
 
-  // static registerComponent(...) { .. }
+  static registerBlock(component: any, options: Component) {
+    this.registerComponent(component, options)
+  }
+
   static registerComponent(component: any, options: Component) {
     const spec = {
       class: component,
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -636,7 +636,7 @@ export class Builder {
   }
 
   static registerBlock(component: any, options: Component) {
-    this.registerComponent(component, options)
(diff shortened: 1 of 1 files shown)


packages/core/src/builder.class.ts
```diff
@@ -635,7 +635,10 @@ export class Builder {
     };
   }
 
-  // static registerComponent(...) { .. }
+  static registerBlock(component: any, options: Component) {
+    this.registerComponent(component, options);
+  }
+
```
mandx, line 641: If this is definitely the direction we want to go, we could attach a deprecation notice to `registerComponent`
```suggestion

  /**
   * @deprecated Use `registerBlock` instead.
   */
```
And if you like you can add more text on why is it being deprecated.

Optionally we could also move the implementation (the method body) here, and have `registerComponent` simply call `registerBlock`.

mandx (commented):
> I'm still torn on this one

So the issue is just the naming? In any case, this doesn't look like a breaking change...

Are we planning to repurpose the name `registerComponent` for something else in the future?

packages/core/src/builder.class.ts
steve8708: as noted in the PR, this is tentative, so no deprecation today is desired. creating an alias just affords us to go the direction, based on where we decide. 

aka if we decided _today_ that we want to standardize on Builder.registerBlock - we can't just go and update our docs accordingly. we have to update hte SDKs, wait for the latest major verisons to go out (takes time and has dependencies), then we can update. it takes a while and sometimes we need to make changes rapidly

so what I propose is add the alias to the code to give us the option to go this direciton, but not force it yet

PR #340, opened 2021-01-25
fix utils package missing index after publish

### LINKED ISSUE #339: v1.0.1 missing index.js
Hi,

The package.json `main` property is set to `/dist/index.js` and when you build locally there is a `/dist/index.js`.
https://github.com/BuilderIO/builder/blob/master/packages/utils/package.json#L5

However if you install version 1.0.1 there is no `/dist/index.js` 

You can inspect the currently published package here - https://unpkg.com/@builder.io/utils@1.0.1 which will redirects to a page saying:

> Cannot find "/dist/index.js" in @builder.io/utils@1.0.1

If you inspect the file tree for the locally installed package you can also see that it is not present for `1.0.1`
image

My current workaround is directly importing like
`import { getAsyncProps } from "@builder.io/utils/dist/get-async-props.js";`

but this seems a bit brittle and is not guaranteed to work in the long term

teleaziz:
closes #339

### DIFF
--- a/packages/utils/.gitignore
+++ b/packages/utils/.gitignore
@@ -1,4 +1,2 @@
-index.js
-*.d.ts
 src/*.js
 src/*.d.ts
--- a/packages/utils/package-lock.json
+++ b/packages/utils/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@builder.io/utils",
-  "version": "1.0.1",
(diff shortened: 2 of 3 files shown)


mrkoreye (approved):

steve8708:
oops meant to hit approved

PR #353, opened 2021-01-29
Fix empty key binding

mrkoreye:
## Description
This PR adds logic to double check that there is a `key` property when setting bindings and actions. Here is a related forum post: https://forum.builder.io/t/site-not-showing-up/418


steve8708 (approved):

steve8708:
Worth considering we want to do !key.trim?.() in case a string that is empty space also could cause this error

steve8708 (approved):

PR #356, opened 2021-01-29
full fix for content prop

steve8708:
one more piece was missing


mandx (approved):

steve8708:
cc @teleaziz fyi found one more usage that was needed

PR #362, opened 2021-02-04
include spaces site url in editing url template

teleaziz:
this allows for using the editin url as a template of the space site url similar to this https://github.com/BuilderIO/nextjs-shopify/commit/d548f8d9ea2a2e3a6ec3ad8340710860f75db724

### DIFF
--- a/plugins/shopify/src/plugin.tsx
+++ b/plugins/shopify/src/plugin.tsx
@@ -85,10 +85,12 @@ Builder.register('plugin', {
       await Promise.all(promises);
 
       // import and register webhooks
+      appState.globalState.showGlobalBlockingLoadingIndicator = true;
       await importProducts('shopify-product');
       await createWebhooks('product', 'shopify-product');
       await importCollections('shopify-collection');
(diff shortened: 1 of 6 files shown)


plugins/shopify/src/plugin.tsx
```diff
@@ -85,10 +85,12 @@ Builder.register('plugin', {
       await Promise.all(promises);
 
       // import and register webhooks
+      appState.globalState.showGlobalBlockingLoadingIndicator = true;
       await importProducts('shopify-product');
       await createWebhooks('product', 'shopify-product');
       await importCollections('shopify-collection');
       await createWebhooks('collection', 'shopify-collection');
+      appState.globalState.showGlobalBlockingLoadingIndicator = false;
```
steve8708, line 93: might be good to handle errors here even with just a modal saying contact our support, or else there could be cases of infinite spinners

steve8708 (approved):

steve8708 (approved):

PR #367, opened 2021-02-05
Don't use variations if DNT is set or tracking is disallowed

mandx:
## Description

This changes attempt to skip all the testing/variation selection code if we can detect that DoNotTrack is set in the browser or if the internal `window.builderNoTrack` flag is set. The idea is to provide a consistent user experience when tracking is disallowed/disabled, which in this case only the default variation should always be used.

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -2021,7 +2021,7 @@ export class Builder {
           testVariationName: cookieVariation.name,
         };
       }
-      if (item.variations && size(item.variations)) {
+      if (this.canTrack && item.variations && size(item.variations)) {
         let n = 0;
         const random = Math.random();
         for (const id in item.variations) {
--- a/packages/react/src/components/variants-provider.component.tsx
+++ b/packages/react/src/components/variants-provider.component.tsx
@@ -20,7 +20,7 @@ function getData(content: BuilderContentVariation) {
 const variantsScript = (variantsString: string, contentId: string) =>
   `
 (function() {
-  if (window.builderNoTrack) {
+  if (window.builderNoTrack || window.navigator.doNotTrack === '1') {
     return;
   }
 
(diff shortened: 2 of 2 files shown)


packages/core/src/builder.class.ts
```diff
@@ -535,7 +523,6 @@ export class Builder {
   static isBrowser = isBrowser;
   static isReactNative = isReactNative;
   static isServer = !isBrowser && !isReactNative;
-  static previewingModel = Builder.isBrowser && getQueryParam(location.href, 'builder.preview');
```
teleaziz, line 538: why Is this change needed?

packages/core/src/builder.class.ts
```diff
@@ -1513,10 +1500,8 @@ export class Builder {
     return this;
   }
 
-  get previewingModel() {
-    const search = this.getLocation().search;
-    const params = QueryString.parse((search || '').substr(1));
-    return params['builder.preview'];
+  get previewingModel(): string | undefined {
+    return Builder.isBrowser ? QueryString.parse((this.getLocation().search || '').substr(1))['builder.preview'] : undefined;
```
steve8708, line 1504: is this change necessary for this PR? I'm concerned with some of these changes being breaking and seemingly out of scope for this PR so could have slipped in silently and problematically

this change needs a revert, getLocation() supports server side _and_ client side, and those using server side rendering (e.g. next.js SSR, angular, etc) need the current behavior

steve8708: aka they expect a server side preview when applicable, or else they worry "why is the content flashing when I preview pages, I think I shouldn't publish this as there seems to be an issue"

steve8708:
@mandx can we address the feedback and get this in

packages/react/src/components/variants-provider.component.tsx
```diff
@@ -104,6 +104,8 @@ export const VariantsProvider: React.SFC<VariantsProviderProps> = ({
   initialContent,
   children,
 }) => {
+  if (!builder.canTrack) return children([initialContent]);
```
steve8708, line 107: I don't know this area of the code so will need @teleaziz to chime in on

steve8708 (approved):

teleaziz (approved):

PR #369, opened 2021-02-09
Try extracting the current Shopify Theme ID out of `state.location.query`

mandx:
... And use it to remotely render Liquid blocks. This allows for working
with these Liquid blocks in non-live themes without having to add them
to the main/live theme.

### DIFF
--- a/packages/shopify/react/components/LiquidBlock.tsx
+++ b/packages/shopify/react/components/LiquidBlock.tsx
@@ -49,7 +49,7 @@ export const LiquidBlock = ({
         return;
       }
 
-      const previewThemeID = builderState?.state?.location?.query?.preview_theme_id;
+      const previewThemeID = Builder.isBrowser && window.Shopify?.theme?.id || builderState?.state?.location?.query?.preview_theme_id;
       fetch(
         `${builder.host}/api/v1/shopify/data/render-liquid-snippet?snippet=${blockName}&apiKey=${
           builderState?.context.apiKey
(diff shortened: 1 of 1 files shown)


packages/shopify/react/components/LiquidBlock.tsx
```diff
@@ -37,19 +38,24 @@ export const LiquidBlock = ({
...
       if (!blockName) {
         return;
       }
+
       if (cache[cacheKey]) {
         setHtml(cache[cacheKey]);
         return;
       }
 
+      const previewThemeID = builderState?.state?.location?.query?.preview_theme_id;
```
steve8708, line 52: did you test this and confirm this is working?

iirc shopify removes the preview_theme_id param automatically (301) and so for other stuff I had to pass a new theme_id param in the editor and use that instead

we may want to also check the current URL theme ID since this stuff really only applies client side anyway and iirc state.location.query can be spotty in some contexts

```suggestion
      const previewThemeID = builderState?.state?.location?.query?.theme_id || (Builder.isBrowser && getQueryParam('theme_id');
```

steve8708: actually I think there is something better - on any shoipfy store I'm seeing a window variable for this

`Shopify.theme.id` - maybe we use that

```suggestion
      const previewThemeID = builderState?.state?.location?.query?.preview_theme_id || (Builder.isBrowser && window.Shopify?.theme?.id;
```

steve8708: let's prioritize getting this in and then we can do a webcomponents dev release

mandx: Yeah, I got the `builderState?.state?.location?.query?.preview_theme_id` from the debugger, and so far is always there.

And you are right, `window.Shopify?.theme?.id` is also present so far, and its value also match our `query.preview_theme_id` so far.

Actually I kind of prefer going with your suggestion, but swapping preferences, like:
```suggestion
      const previewThemeID = Builder.isBrowser && window.Shopify?.theme?.id || builderState?.state?.location?.query?.preview_theme_id;
```
Thoughts?

steve8708: 👍

steve8708 (approved):

PR #374, opened 2021-02-10
allow <BuilderConent name="..." /> to be consistent with BuilderComponent

steve8708:
instead of modelName="..."

### DIFF
--- a/packages/react/src/components/builder-content.component.tsx
+++ b/packages/react/src/components/builder-content.component.tsx
@@ -4,12 +4,9 @@ import { NoWrap } from './no-wrap';
 import { applyPatchWithMinimalMutationChain } from '../functions/apply-patch-with-mutation';
 import { VariantsProvider } from './variants-provider.component';
 
-export interface BuilderContentProps<ContentType> {
+export type BuilderContentProps<ContentType> = {
   contentLoaded?: (data: any, content: any) => void;
   contentError?: (error: any) => void;
-  model: string;
-  /** @alias for model */
-  modelName: string;
   options?: GetContentOptions;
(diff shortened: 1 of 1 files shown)


packages/react/src/components/builder-content.component.tsx
```diff
@@ -7,6 +7,8 @@ import { VariantsProvider } from './variants-provider.component';
 export interface BuilderContentProps<ContentType> {
   contentLoaded?: (data: any, content: any) => void;
   contentError?: (error: any) => void;
+  model: string;
+  /** @alias for model */
```
mrkoreye, line 11: Does this comment make it so typescript knows that `modelName` is not required if you pass `model`, and vice versa? If not, then won't this change make it so bot `model` and `modelName` are required?

mandx: Yeah, this looks like will make both props required.

I believe the way to accomplish that is with something like this:
```typescript
export type BuilderContentProps<ContentType> = {
  // ... the rest of the props
} & ({ model: string } | { modelName: string });
```

steve8708: doh! great catch, thanks!

packages/react/src/components/builder-content.component.tsx
```diff
@@ -25,6 +24,9 @@ export class BuilderContent<ContentType extends object = any> extends React.Comp
   get builder() {
     return this.props.builder || builder;
   }
+  get name() {
```
mandx, line 27:
```suggestion
  get modelName(): string {
```
because explicitness?

mandx (approved):

PR #383, opened 2021-02-21
added headless homepage example

russellhtan:
## Description

Added the pure headless homepage example that uses data models only. It renders on its own new path _/homepage-full-headless_.

Because I designed the homepage using Builder, then exported with Builder's "export code" option to styled-components, I had to add the styled-components library to the example.

Also, due to same design decisions above, there are 16 auto-generated divs in _src/pages/HomepageFullHeadless.jsx_ that should probably be renamed. I left them as-is for now, in case someone on the team has suggestion for how to efficiently adjust them.

@teleaziz can you do the code review?
@steve8708 FYI

_Screenshot_

image


steve8708 (approved):

steve8708:
works for me 👍

PR #386, opened 2021-02-25
Update .env.production

gofurtheronline:
## Description

_Screenshot_


steve8708:
hey @gofurtheronline - it seems you have put a private key in here, I don't think this is what you are wanting to do

closing this out and I would strongly suggest to revoke that private key and generate a new one

PR #388, opened 2021-02-25
use more specific class name for dialog main container

donnieblando:
## Description

The plugin for cloudinary images is using a classname to identify the html containing which is too generic (.container).
This can easily lead to errors, in case any other html component on the page uses the same className.
Indeed, in our case, it is clashing with another plugin we are using.

With this MR we are updating the className to be more specific


steve8708:
thanks @donnieblando - just deployed the update!

PR #390, opened 2021-03-01
 lazy load images with IntersectionObserver for better performance

steve8708:


mandx (approved):

mrkoreye:
Bye bye React native package?

steve8708:
yeah, it was legacy code needing removal, and causing confusion as the files were named the same as the react package so if you open up any component, e.g. `Image.tsx` there is a 50/50 chance you are editing the wrong package w/o noticing

will be replaced with a fresh JSX Lite implementation when needed

PR #391, opened 2021-03-01
wrap web-components async in an IIFE

teleaziz:


packages/webcomponents/tools/make-unpkg-dist.ts
```diff
@@ -9,7 +9,7 @@ async function main() {
   // TODO: bootstrap script goes here... hmmm...
   // TODO: export basic builder stuff from here so people can use js, or System.import name it hmm
   const newFileStr = (useAngular = false) =>
-    `if (typeof window !== 'undefined' && !window.builderWebcomponentsLoaded) {\n` +
+    `(() => { if (typeof window !== 'undefined' && !window.builderWebcomponentsLoaded) {\n` +
```
mandx, line 12:
```suggestion
    `(function () { if (typeof window !== 'undefined' && !window.builderWebcomponentsLoaded) {\n` +
```
Maybe go for safer syntax for this bootstrapping code?

mandx (commented):
I though that Rollup can do the IIFE wrapping for us?

steve8708 (approved):
great, also agree w/ armando might as well support es5 since it's easy

mandx (approved):

PR #395, opened 2021-03-02
support children in symbols

steve8708:
most of the work is in the first commit, second commit is prettier


packages/react/src/blocks/Symbol.tsx
```diff
@@ -102,6 +102,8 @@ class SymbolComponent extends React.Component<SymbolProps> {
       key += ':' + dataString;
     }
 
+    console.log('block 2', this.props.builderBlock);
+
```
mandx: Remove?

packages/react/src/blocks/SymbolDropzone.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+type DropzoneProps = {
+  name: string;
+};
+
+export function Dropzone(props: DropzoneProps) {
+  const { name } = props;
+  const context = useContext(BuilderStoreContext);
+
+  console.log('HELLO', context.context.symbolId, context, name);
```
mandx: Hello?

mandx (approved):

packages/react/src/blocks/Symbol.tsx
steve8708: these logs were removed in ad9b6c7 👍

PR #398, opened 2021-03-04
reduce amount of tags used for Builder content

steve8708:
currently every piece of Builder content adds 4 wrapping tags. e.g.a <BuilderComponent>, Symbol, etc

this reduces that down to 2, and we can ultimately get it down to 1 with additional work, but these were the obvious quick wins to remove excess


mrkoreye (approved):
I did a quick search in our codebase for `builder-content-id` since I know that we use that attribute to do things like add overlays in the editor. From a quick look it appears that these changes should be fine, but we might get some weirdness if any of the css assumed that those attributes were on different elements

steve8708:
in messing around deeper, think I'm going to close this in favor of our v2 jsx-lite powered codegen when live. will be more small, minimal, clean, etc

PR #400, opened 2021-03-04
hide intersection observer usage behind a flag

teleaziz:
~Intersection observer was causing issues in carousels, can be fixed with setting a rootMargin but not when the users has entered negartive margins on the slide, even then it's for some reason not working when iframed ( only when used in carousel)~

The issue only occur when users have negative margins on a slide in a carousel

Related to:
https://github.com/w3c/IntersectionObserver/issues/301
https://github.com/w3c/IntersectionObserver/issues/266

### DIFF
--- a/packages/react/src/blocks/Image.tsx
+++ b/packages/react/src/blocks/Image.tsx
@@ -193,7 +193,7 @@ class ImageComponent extends React.Component<any, { imageLoaded: boolean; load: 
         this.setState({
           load: true,
         });
-      } else if (this.props.useIO && typeof IntersectionObserver === 'function' && this.pictureRef) {
+      } else if (this.props.useIntersectionObserver && typeof IntersectionObserver === 'function' && this.pictureRef) {
         const observer = (this.intersectionObserver = new IntersectionObserver(
           (entries, observer) => {
             entries.forEach(entry => {
--- a/packages/react/src/blocks/Image.tsx
+++ b/packages/react/src/blocks/Image.tsx
@@ -193,7 +193,7 @@ class ImageComponent extends React.Component<any, { imageLoaded: boolean; load: 
         this.setState({
           load: true,
         });
-      } else if (typeof IntersectionObserver === 'function' && this.pictureRef) {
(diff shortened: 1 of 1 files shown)


packages/react/src/blocks/Image.tsx
```diff
@@ -193,7 +193,7 @@ class ImageComponent extends React.Component<any, { imageLoaded: boolean; load:
         this.setState({
           load: true,
         });
-      } else if (typeof IntersectionObserver === 'function' && this.pictureRef) {
+      } else if (this.props.useIO && typeof IntersectionObserver === 'function' && this.pictureRef) {
```
mrkoreye, line 196: Can we make the prop name more clear? 
```suggestion
      } else if (this.props.useIntersectionObserver && typeof IntersectionObserver === 'function' && this.pictureRef) {
```

steve8708:
interesting, great find. I wonder if its time to just lean on `loading="lazy". I think we can if google page speed insights properly respects it

aka no more scroll listneners or intersection observers, just always render the `<img`, it already has `loading="lazy"` so if we just do a test and verify that google page speed insights doesn't penalze for this (aka they are using a recent version of chrome) I'm ok with removing all of this stuff and doing that (we can remove a lot of code)

mrkoreye:
My only concern then would be support on mobile, since it looks like iOS safari does not have support yet (it is behind a flag): https://caniuse.com/loading-lazy-attr. So google page speed will not complain and thus will not ding someone's search ranking, but mobile performance could end up being really bad for real users

steve8708:
totally fair point, maybe we leave things as-is and just accept theres a weird negative margin browser bug with IntersectionObserver that we can inform customers how to work around, and then when iOS safari supports loading=lazy we lean fully on that?

mrkoreye:
Sounds good. Maybe we can make this flag true by default for new content, then if we detect a negative margin in the editor we flip it to false?

teleaziz:
closing in favor of user education https://forum.builder.io/t/images-are-not-lazy-loaded-when-used-in-a-box-with-negative-margins-inside-a-scroller/492

PR #402, opened 2021-03-05
update error message and allow overriding of api host

teleaziz:


packages/core/src/builder.class.ts
```diff
@@ -41,6 +41,7 @@ export const validEnvList = [
   'cloud',
   'fast',
   'cdn2',
+  'forceBuilderCDN',
```
mrkoreye, line 44:
```suggestion
  'cdn',
```

packages/core/src/builder.class.ts
```diff
@@ -1785,14 +1787,19 @@ export class Builder {
       case 'development':
       case 'dev':
         return 'http://localhost:5000';
-      default:
+      case 'forceBuilderCDN':
```
mrkoreye, line 1790:
```suggestion
      case 'cdn':
```

mrkoreye (approved):

packages/core/src/builder.class.ts
mandx, line 44:
```suggestion
  'cdn-prod',
```
Reads more on par with the rest...

mandx (approved):

PR #416, opened 2021-03-12
possible fix for video autoplay on iOS Safari

steve8708:


mrkoreye (approved):

ca136 (approved):

steve8708:
so I've been testing this and a bunch of other updates on the iOS simulator and none seem to work 😭 

I did notice.... the our current live code works when the (simulator) is in landscape, but not portrait... I have no clue what's going on here

mrkoreye:
maybe iOS sent an update that breaks it? does an old version of our code work?

steve8708:
damn, just confirmed it actually does work on `1.3.21` which is before the removeAttribute update. I'll keep poking

steve8708:
............................................................................................. it was user error

PR #422, opened 2021-03-23
add roleOption to component schema and restrict custom code access to…

teleaziz:
… developer roles

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -325,6 +325,9 @@ export interface Component {
   defaultChildren?: BuilderElement[];
   defaults?: Partial<BuilderElement>;
   hooks?: { [key: string]: string | Function };
+  /**
+   * Hide your component in editor, useful for gradually deprecating components
+  */
   hideFromInsertMenu?: boolean;
   // For webcomponents
   tag?: string;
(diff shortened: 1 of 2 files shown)


packages/core/src/builder.class.ts
```diff
@@ -375,8 +378,17 @@ export interface Component {
...
+
+  /**
+   * Use to restrict access to your component based on a the current user role options
+   * By default components will show to every user role
+   * for more information on roles in builder check https://www.builder.io/c/docs/guides/roles-and-permissions
+  */
+  roleOption?: RoleOption;
 }
 
+type RoleOption = 'read' | 'publish' | 'editCode' | 'editDesigns' | 'admin' | 'create';
```
steve8708, line 390: similar comment to above, these are `permissions` not `roles`

packages/core/src/builder.class.ts
```diff
@@ -325,6 +325,9 @@ export interface Component {
   defaultChildren?: BuilderElement[];
   defaults?: Partial<BuilderElement>;
   hooks?: { [key: string]: string | Function };
+  /**
+   * Hide your component in editor, useful for gradually deprecating components
+  */
```
steve8708, line 330: jsdoc 👍 👍 👍

steve8708:
good idea 👍

steve8708 (approved):

PR #435, opened 2021-04-09
Typedocs for BuilderContentProps

Dkendal:
Typedoc changes.

### DIFF
--- a/packages/react/src/components/builder-content.component.tsx
+++ b/packages/react/src/components/builder-content.component.tsx
@@ -5,14 +5,46 @@ import { applyPatchWithMinimalMutationChain } from '../functions/apply-patch-wit
 import { VariantsProvider } from './variants-provider.component';
 
 export type BuilderContentProps<ContentType> = {
+  /**
+   * Callback to run when content is fetched and loaded.
+   */
   contentLoaded?: (data: any, content: any) => void;
+  /**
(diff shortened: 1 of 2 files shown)


packages/react/src/components/builder-content.component.tsx
```diff
@@ -5,14 +5,46 @@ import { applyPatchWithMinimalMutationChain } from '../functions/apply-patch-wit
...
+   */
   contentLoaded?: (data: any, content: any) => void;
+  /**
+   * Callback to run if an error occurred while fetching content.
+   */
   contentError?: (error: any) => void;
   options?: GetContentOptions;
+  /**
+   * Placeholder content displayed while real content is fetched.
+   */
```
mandx, line 19: I kind of want to expand more on what happens with this content when the real content arrives (I'm assuming it gets replaced by the real content) _or_ when fetching fails and we get an error (I think nothing happens, this contents stays and it's our responsibility to show errors ourselves if that's what we want?).

mandx (approved):

packages/react/src/components/builder-content.component.tsx
Dkendal: I think the initial description is wrong. I'm going to remove it until I have a better idea of what it does.

PR #437, opened 2021-04-09
Add jsdoc for BuilderContent and BuilderComponent.

Dkendal:
Adds jsdoc for these two classes.

### DIFF
--- a/packages/react/src/components/builder-component.component.tsx
+++ b/packages/react/src/components/builder-component.component.tsx
@@ -390,6 +390,18 @@ function searchToObject(location: Location | Url) {
   return obj;
 }
 
+/**
+ * Responsible for rendering Builder content of type: 'page' or 'section' to
+ * react components. It will attempt to fetch content from the API based on
+ * default used attributes (URL path, and device type) unless a `BuilderContent`
+ * object is provided to `props.content`
+ *
+ * Use it to mount content in desired location, enable editing in place when
+ * previewed in the editor.
+ *
+ * Supports server-side-rendering when passed the content json as
+ * `props.content`.
+ */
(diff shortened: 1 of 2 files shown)


packages/react/src/components/builder-content.component.tsx
```diff
@@ -45,6 +45,12 @@ export type BuilderContentProps<ContentType> = {
   content?: BuilderContent;
 } & ({ model: string } | { modelName: string }); // model and modelName are aliases of the same thing¸
 
+/**
+ * When passed content json explicitly it'll calculate a/b tests on the content
+ * and pass the winning variation down to the children function. If then content
+ * prop was omitted it'll try to fetch matching content from your Builder
+ * account based on the default user attributes and model.
+ */
```
mandx, line 53: Maybe worth the disclaimer that test selection, etc happens only if DoNotTrack is false (which in that case, only the default variation is used)

mandx (approved):

packages/react/src/components/builder-component.component.tsx
```diff
@@ -390,6 +390,18 @@ function searchToObject(location: Location | Url) {
   return obj;
 }
 
+/**
+ * Responsible for rendering Builder content of type: 'page' or 'section' to
+ * react components. It will attempt to fetch content from the API based on
+ * default used attributes (URL path, and device type) unless a `BuilderContent`
```
teleaziz, line 396:
```suggestion
 * defined user attributes (URL path, device type, or any custom targeting) unless a `BuilderContent`
```

packages/react/src/components/builder-content.component.tsx
Dkendal: Done

PR #449, opened 2021-04-26
when preview field is not defined prompt user to pick one

teleaziz:
This is related to the issue here, https://github.com/BuilderIO/nextjs-shopify/issues/14, it'll prompt the user to pick a preview item when it's not defined.

### DIFF
--- a/packages/commerce-plugin-tools/src/editors/ResourcesPicker.tsx
+++ b/packages/commerce-plugin-tools/src/editors/ResourcesPicker.tsx
@@ -71,6 +71,7 @@ export const ResourcePicker: React.FC<
     api: CommerceAPIOperations;
     omitIds?: string[];
     resourceName: string;
+    title?: string;
   }
 > = props => {
   const store = useLocalStore(() => ({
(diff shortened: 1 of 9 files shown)


packages/commerce-plugin-tools/src/editors/ResourcesPicker.tsx
```diff
@@ -229,6 +234,19 @@ export const ResourcesPickerButton: React.FC<ResourcesPickerButtonProps> = props
...
+  useEffect(() => {
+    if (
+      props.context.globalState.globalDialogs.length === 0 &&
+      props.context.designerState?.editingContentModel &&
+      props.isPreview &&
+      (!props.value || store.error)
+    ) {
+      requestAnimationFrame(() =>
+        store.showPickResouceModal(`Pick a ${props.resourceName} to preview`)
+      );
```
mandx, line 246: Why not `setTimeout` instead?
```suggestion
      setTimeout(() =>
        store.showPickResouceModal(`Pick a ${props.resourceName} to preview`)
      );
```

mandx (approved):

packages/commerce-plugin-tools/src/editors/ResourcesPicker.tsx
teleaziz: both seem to work for me here, is there anything specific against using requestAnimationFrame?

mandx: `requestAnimationFrame` is rendering focused so the callback fires in the middle of the paint cycle, while `setTimeout` is more for more general usage. My point is simply that browsers can optimize/throttle things differently, and `requestAnimationFrame` could maybe change its behavior in the future, as browsers implement different rendering performance techniques.

mandx (approved):

PR #457, opened 2021-05-04
option to force use of require to handle missing vm2 in nextjs serverless environment

teleaziz:
This fixes an issue where binding was not running correctly in vercel/netlify serverless environment, adds an option to force the use of `require`

### DIFF
--- a/packages/react/src/functions/string-to-function.ts
+++ b/packages/react/src/functions/string-to-function.ts
@@ -20,7 +20,8 @@ export function stringToFunction(
   str: string,
   expression = true,
   errors?: Error[],
-  logs?: string[]
+  logs?: string[],
+  forceRequire?: boolean,
 ): BuilderEvanFunction {
   /* TODO: objedct */
   if (!str || !str.trim()) {
@@ -118,7 +119,7 @@ export function stringToFunction(
         // for the server build
         // TODO: cache these for better performancs with new VmScript
         // tslint:disable:comment-format
-        const { VM } = safeDynamicRequire('vm2');
+        const { VM } = forceRequire ? require('vm2') : safeDynamicRequire('vm2');
         const [state, event] = args;
         return new VM({
           timeout: 100,
--- a/packages/react/package-lock.json
+++ b/packages/react/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@builder.io/react",
-  "version": "1.1.44",
+  "version": "1.1.45-0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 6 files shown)


mandx (approved):

packages/react/src/functions/string-to-function.ts
```diff
@@ -118,7 +119,7 @@ export function stringToFunction(
         // for the server build
         // TODO: cache these for better performancs with new VmScript
         // tslint:disable:comment-format
-        const { VM } = safeDynamicRequire('vm2');
+        const { VM } = forceRequire ? require('vm2') : safeDynamicRequire('vm2');
```
steve8708, line 122: so this will reintroduce the problem that safeDynamicRequire aims to solve, which is that client side bundlers like webpack will see the `require('vm2')` and try to bundle which we don't want

I think we need to find another approach here, also ideally one without all the context/props passing and can just be right here. I would have to think on to offer alternative suggestions though, but jsut want to point out what the prior code was doing and why first

steve8708: one thing that would help me - do we know why `eval('require')` doesn't work on vercel/netlify? should we do `new Function('return require')` for instance? or try `const _require = require` (iirc that second option didn't solve the bundling problem)

alternatively we could use build macros for separate browser and server builds, but that also had issues in the past too

open to other ideas but just wanted to add additional context

teleaziz: The issue comes from when this line runs in serverless netlify/vercel 
```
eval('require')('vm2') 
```
is throwing`vm2 not found` error, even when adding vm2 as a dependency, what do you think if we let the user pass their vm2 instance here?

steve8708:
After thinking on this - the way to go here might be an async import. at least this way if vm2 is built for the browser, it is in a separate module/bundle that is never used

the tricky part tho is that this is in a synchronous function, so we may need to do something like this at the root of the file

```
let VM: null | typeof import('vm2').VM

if (Builder.isServer) {
  import('vm2').then(mod => VM = mod.VM)
}

// then later when needed
if (!VM) {
  throw new Error('vm not loaded')
}
```
in theory, the way modules are loaded in node.js, this should rarely if ever be an issue

one note though - this would require preserving this dynamic import in the compiled output. not sure if our rollup build does this by default, but as a side note if it doesn't, we may want to explore removing rollup for this package all together as I realized it's only consumed by other projects that use their own bundler, so bundling with rollup ahead of time is not optimal anyway and should go some time or another (so if supporting my suggestion with the current build is too cumbersome, we may want to consider just moving to plain `tsc` to build this package)

steve8708:
I know a lot going on but want to make sure we don't forget about this

teleaziz:
closing this, this approach turned out to not be useful and defeats the purpose of `safeDynamicRequire`, will continue digging and reopen

PR #463, opened 2021-05-06
update DesignerState interface

teleaziz:


mrkoreye (approved):
Can you add a comment above each one about what they are for/do?

mandx (approved):

PR #465, opened 2021-05-07
chore: add basic bazel support

mhevery:
## Description

Adds basic Bazel support to get the discussion going.


WORKSPACE
```diff
@@ -0,0 +1,58 @@
...
+# Check the rules_nodejs version and download npm dependencies
+# Note: bazel (version 2 and after) will check the .bazelversion file so we don't need to
+# assert on that.
+load("@build_bazel_rules_nodejs//:index.bzl", "check_rules_nodejs_version", "node_repositories", "npm_install")
+
+check_rules_nodejs_version(minimum_version_string = "2.2.0")
+
+# Setup the Node.js toolchain
+node_repositories(
+    node_version = "15.0.0",
```
Dkendal, line 26: We're currently on node v12 / v10 in production, and the majority of packages use the < v14 / npm6 lockfile.

WORKSPACE
```diff
@@ -0,0 +1,58 @@
...
+
+# Bellow is an example of how we could run more than one npm install in a project.
+#
+# For now we have decided to have a single node_modules at the root because it centralizes
+# our dependencies and is more inline with mono-repo.
+#
+# TODO(npm): make sure that the package.json in individual packages/*/package.json have
+#       consistent versions from the global one. This will most likely have to be don
+#       by having those files generated from the main package.json.
+#
```
Dkendal, line 45: This would be nice one day :smile:

package.json
```diff
@@ -1,15 +1,38 @@
 {
   "private": true,
   "devDependencies": {
+    "@angular/common": "^11.2.13",
```
Dkendal, line 4: Are the angular dependencies supposed to be pulled up onto the root package.json? Or are these hoisted from packages/angular?

Dkendal:
Awesome! Can't wait to have a more well managed repo :)

WORKSPACE
steve8708: latest LTE is 14 so maybe we go w/ that

steve8708 (approved):

.bazelignore
```diff
@@ -0,0 +1,24 @@
...
+packages/native/node_modules
+packages/plugin-loader/node_modules
+packages/react/node_modules
+packages/shopify/node_modules
+packages/storybook/node_modules
+packages/utils/node_modules
+packages/vue/node_modules
+packages/webcomponents/node_modules
+packages/widgets/node_modules
+packages/wordpress/node_modules
```
steve8708, line 24: dang, I take it bazelignore doesn't take wildcards or any other type of pattern to block out all node_modules like we can in gitignore/npmignore?

Dkendal: Doesn't look like it https://github.com/bazelbuild/bazel/issues/7093

mhevery: unfortunately does not. :-(

package.json
mhevery: Right, so this is one of the discussions we should have.

There are two schools of thought on this one, and both have pros/cons:
- In mono-repo you should have a one-version policy and there fore you should have one place to manage all of your versions etc... The downside is that you will need to have a way to make package.json for individual sub-packages. This is what Angular uses.
- The other option is to have a mano-repo, but really have a different version for each sub-package. Which I think is what you are suggesting.


A generously don't know what the right answer is here. I have discussed this for a bit with @steve8708 and we decided to go the first route, but I think it is a bit riskier. But would love to know what other opinions are.

WORKSPACE
mhevery: Right, so the decision to make is:
1) do we want to centrally manage the versions?
2) or do we want to have sub-packages in each folder? 

(see my other comment)

WORKSPACE
mhevery: 👍

package.json
mhevery: Mono-repo does not really mean anything. It is whatever you define it to be.
Yes I went down path 1 where the versions are shared between packages but ran into an issue, so I am abandoning it and going down path 2 which is what you are advocating here.

packages/angular/ng-package.json
```diff
@@ -1,5 +1,6 @@
 {
   "$schema": "./node_modules/ng-packagr/ng-package.schema.json",
+  "dest": "../../bazel-out/darwin-fastbuild/bin/packages/angular/ng-pkg",
```
mhevery, line 3: This is a hack and needs to be properly fixed. Still discussing with @alexeagle on how best to do this.

PR #477, opened 2021-05-25
handle JSON parse errors in node.js

### LINKED ISSUE #476: Hard crash when `JSON.parse` receives non-JSON content
I was testing query parameters with the Builder SDK when apparently I tested too many iterations and Google blocked me from further queries...

I get this error in my terminal
```
undefined:1
<html><head><meta http-equiv="content-type" content="text/html; charset=utf-8"/><title>Sorry...</title><style> body { font-family: verdana, arial, sans-serif; background-color: #fff; color: #000; }</style></head><body><div><table><tr><td><b><font face=sans-serif size=10><font color=#4285f4>G</font><font color=#ea4335>o</font><font color=#fbbc05>o</font><font color=#4285f4>g</font><font color=#34a853>l</font><font color=#ea4335>e</font></font></b></td><td style="text-align: left; vertical-align: bottom; padding-bottom: 15px; width: 50%"><div style="border-bottom: 1px solid #dfdfdf;">Sorry...</div></td></tr></table></div><div style="margin-left: 4em;"><h1>We're sorry...</h1><p>... but your computer or network may be sending automated queries. To protect our users, we can't process your request right now.</p></div><div style="margin-left: 4em;">See <a href="https://support.google.com/websearch/answer/86640">Google Help</a> for more information.<br/><br/></div><div style="text-align: center; border-top: 1px solid #dfdfdf;"><a href="https://www.google.com">Google Home</a></div></body></html>
^

SyntaxError: Unexpected token < in JSON at position 0
    at JSON.parse (<anonymous>)
    at IncomingMessage.<anonymous> (/Users/tony/code/project/node_modules/@builder.io/sdk/dist/index.cjs.js:2060:34)
    at IncomingMessage.emit (events.js:327:22)
    at IncomingMessage.EventEmitter.emit (domain.js:486:12)
    at endReadableNT (_stream_readable.js:1327:12)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
error Command failed with exit code 1.
```

Which renders into this 
image

While this is a bit of a pain, I can understand this. They think I'm a bot and they're protecting your infrastructure. I suspect I'll be allowed to query again soon.

The issue I have is that this error causes our app to hard crash and even though the `builder.get` is wrapped in a try/catch the line `@builder.io/sdk/dist/index.cjs.js:2060:34` still manages to hard crash the process. Our error handling doesn't get a chance to catch the error.

I believe this might be related the an outage we experienced a few days ago too where non-JSON content from an `Internal Error` was being thrown.

steve8708:
fixes #476

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -1939,7 +1939,11 @@ export class Builder {
 
           // The whole response has been received. Print out the result.
           resp.on('end', () => {
-            resolve(JSON.parse(data));
+            try {
+              resolve(JSON.parse(data));
+            } catch (err) {
+              reject(err)
+            }
           });
         })
         .on('error', (error: any) => {
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -1942,7 +1942,7 @@ export class Builder {
             try {
               resolve(JSON.parse(data));
             } catch (err) {
-              reject(err)
+              reject(err);
             }
           });
         })


mandx (approved):

mhevery (approved):

mrkoreye (approved):

steve8708:
published to NPM in `@builder.io/sdk@1.1.23-11` and `@builder.io/react@1.1.45-23`

packages/core/src/builder.class.ts
```diff
@@ -1939,7 +1939,11 @@ export class Builder {
 
           // The whole response has been received. Print out the result.
           resp.on('end', () => {
-            resolve(JSON.parse(data));
+            try {
+              resolve(JSON.parse(data));
+            } catch (err) {
+              reject(err);
```
ap-shahar, line 1945: @steve8708 in the past based on the bug @tonytamps reported #476  we used to get data as an HTML page if the machine that's running the query flagged as a bot. This PR seems to be fixing the crash, but we would still reject here which is why on our code what happens is:
```
builder
 .get(model, builderOptions)
 .promise()
 .then((res) => {
   // This code is never reached
  })
  .catch((ex) => {
    // This is where we log the unexpected token
   })
```
and the error we get in the rejection fits a JSON.parse exception `Unexpected token < in JSON at position 0`.

this seems to be the root cause of the outage in the last week of October. What can we change on our end to use the SDK in such a way that we won't get flagged as a bot by any of your services?

steve8708: ah yes - in the cases where we failed to fetch data, I believe the above behavior is correct right? `.catch` will run forwarding whatever unexpected error occurred, as opposed to `.then` as it was not a success

re: bot detection, we actually found out what the issue was and fortunately it was not bot detection. Fastly had a POP node with a poisoned cache that was not being purged correctly. the reason that switching off of the SDK and to the REST APIs circumvented this is that the API parameters sent manually are slightly different than what the SDK sends so we didn't have any more usage of the poisoned caches (as the full URL is the cache key)

will have more info for you on this later

PR #480, opened 2021-05-28
Fix broken link to renamed dropdown selector and cleanup example url

michaelfavia:
## Description
My entry for the trivial patch of the month club. Was just walking through the docs and hit a broken link. I know its alot to keep in sync so thought I might lend a quick hand.


steve8708:
Great catch, thanks @michaelfavia !

PR #491, opened 2021-06-11
update nextjs-amp example, add preconfigured clone and make it previe…

teleaziz:
…wable/editable under editor


examples/next-js-amp/pages/preview.js
```diff
@@ -0,0 +1,5 @@
+export const config = { amp: 'hybrid'};
+
+import page from './[[...path]]';
+
+export default page;
```
mandx, line 5: Is this `/preview` route needed for previewing in the Builder editor?

mandx (approved):
Maybe run this through `prettier`?

examples/next-js-amp/pages/preview.js
teleaziz: yeah it simulates amp and it's a lot closer to the end result

PR #492, opened 2021-06-15
utils: add ability to extend inline symbol data

teleaziz:
This makes it easy to SSR inlined symbols on headless sites

### DIFF
--- a/packages/utils/package-lock.json
+++ b/packages/utils/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@builder.io/utils",
-  "version": "1.0.3",
+  "version": "1.0.4",
   "lockfileVersion": 1,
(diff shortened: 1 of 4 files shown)


steve8708:
Don't inline symbols already SSR? If the data is fetched correctly this should already work

teleaziz:
@steve8708 the case this solves is when the inlined symbol relies on services passed through the context prop:

for example a product box that takes the handle as input would fetch the product object using context.productsService:

```
state.productObject = await context.productsService.getProductByHandle( state.productHandle);
```

with this update I can change it to read the value from the immediately available object which will fix ssr :

```
state.productObject = state.initialState?.product || await ...
```

steve8708 (approved):

PR #498, opened 2021-07-06
[commerce-plugin-tools]:  Wait for user initialized before accessing organization settings

teleaziz:


packages/commerce-plugin-tools/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@builder.io/commerce-plugin-tools",
-  "version": "0.2.1-2",
+  "version": "0.2.1-3",
```
mandx, line 3: This is a breaking change of the public API, we could follow Semver and bump the leftmost nonzero digit in the version string.

teleaziz:
Closing this issue, false positive, it was missing because of a user misconfigured externals

PR #501, opened 2021-07-13
allow restricting registered components to a list of models

teleaziz:
This come up a couple of times developers asking for a way to restrict a registered component to a subset of models

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -494,7 +494,7 @@ export interface Component {
   tag?: string;
   static?: boolean;
   /**
-   * Passing a list of models will restrict using the component to only the models defined here
+   * Passing a list of model names will restrict using the component to only the models listed here, otherwise it'll be available for all models
    */
   models?: string[];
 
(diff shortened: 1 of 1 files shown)


packages/core/src/builder.class.ts
```diff
@@ -830,7 +834,11 @@ export class Builder {
       ...options,
     };
     this.addComponent(spec);
-    if (isBrowser) {
+    const editable =
+      options.models && this.singletonInstance.editingModel
+        ? isBrowser && options.models.includes(this.singletonInstance.editingModel)
+        : isBrowser;
+    if (editable) {
```
teleaziz, line 841: this line and the above are the only real changes, the rest is prettier

packages/core/src/builder.class.ts
```diff
@@ -493,6 +493,10 @@ export interface Component {
   // For webcomponents
   tag?: string;
   static?: boolean;
+  /**
+   * Passing a list of models will restrict using the component to only the models defined here
+   */
+  models?: string[];
```
mrkoreye, line 499: Is this model `id` or `name`? Maybe just specify in the comment above

mrkoreye (approved):

Dkendal (approved):

PR #504, opened 2021-07-20
fix(Video): hide child div container if no children

ca136:
This was covering the video controls preventing the user from interacting with the video.


ca136:
It's been a while since I worked with SDKs, so I haven't had a chance to test this yet.

mandx (approved):

steve8708:
@ca136 merge?

PR #539, opened 2021-08-19
add number field max,min,step to Input interface

teleaziz:


mrkoreye (approved):
It would be good to update the doc here to include this: https://www.builder.io/c/docs/custom-react-components#:~:text=the%20schema%20of%20inputs%20is%20as%20follows

PR #563, opened 2021-09-16
React native video component

steve8708:


package.json
```diff
@@ -46,6 +46,7 @@
     "lint:fix": "prettier --write '**/*.{js,jsx,ts,tsx}'"
   },
   "dependencies": {
+    "@types/react-native-video": "^5.0.9",
```
teleaziz, line 49: was this meant to be on the top level package.json

teleaziz (approved):

package.json
steve8708: doh no :D thanks for catching

PR #567, opened 2021-09-17
BigCommerce Integration

shyamseshadri:
## Description

BigCommerce integration. Uses REST APIs for BigCommerce as per https://developer.bigcommerce.com/api-reference/store-management/catalog/. 

Readme yet to be updated.

Not yet ready for merging, creating for initial comments / thoughts.

### DIFF
--- a/plugins/bigcommerce/src/plugin.ts
+++ b/plugins/bigcommerce/src/plugin.ts
@@ -61,19 +61,22 @@ registerCommercePlugin(
           return product;
         },
 
-        /* async findByHandle(handle: string) {
-          const response = await fetch(
-            baseUrl(`api/catalog_system/pub/products/search/${handle}/p`),
+        async findByHandle(handle: string) {
+          const queryParams = '?limit=100&include=primary_image';
+          const response: any = await fetch(
+            baseUrl(`products${queryParams}`),
             {
               headers,
             }
-          ).then(res => res.json());
-          const product = response.map(transformProduct)[0];
-          return product;
-        }, */
+          ).then(res => {
+            return res.json();
+          });
+          const products = response.data.map(transformResource);
+          return products.find((each: any) =>  each.handle === handle);
+        },
         async search(search: string) {
-          const queryParams = '?include=primary_image' +
-             (search ? `&name=${search}` : '');
+          let queryParams = '?limit=100&include=primary_image';
+          queryParams += (search ? `&keyword=${search}` : '');
           const response: any = await fetch(
             baseUrl(`products${queryParams}`),
             {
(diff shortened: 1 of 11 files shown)


plugins/bigcommerce/src/plugin.ts
```diff
@@ -0,0 +1,153 @@
...
+          const product =
+            basicCache.get(key) ||
+            (await fetch(baseUrl(`products/${id}?include=primary_image`), { headers })
+              .then(res => res.json())
+              .then(transformResource));
+          basicCache.set(key, product);
+          return product;
+        },
+
+        /* async findByHandle(handle: string) {
```
teleaziz, line 64: although this is optional in the interface it's still needed for all product/collection resources

plugins/bigcommerce/src/plugin.ts
```diff
@@ -0,0 +1,153 @@
...
+            options: {
+              product: id,
+            },
+          };
+        },
+      },
+      category: {
+        async findById(id: string) {
+          const key = `${id}categoryById`;
+          // https://{accountName}.{environment}.com.br/api/catalog/pvt/category/categoryId
```
teleaziz, line 104: old comment?

plugins/bigcommerce/src/plugin.ts
```diff
@@ -0,0 +1,153 @@
...
+          const category =
+            basicCache.get(key) ||
+            (await fetch(baseUrl(`categories/${id}`), { headers })
+              .then(res => res.json())
+              .then(transformResource));
+          basicCache.set(key, category);
+          return category;
+        },
+
+        /* async findByHandle(handle: string) {
```
teleaziz, line 114: same as above, findByHandle would be useful for categories as well

plugins/bigcommerce/src/plugin.ts
teleaziz: expanding on the need for handles (slugs / url_keys), it allows the developer to read the handle from url params and send it as custom targeting attributes in one api call , as opposed to IDs, where they'll need to fetch the product object from the `handle` in url and they send the ID to builder for targeting

shyamseshadri: So Bigcommerce doesn't have an API to fetch by handle. Instead, I've just done a get all products, and searched for a matching product with the handle. Not sure if I should deal with pagination use-cases, where it might not  be in the first 100 results. Whats the recommendation for those usecases?

shyamseshadri:
@teleaziz this is ready for review and merge I believe.

teleaziz (approved):

PR #570, opened 2021-09-17
[data-plugin-tools] add library with helpers to save settings and check if keys are still needed

teleaziz:
This is is still alpha, would need an editor release


samijaber (approved):
🏆

PR #571, opened 2021-09-17
Add Contentful data plugin

teleaziz:
~still a draft just wanted to showcase the usage of data plugin tools~ , ~missing a doc~


plugins/contentful/src/plugin.ts
```diff
@@ -0,0 +1,119 @@
...
+                type: 'object',
+                friendlyName: `${type.name} fields`,
+                subFields: acceptableFields.map(field => ({
+                  ...field,
+                  type: field.type === 'Symbol' ? 'Text' : field.type.toLowerCase(),
+                  name: field.id,
+                  friendlyName: field.name,
+                })),
+              } as any);
+            }
```
steve8708, line 72: ooo what does this `fields` stuff do?

plugins/contentful/src/plugin.ts
```diff
@@ -0,0 +1,119 @@
...
+    const accessToken = settings.get('accessToken')?.trim();
+    const client = await contentful.createClient({
+      space: spaceId,
+      accessToken,
+    });
+    console.log(' here contentful ', client);
+    return {
+      async getResourceTypes() {
+        const contentTypes = await client.getContentTypes();
+        console.log(' here content types are ', contentTypes);
```
steve8708, line 43: remove?

plugins/contentful/src/plugin.ts
teleaziz: it allows querying model by fields, to form a query at the end `&fields.firstName=Steve`, recording a demo to showcase this

steve8708: nice!!!!!! I thought that would end up being really hard to support, but this has huge value

how hard would it be to add this to contentstack too @samijaber ?

and is there an easy way to pick a field to sort on? if we can have basic querying by fields and basic sorting by fields that covers a ton of use cases and where I was hoping we'd get to (but didn't think we'd be able to so soon!!)

plugins/contentful/src/plugin.ts
```diff
@@ -0,0 +1,119 @@
...
+  // Observable map of the settings configured above
+  async settings => {
+    // Get public key input from user
+    const spaceId = settings.get('spaceId')?.trim();
+    const accessToken = settings.get('accessToken')?.trim();
+    const client = await contentful.createClient({
+      space: spaceId,
+      accessToken,
+    });
+    console.log(' here contentful ', client);
```
mrkoreye, line 39:
```suggestion
```

plugins/contentful/src/plugin.ts
samijaber: Very cool! Will connect w/ Aziz about this (please share your demo whenever), as I avoided digging into how these `inputs` and decided to leave it for later.

Will merge my v1 of the plugin as-is, and look into improving querying/filtering/searching in a 2nd pass.

teleaziz: shared demo in slack

teleaziz: Screen Shot 2021-09-20 at 2 31 34 PM
a
Screen Shot 2021-09-20 at 2 31 40 PM
added a field to order results, and hid the fields query as advanced with some helper texts

steve8708:
can we / should we get this in?

plugins/contentful/src/plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+                const omitValue = fields[key] === '';
+                return {
+                  ...acc,
+                  ...(omitValue ? {} : { [key]: fields[key] }),
+                };
+              }, {});
+            }
+
+            const params = qs.stringify(
+              { ...options, fields, access_token: accessToken, content_type: type.sys.id },
```
samijaber, line 119: it's a bit surprising to me that we're spreading all of `options`. Can we more explicitly only provide the subset we want? I know that `options.fields` and `options.entry` are not meant to be passed. I guess it's `options.include` and `options.order` that are passed along to the Contentful API as-is?

plugins/contentful/src/plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+              { allowDots: true, skipNulls: true }
+            );
+            // by query
+            return `https://cdn.contentful.com/spaces/${spaceId}/environments/master/entries?${params}`;
```
samijaber, line 123: small detail: I'd extract the common `https://cdn.contentful.com/spaces/${spaceId}/environments/master/entries/` between this and the `optoins.entry` use-case into a variable

mrkoreye (approved):

plugins/contentful/src/plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+        const contentTypes = await client.getContentTypes();
+        return contentTypes.items.map(type => ({
+          name: type.name,
+          id: type.sys.id,
+          canPickEntries: true,
+          inputs: () => {
+            const acceptableFields = type.fields.filter(field =>
+              ['Text', 'Boolean', 'Number', 'Symbol'].includes(field.type)
+            );
+            const fields: any = [
```
samijaber, line 52: I think you can use the `Input` type from our SDK no?

```suggestion
            const fields: Input[] = [
```

samijaber (approved):
🆗

plugins/contentful/src/plugin.ts
teleaziz: optins.entry is handles separately https://github.com/BuilderIO/builder/pull/571/files/9d8825f8501303464a1fe96e36aa7ddffb6e76e8#diff-451568f4e0f9d1e066b70752497ac776e51de99ca17317c5be4ea2ffab82eb90R102

PR #572, opened 2021-09-20
Contentstack data plugin

samijaber:
## Description

Implements contenstack data provider plugin


packages/commerce-plugin-tools/src/interfaces/resource.ts
```diff
@@ -5,8 +5,8 @@ export interface Image {
 }
 
 export interface Resource {
-  id: number;
+  id: string | number;
   title: string;
-  handle: string;
+  handle?: string;
```
samijaber, line 10: no rush on deploying/publishing these changes since I'm no longer using this lib, but figured I'd keep them in the PR since they improve the types

teleaziz (approved):

PR #575, opened 2021-09-21
Contentstack sorting/searching

samijaber:
## Description

this PR adds the ability to order and filter Contentstack data

_Screenshot_
https://share.getcloudapp.com/8Lu5EnwW

### DIFF
--- a/plugins/contentstack/src/plugin.ts
+++ b/plugins/contentstack/src/plugin.ts
@@ -8,8 +8,10 @@ import {
 } from '@builder.io/data-plugin-tools';
 import kebabCase from 'lodash/capitalize';
 import capitalize from 'lodash/kebabCase';
+import mapValues from 'lodash/mapValues';
 import appState from '@builder.io/app-context';
 import qs from 'qs';
+import { Input } from '@builder.io/sdk';
 
 function humanCase(str = '') {
   if (str.includes('$')) {
@@ -85,7 +87,7 @@ registerDataPlugin(
         const contentTypesResponse = await Stack.getContentTypes();
         // `any` override is to fix https://github.com/contentstack/contentstack-javascript/pull/61
         const contentTypes = (contentTypesResponse as any).content_types as ContentType[];
-        const contentTypesWithReferences = contentTypes.map(contentType => {
+        const augmentedContentTypes = contentTypes.map(contentType => {
           const references = contentType.schema.filter(field => field.data_type === 'reference');
           // https://www.contentstack.com/docs/developers/apis/content-delivery-api/#include-reference
           const referenceSearchParams = references.map(field => field.uid);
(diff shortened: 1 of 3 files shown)


plugins/contentstack/src/plugin.ts
```diff
@@ -76,57 +78,127 @@ registerDataPlugin(
...
             name: humanCase(model.title),
             id: model.uid,
             description: model.description,
-            inputs: () => [{ name: 'limit', type: 'number', defaultValue: 10 }],
+            inputs: () => {
+              const subFields = model.schema
+                // searching by ref UID seems pointless, so we exclude that for now.
+                // In the future, we could add searching within references:
+                // https://www.contentstack.com/docs/developers/apis/content-delivery-api/#reference-search-equals
+                .filter(field => field.data_type !== 'reference')
```
teleaziz, line 112: what do you think of only picking `text`, `number`, `boolean`  ? the rest would be harder to guarantee we have editors for it, and there might be some custom types on the contentStack side that doesn't map well to Builder types

plugins/contentstack/src/plugin.ts
```diff
@@ -117,14 +118,23 @@ registerDataPlugin(
                 )}&${buildHeaders()}`;
               };
 
-              const envName = `environment=${environmentName}`;
               if (options.entry) {
-                return buildUrl(
-                  `entries/${options.entry}?${envName}&${model.referenceSearchParams}`
+                const query = qs.stringify(
```
Dkendal: I assume we're using this because [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) doesn't support IE, is that correct?

I don't know if we actually support IE, so you might be able to use that instead.

plugins/contentstack/src/plugin.ts
```diff
@@ -112,24 +111,62 @@ registerDataPlugin(
...
+                .filter(field => field.data_type !== 'reference')
+                .map(
+                  (field): Input => ({
+                    name: field.uid,
+                    friendlyName: field.display_name,
+                    type: field.data_type,
+                  })
+                );
+
+              const inputs: Input[] = [
```
Dkendal, line 121: This big block could probably be moved to function like `buildInputs` in the same module.

Dkendal (approved):
looks good, just had the question about adding qs as a dependency. If it's the case the URLSearchParams doesn't fit this need then how does `qs` compare to the npm package [`query-string`](https://github.com/sindresorhus/query-string). My hesitation with `qs` is that it is [quite large, relatively speaking](https://bundlephobia.com/package/qs@6.10.1)

plugins/contentstack/src/plugin.ts
samijaber: I actually ended up replacing `qs` with `URLSearchParams` for convenience!

plugins/contentstack/src/plugin.ts
samijaber: Done in 27e07d1

plugins/contentstack/src/plugin.ts
samijaber: (Ugh, wrote something here but it didn't send)

I think the default text editor will work for pretty much everything, because all fields are searched in the same way using regex: https://www.contentstack.com/docs/developers/apis/content-delivery-api/#search-by-regex

There are also tons of data types, most of which sound like variations of `text`:
image

I would prefer to ship it as-is, and keep an eye on user feedback. If we hear that some fields aren't being searched properly, we can go in and adjust those particular data-types. The alternative is to test each data type which is going to take a bit too much time IMO, I'd rather see how important it is.

PR #579, opened 2021-09-22
fix interface in nextjs-simple

teleaziz:


mrkoreye (approved):

Ifunanyacollins (commented):
Hi @teleaziz can you also reflect this change on the website here is the link https://www.builder.io/blog/visual-next-js

PR #583, opened 2021-09-29
[vtex] do not require environment as input

teleaziz:
Environment input was superfluous and it should be always pointing to the stable vtex env

### DIFF
--- a/plugins/vtex/src/plugin.ts
+++ b/plugins/vtex/src/plugin.ts
@@ -1,4 +1,5 @@
 import { registerCommercePlugin } from '@builder.io/commerce-plugin-tools';
+import { Resource } from '@builder.io/commerce-plugin-tools/dist/types/interfaces/resource'
 import pkg from '../package.json';
 
 registerCommercePlugin(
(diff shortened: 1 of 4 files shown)


mandx (approved):

plugins/vtex/src/plugin.ts
```diff
@@ -37,14 +31,11 @@ registerCommercePlugin(
...
     const accountName = settings.get('accountName');
-    const isDev = settings.get('environment') === 'development';
     const environment = 'vtexcommercestable';
 
     const baseUrl = (url: string) => {
       const endUrl = `https://${accountName}.${environment}.com.br/${url}`;
-      return `${
-        isDev ? 'http://localhost:5000' : 'https://builder.io'
-      }/api/v1/proxy-api?url=${encodeURIComponent(endUrl)}`;
+      return `https://builder.io/api/v1/proxy-api?url=${encodeURIComponent(endUrl)}`;
```
samijaber, line 38: Can we use `apiRoot` so that this works locally as well?

```suggestion
      return `${appState.config.apiRoot()}/api/v1/proxy-api?url=${encodeURIComponent(endUrl)}`;
```

steve8708:
would also highly suggest adding `helperText` to these fields, even if just a link to vtex docs or a simple description or ideally where to grab those values

PR #586, opened 2021-10-04
[shopify-plugin] interpolate ids directly and not rely on evaluating URL on the backend

teleaziz:


mandx (approved):

plugins/shopify/src/editors/ShopifyCollectionPicker.tsx
```diff
@@ -204,7 +204,7 @@ export class ShopifyCollectionPicker extends SafeComponent<ShopifyCollectionPick
     return {
       '@type': '@builder.io/core:Request',
       request: {
-        url: `${appState.config.apiRoot()}/api/v1/shopify/collections/{{this.options.collection}}.json?apiKey=${
+        url: `${appState.config.apiRoot()}/api/v1/shopify/collections/${collectionId}.json?apiKey=${
```
samijaber, line 207: If we're interpolating directly, I think we can get rid of the `options` object below, right? Or is it used for something else?

https://github.com/BuilderIO/builder/blob/3bf85e43c5dc9d307267d993f2e1ff5feb3845b5/plugins/shopify/src/editors/ShopifyCollectionPicker.tsx#L211-L213

plugins/shopify/src/editors/ShopifyProductPicker.tsx
```diff
@@ -258,7 +258,7 @@ export class ShopifyProductPicker extends SafeComponent<ShopifyProductPickerProp
     return {
       '@type': '@builder.io/core:Request',
       request: {
-        url: `${appState.config.apiRoot()}/api/v1/shopify/products/{{this.options.product}}.json?apiKey=${
+        url: `${appState.config.apiRoot()}/api/v1/shopify/products/${productId}.json?apiKey=${
```
samijaber, line 261: Same thing here for the `options` object

samijaber (approved):

plugins/shopify/src/editors/ShopifyProductPicker.tsx
teleaziz: it's used by the plugin to identify the current value

PR #606, opened 2021-10-21
hide mutation from insert menu by default

tgaribal:
set Mutation element hidden from Insert menu by default


teleaziz:
merging to be included in the pending release

PR #651, opened 2021-12-02
fix: little fixes in create-builder

manucorporat:
- Minimun version
- Hide more npm logging
- Not open browser by create react app


mrkoreye:
🎉

PR #652, opened 2021-12-03
Fix: vue SDK

samijaber:
## Description

- @builder.io/sdks: Adds a missing re-export of `is-editing`, which is used by our vue-nuxt starter: b6f36c745abb441235f97e0c0aec9d5bc42a39aa
- @builder.io/sdks: updates the version of `mitosis-cli` that is imported to 0.0.8, which I think was our actual intention (the lib breaks when trying to develop with the current 0.0.6 version): 7d77c5484041f510b0a8608314172bc675679bdc

## TO-DO

- [x] build and deploy new version of the generated SDKs
- [x] update version of @builder.io/sdk-vue used by vue-nuxt starter
- [ ] merge this PR


packages/sdks/overrides/react-native/src/index.ts
```diff
@@ -12,6 +12,7 @@ export { default as Section } from './blocks/section.lite';
 export { default as Fragment } from './blocks/fragment.lite';
 export { default as RenderContent } from './components/render-content.lite';
 
+export * from './functions/is-editing';
```
samijaber, line 15: Did I need to add this line to all of the `overrides/*` files as well? I did it to follow the convention but I don't fully understand what's happening here.

steve8708 (approved):

packages/sdks/overrides/react-native/src/index.ts
steve8708: yeah probably, overrides are a heavy handed way to totally replace a file per output target

manucorporat:
is the content of this PR mostly generated code?

samijaber:
> is the content of this PR mostly generated code?

@manucorporat yup! Except for the files I added comments to, everything is auto-generated. :)

PR #660, opened 2021-12-08
[readme] pass addToCart to context

teleaziz:
this will help answer a common question we get is `can I add to cart from a builder landing page`


mrkoreye (approved):
🏆

PR #669, opened 2021-12-17
fix: nextjs starter

manucorporat:


starters/create-builder/nextjs/components/SourceCodeLink.tsx
```diff
@@ -6,10 +6,10 @@ export interface SourceCodeLinkProps {
   column?: number;
 }
 
-export const SourceCodeLink: FunctionComponent<SourceCodeLinkProps> = ({fileName, line, column, children, ...rest}) => {
+export const SourceCodeLink: FunctionComponent<SourceCodeLinkProps> = ({children}) => {
```
samijaber, line 9: I'm confused by this change. We're no longer using any of the props from `SourceCodeLinkProps`. Are they not needed anymore? Can we remove the Props?

PR #670, opened 2021-12-17
actions for react native

steve8708:


packages/sdks/overrides/vue/src/functions/get-block-actions.ts
```diff
@@ -0,0 +1,18 @@
+import { evaluate } from './evaluate';
+
+export function getBlockActions(options: { block: any; context: any; state: any }): any {
+  const obj: any = {};
+  if (options.block.actions) {
+    for (const key in options.block.actions) {
```
mandx, line 6:
```suggestion
    for (const key of options.block.actions) {
```
or
```suggestion
    for (const key in options.block.actions) {
      if (!options.block.actions.hasOwnProperty(key) {
        continue;
      }
```

mandx (approved):

packages/sdks/overrides/vue/src/functions/get-block-actions.ts
samijaber: Thanks for the pointer @mandx , I'll factor this in before publishing Steve's changes 🙏🏽

PR #675, opened 2021-12-23
[preview-logic] add plugin to compute preview url based on content fi…

teleaziz:
…elds and targeting attributes

# Builder.io Preview Logic plugin

This plugin allow advanced computation of the preview url inside Builder's editor, useful for cases where you have a special routing logic in your preview site based on the content [`custom fields`](https://www.builder.io/c/docs/custom-fields) or [`targeting attributes`](https://www.builder.io/c/docs/guides/targeting-and-scheduling#custom-targeting), for example routing based on locales, or routing a blog post based on a `slug` custom field.

## Installation

Go to [builder.io/account/space](https://builder.io/account/space) and type `@builder.io/plugin-preview-logic` in the text input, then hit save.

Installation screenshot

After hitting save, the app will reload with the plugin code installed,

Now from the same menu, hit `edit plugin settings`

Plugin Settings screenshot

Add a preview setting by hitting `+ preview setting`

Pick the model you'd like to add the preview setting to

Model picker screenshot

Write the code you'd like to run to compute the preview url, in this code you'll have access to three objects:

- `content`: a json representation fo the current state of content, access properties directly for example: `content.data.title`
- `space`: the space settings: siteUrl, name, publicKey
- `targeting` an object representing the content targeting settings, for example: `targeting.urlPath`

Always return a string representing the final preview URL you'd like the app to use.

### DIFF
--- a/plugins/preview-logic/.gitignore
+++ b/plugins/preview-logic/.gitignore
@@ -1,0 +1,12 @@
+node_modules
+coverage
+.nyc_output
+.DS_Store
+*.log
+.vscode
+.idea
+dist
+compiled
+.awcache
+.rpt2_cache
+docs
--- a/plugins/preview-logic/.travis.yml
+++ b/plugins/preview-logic/.travis.yml
@@ -1,0 +1,20 @@
+language: node_js
+cache:
+  directories:
+    - ~/.npm
+notifications:
+  email: false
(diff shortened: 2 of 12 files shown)


teleaziz:
closed in favor of a native feature

PR #683, opened 2022-01-04
Support pixel for native

steve8708:
Our current pixel object assumes the environment is the DOM and sends it as

```
{ tagName: 'img', properties: { src: '..' } }
```

but we need a native compatible one for native SDKs, so this PR adds that and also gives us the option to move all pixels to a naive compatible one

### DIFF
--- a/packages/react/src/blocks/Image.tsx
+++ b/packages/react/src/blocks/Image.tsx
@@ -283,6 +283,7 @@ class ImageComponent extends React.Component<any, { imageLoaded: boolean; load: 
       srcset = this.getSrcSet();
     }
 
+    const isPixel = this.props.builderBlock?.id.startsWith('builder-pixel-');
     const { fitContent } = this.props;
 
     return (
@@ -331,7 +332,7 @@ class ImageComponent extends React.Component<any, { imageLoaded: boolean; load: 
                   },
                 }),
               }}
-              loading="lazy"
+              loading={isPixel ? 'eager' : 'lazy'}
               className={'builder-image' + (this.props.className ? ' ' + this.props.className : '')}
               src={this.image}
               {...(!amp && {
--- a/packages/sdks/overrides/react-native/src/functions/transform-block.ts
+++ b/packages/sdks/overrides/react-native/src/functions/transform-block.ts
@@ -1,7 +1,8 @@
 import type { BuilderBlock } from '../types/builder-block';
 
 export function transformBlock(block: BuilderBlock): BuilderBlock {
-  // Map the DOM-based pixel format to a native compatible one
+  // Map the DOM-based pixel format to a native compatible one.
+  // react-native doesn't support DOM specific fields like `tagName` and `properties`, but any browser framework (e.g. Vue, etc.) do
   if (block.id.startsWith('builder-pixel-')) {
     return {
       ...block,
--- a/packages/sdks/overrides/react-native/src/functions/transform-block.ts
+++ b/packages/sdks/overrides/react-native/src/functions/transform-block.ts
@@ -1,4 +1,6 @@
-export function transformBlock(block: any) {
+import type { BuilderBlock } from '../types/builder-block';
+
+export function transformBlock(block: BuilderBlock): BuilderBlock {
   // Map the DOM-based pixel format to a native compatible one
   if (block.id.startsWith('builder-pixel-')) {
     return {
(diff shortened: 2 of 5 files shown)


packages/react/src/blocks/Image.tsx
```diff
@@ -331,7 +332,7 @@ class ImageComponent extends React.Component<any, { imageLoaded: boolean; load:
                   },
                 }),
               }}
-              loading="lazy"
+              loading={isPixel ? 'eager' : 'lazy'}
```
mandx, line 335: TIL `lazy` is not used with JS disabled: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading

packages/sdks/src/functions/transform-block.ts
```diff
@@ -0,0 +1,4 @@
+// Noope way for targets to make modifications to the block object if/as needed
+export function transformBlock(block: any) {
```
mandx, line 2:
```suggestion
export function transformBlock<T>(block: T): T {
```

packages/sdks/overrides/react-native/src/functions/transform-block.ts
```diff
@@ -0,0 +1,15 @@
+export function transformBlock(block: any) {
```
mandx, line 1:
```suggestion
export function transformBlock<T>(block: T): T {
```

mandx (approved):

packages/react/src/blocks/Image.tsx
steve8708: oh ya JS is no the concern here, the concern is purely that we need the pixel to always eager load no matter what, even if not in the viewport

packages/sdks/src/functions/transform-block.ts
steve8708: block will always be a BuilderElement but we need to make an update to the mitosis CLI to be able to support types wit the way we do transforms (will be tricky so just using `any` in the meantime). I can update with a note on this though

packages/react/src/blocks/Image.tsx
steve8708: that's interesting tho that lazy doesn't work with JS disabled, definitely unexpected

mandx: yeah yeah, it's not related to the PR :D it's just a broken assumption of mine of how native lazy loading works

packages/sdks/overrides/react-native/src/functions/transform-block.ts
samijaber, line 1: Can you elaborate/add a comment about why this override is only for react-native and not for vue?

steve8708: it's this https://github.com/BuilderIO/builder/blob/native-pixel/packages/sdks/overrides/react-native/src/functions/transform-block.ts#L4:L4

I didn't put it at the top bc we may do other transforms, so that specific transform is specifically bc react native doesn't support DOM specific fields like `tagName` and `properties`, but any browser framework (e.g. Vue, etc) does

steve8708: I can elaborate on that comment if you have any suggestions on how it can be more clear

samijaber: Thanks, that context is useful!

> I didn't put it at the top bc we may do other transforms

Good call.

Just copy pasting your github comment above into the code would be great! I added a suggestion below.

packages/sdks/overrides/react-native/src/functions/transform-block.ts
```diff
@@ -0,0 +1,15 @@
...
+  // Map the DOM-based pixel format to a native compatible one
```
samijaber, line 4:
```suggestion
  // Map the DOM-based pixel format to a native compatible one
  // react-native doesn't support DOM specific fields like `tagName` and `properties`, but any browser framework (e.g. Vue, etc.) does
```

packages/sdks/src/functions/transform-block.ts
steve8708: found a better way, now uses the BuilderBlock type like other areas (IDE will complain about the import path from the overrides folder but that is a common issue)

we could move to using absolute imports come to think of it to fix the types issue. so anything in overrides imports from like `@/functions/foo` instead of `../functions/foo` (which often doesn't exist in the `overrides/*` folders so vscode will complain 

hadn't thought of that until just now. something we can do in the future to improve the DX cc @samijaber (tho will have to update the mitosis CLI stuff to handle so is no small job but can do sometime in the future when we have time)

PR #676, opened 2021-12-26
Export RenderBlock in the React Native SDK

saary:
## Export RenderBlock in the React Native SDK
RenderBlock is an important base construct that is needed in order to introduce higher level constructs such as containers.
Exposing it allows to build sophisticated custom containers.

PR #685, opened 2022-01-05
Export renderblock rn sdk

samijaber:
## Description

Reinstates #676, and:

- adds `getEventHandlerName` fn override to reduce duplication
- removes redundant `index.ts` override for react-native

TO-DO:
- [x] publish RN SDK
- [x] publish Vue SDK


packages/sdks/overrides/vue/src/functions/get-block-actions.ts
```diff
@@ -1,26 +0,0 @@
...
-}): any {
-  const obj: any = {};
-  if (options.block.actions) {
-    for (const key in options.block.actions) {
-      if (!options.block.actions.hasOwnProperty(key)) {
-        continue;
-      }
-
-      const value = options.block.actions[key];
-      obj['v-on:' + key.toLowerCase()] = (event: Event) =>
```
samijaber, line 16: This file as duplicated just for the sake of this key value, so I added a separate fn `getEventHandlerName` and overwrote just that instead.

steve8708 (approved):

packages/sdks/output/vue/nuxt2/src/blocks/columns.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-  <div class="builder-columns div-1083i9ahz0v">
+  <div class="builder-columns div-38o6qlysr38">
     <div
-      class="builder-column div-1083i9ahz0v-2"
+      class="builder-column div-38o6qlysr38-2"
```
steve8708, line 4: one random thing we should do at some point is use hashes for this instead of stringified random numbers so these are deterministic (woon't keep showing up in our diffs)

steve8708: this is unrelated to this PR just fYI in case you are seeing these and thinking its janky, it is solvable whenever we have a moment :D

packages/sdks/overrides/vue/src/functions/get-block-actions.ts
steve8708: awesome

steve8708: great idea

packages/sdks/output/vue/nuxt2/src/blocks/columns.vue
samijaber: Oh yeah, I saw the line of code and just need to get around seeding it with a constant so it stops doing that. It's been driving me nuts 😄

samijaber: https://github.com/BuilderIO/mitosis/blob/a7d1f681d0cb8356054380f6ea47c6b19eac22e5/packages/cli/src/build/build.ts#L38

PR #686, opened 2022-01-05
fix(examples): update react version

mrkoreye:
## Description

Update the Builder React package in the next-js-simple starter to the latest version


mrkoreye:
Closing because we already use the latest react version, I just had an old version locally

PR #687, opened 2022-01-05
make sure that if we send pixels in the new format they work too

steve8708:


packages/sdks/overrides/react-native/src/functions/transform-block.ts
```diff
@@ -3,7 +3,7 @@ import type { BuilderBlock } from '../types/builder-block';
 export function transformBlock(block: BuilderBlock): BuilderBlock {
   // Map the DOM-based pixel format to a native compatible one.
   // react-native doesn't support DOM specific fields like `tagName` and `properties`, but any browser framework (e.g. Vue, etc.) do
-  if (block.id.startsWith('builder-pixel-')) {
+  if (block.id.startsWith('builder-pixel-') && !block.component) {
```
samijaber, line 6: Could you expand the comment above the code to explain why this check is needed? It isn't clear to me at face-value

samijaber (approved):

packages/sdks/overrides/react-native/src/functions/transform-block.ts
steve8708, line 6:
```suggestion
  // Additionally, we only want to convert the DOM pixel format (looks like `{ tagName: 'img', properties: { src: '...' } }`)
  // but we may move to a DOM-less pixel format that uses `{ component: { name: 'Image', ... } }` so if a `component` property is provided
  // we assume this accounts for native as well and we should not transform this block
  if (block.id.startsWith('builder-pixel-') && !block.component) {
```

steve8708: @samijaber better?

samijaber: Yes, perfect. Thanks 👍🏽

PR #701, opened 2022-01-12
Feature/afosto-plugin

gijsbotje:
Afosto plugin to retrieve data from a webshop.

Enter the shop url in the configuration and use the fieldtypes to get products and categories

### DIFF
--- a/plugins/afosto/.gitignore
+++ b/plugins/afosto/.gitignore
@@ -1,0 +1,12 @@
+node_modules
+coverage
+.nyc_output
+.DS_Store
+*.log
+.vscode
+.idea
+dist
+compiled
+.awcache
+.rpt2_cache
+docs
--- a/plugins/afosto/.travis.yml
+++ b/plugins/afosto/.travis.yml
@@ -1,0 +1,20 @@
+language: node_js
+cache:
+  directories:
+    - ~/.npm
+notifications:
+  email: false
+node_js:
+  - '10'
+  - '11'
+  - '8'
+  - '6'
+script:
+  - npm run test:prod && npm run build
+after_success:
+  - npm run travis-deploy-once "npm run report-coverage"
+  - if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then npm run travis-deploy-once "npm run deploy-docs"; fi
+  - if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then npm run travis-deploy-once "npm run semantic-release"; fi
+branches:
+  except:
+    - /^v\d+\.\d+\.\d+$/
--- a/plugins/afosto/CONTRIBUTING.md
+++ b/plugins/afosto/CONTRIBUTING.md
@@ -1,0 +1,17 @@
+We're really glad you're reading this, because we need volunteer developers to help this project come to fruition. 👏
+
+## Instructions
+
+These steps will guide you through contributing to this project:
+
+- Fork the repo
+- Clone it and install dependencies
+
+      		git clone https://github.com/YOUR-USERNAME/typescript-library-starter
+      		npm install
+
+Keep in mind that after running `npm install` the git repo is reset. So a good way to cope with this is to have a copy of the folder to push the changes, and the other to try them.
+
+Make and commit your changes. Make sure the commands npm run build and npm run test:prod are working.
+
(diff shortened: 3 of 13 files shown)


samijaber (changes requested):
Hey @gijsbotje, thanks for setting up this new plugin!

We published a dev version of it at `@builder.io/plugin-afosto@0.0.1`. We added it and tried importing products, that worked just fine. But when we tried with categories, we got no results. Even though the search auto-suggest returns a lot of results with categories. Could you take a look on your end and see why that's happening?

Once that's resolved, we should be able to get this merged for you 👍🏽.

gijsbotje:
@samijaber Thanks for checking it. What is it you expect the categories to return? I haven't created a component for a category yet, so I haven't thoroughly tested it.
If the issue is that it doesn't show options when the field is opened, that's just how our suggest endpoint works. It needs a `q` value to find pages/collections. Right now it just fetches data from the actual shop, in the future it should use an API, but we aren't there yet. It also returns blank when you do the same request within, I think, 300ms.

When I test the field `Afosto category handle` I can search for pages and select and save it.

I did add a header to the `getRequestObject` functions to fetch the JSON instead of HTML. Maybe that was what you were missing :)

gijsbotje:
@samijaber do you have some documentation of what is possible with the getRequestObject function? could I also transform the response to something that would be a stripped version of what is returned now?

gijsbotje:
@samijaber Any estimate when this will be reviewed again?

samijaber:
Hey @gijsbotje, I was away for the past 2 weeks, hence the slowness in responses. Thanks for your patience!

Here's a screencast outlining what I'm seeing:
https://www.loom.com/share/4b23a91b53e4413fb33299ceb6dbfb4b

I think one confusing thing for us is that the same API endpoint is used to search for both products & collections. It looks like, if `q` is empty, then it return all products, but if it has a `q=` value, then it returns a filtered set of collections. We would expect 2 separate endpoints, each allowing to search (or get all) values for both products & collections.

> do you have some documentation of what is possible with the getRequestObject function? could I also transform the response to something that would be a stripped version of what is returned now?

We don't currently have that documented. I will work on getting some documentation up and running for you soon.

samijaber (changes requested):
@gijsbotje Thanks for the explanation.

In its current state, this plugin does not have all the features we expect. We prefer to wait until bugs are resolved (i.e. both products & collections have a working search endpoint) before publishing a plugin and integrating it into our ecosystem. So unfortunately, this means we won't be able to approve this PR at this time.

I will keep this PR open, and feel free to ping me whenever the necessary changes have been made!

gijsbotje:
@samijaber I've created tasks for my colleague to take a look at what we can change to better suite your expectations. Will get back to you when/if we change it.

gijsbotje:
@samijaber We tweaked the endpoint a little, could you guys check it out?

samijaber:
@gijsbotje Thanks! I just tested it on my end, and the search seems to work properly for both products & categories! 👍🏽

samijaber:
@gijsbotje I just took another look and tried to use your plugin in a demo Builder space. Searching for the data works perfectly! 

But when it comes to using the data in the editor, there is one big issue: the requests stored by the plugin return _very_ large product & category files. I tried a few as examples and they were 2.5-3mb in size. Unfortunately, this makes the dynamic data feature really impractical to use with Afosto data.

The reason for that is that our data bindings code will go through all of the data, and make every single value in you data a potential item in the list. There are about 50,000-60,000 values in each one of your products/categories.

I have recorded a loom to show you: I start by using the contentstack plugin to show how other plugins work, and then tried the Afosto plugin. https://www.loom.com/share/f02f1cc6ef684a88a76f7ca4ffcc0ab5

I'm not sure how you want to approach resolving this on your end. Ultimately, the URLs stored by your plugin need to return much smaller pieces of content, otherwise, the plugin won't be practical to use. 🤔

For reference, here are examples of the URLs that I am talking about (in `curl` form). They're the values generated by `getRequestObject`:

- product: `curl --request GET  --url 'https://www.azalp.nl/keter-17192190-manor-43-2e-kans-128-103-1-2.htm' --header 'Content-Type: application/json'`
- category: `curl --request GET  --url 'https://www.azalp.nl/filter-flosser/' --header 'Content-Type: application/json'`

mhevery:
Hey, there is no activity on this PR. Can we get this PR into mergeable state (or should we abandon it and close it)

samijaber:
Closing due to inactivity.

PR #702, opened 2022-01-12
Examples: vue-storefront

samijaber:
## Description

- Adds the skeleton for a vue-storefront example
- fixes a bug in the vue SDK package.json exports

This commit range shows the diff without the very big initialization of vue-storefront: https://github.com/BuilderIO/builder/pull/702/files/ec801c061240cfb23be4995e07a53e08dc6a9185..f0acab6a87bf63f0c9271b84e40b18002bf48605


examples/vue-nuxt/package.json
```diff
@@ -10,7 +10,7 @@
     "generate": "NODE_PRESERVE_SYMLINKS=1 nuxt generate"
   },
   "dependencies": {
-    "@builder.io/sdk-vue": "0.0.1-44",
+    "@builder.io/sdk-vue": "^0.0.1-47",
```
Dkendal, line 13: given that this is a v0 package and there will be a lot of change it might make sense to lock it to an exact version to guarantee that it continues to work

Dkendal (approved):

PR #746, opened 2022-01-26
ENG 1962 prototype vulnerability

Dkendal:
## Description

Fixes a prototype pollution vulnerability in the QueryString class.

While the using `Object.create(null)` is sufficient to prevent global
pollution I've also change the behaviour such that it is now an error to
parse path segments "__proto__" and "prototype" to alert you to
potential prototype pollution attempts and prevent objects created with
`deepen` from being merged into other object that aren't hardened against this
attack.

BREAKING CHANGE: The QueryString class will now throw an error if you
attempt to parse a string that contains the path segments `__proto__` or
`prototype` to warn against prototype pollution and prevent upstream
attacks.

## Changelist

- chore(core): add jest
- chore(core): add jest configuration
- chore(core): add typescript transform for jest
- test(core): add failing test to demonstrate issue ENG-1962
- fix(core)!: prevent prototype pollution
- refactor(core): use for loop over while+shift


packages/core/Earthfile
```diff
@@ -0,0 +1,25 @@
...
+build:
+  FROM +prepare
+  RUN npm run build
+  SAVE ARTIFACT dist AS LOCAL dist
+  SAVE IMAGE builder.io/core:latest
+
+test:
+  FROM +prepare
+  COPY jest.config.ts .
+  RUN npm run test
```
samijaber, line 25: 👀 Mind outlining what this is exactly? Is it a replacement for npm scripts?

packages/core/src/classes/query-string.class.ts
```diff
@@ -71,3 +74,8 @@ export class QueryString {
     return _res;
   }
 }
+
+function assertAllowedPropertyName(name: string): asserts name {
+  if (PROPERTY_NAME_DENY_LIST.indexOf(name) >= 0)
+    throw new Error(`Property name "${name}" is not allowed`);
+}
```
samijaber, line 81: TIL about https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions!

In our case, you're doing `assertAllowedPropertyName(part)` and then asserting just that string value. From reading the TS docs, shouldn't it be closer to:

```ts
assertAllowedPropertyName(!PROPERTY_NAME_DENY_LIST.includes(name), name)

function assertAllowedPropertyName(condition: boolean, name: string): asserts condition {
  if (!condition)
    throw new Error(`Property name "${name}" is not allowed`);
}
```

where we pass the condition as the argument? I've never used these assertions so not fully sure how they're meant to work!

samijaber (approved):

packages/core/src/classes/query-string.class.ts
Dkendal: Not quite, the assertion function can be used to evaluate an expression, and inform the compiler that that expression must be true for the rest of the evaluation - but it can also be used to narrow a type.

In the narrowing example it's saying that a type can be cast what ever the assert is specifying with `is`, if `is` is missing then it's not narrowing but more of a developer hint that this might throw.

If typescript's type system allowed what I would actually be specifying is `asserts name is not "__proto__" | "prototype"`, but it doesn't.

Dkendal: there's some other examples here https://www.typescriptlang.org/play#example/assertion-functions

packages/core/Earthfile
Dkendal: It's like docker - it doesn't necessarily have be in this PR but it will be a part of the internal build system and if I have to build these packages it will be required at some point.

tl;dr lets you build a docker image and documents what the steps are to build this package.

mrkoreye (approved):

mandx (approved):

PR #757, opened 2022-02-01
Bug/defaulterrorpage rerender

codercatdev:
## Description

DefaultErrorPage was leaving styles that were not good for builder's BuilderComponent when previewing. This allows the DefaultErrorPage to be fully removed on preview.

_Screenshot_

### DIFF
--- a/starters/create-builder/nextjs/pages/[[...page]].tsx
+++ b/starters/create-builder/nextjs/pages/[[...page]].tsx
@@ -1,15 +1,15 @@
-import type { GetStaticPropsContext, InferGetStaticPropsType } from 'next';
-import { useRouter } from 'next/router';
-import { BuilderComponent, Builder, builder } from '@builder.io/react';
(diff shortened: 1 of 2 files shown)


starters/create-builder/nextjs/pages/[[...page]].tsx
```diff
@@ -63,9 +65,6 @@ export default function Page({
       <Head>
```
Comment (teleaziz) [inline starters/create-builder/nextjs/pages/[[...page]].tsx:65]: I think we should keep the  `  <meta name="robots" content="noindex" />` in the server rendered string, if no page:

```

  return (
    <>
      <Head>
       { !page &&   <meta name="robots" content="noindex" />}
        <meta name="viewport" content="width=device-width, initial-scale=1" />
      </Head>
      <BuilderComponent model="page" content={page} />
    </>
  );


```

teleaziz (changes requested):

codercatdev:
Closing for now, probably need to revisit for a better way of doing this.

PR #503, opened 2021-07-16
JSX Lite generated SDKs

steve8708:
This is to add feature parity for all frameworks we support, and support many more with ease. This adds alpha support for

- Vue
- React Native
- SolidJS

And opens up the possibilities to easily support many other frameworks like Qwik, Svelte, etc with full feature parity and best performance

PR #758, opened 2022-02-02
Update json-schema README.md

sschwartz-builder:
Added setup commands for json-schema package.

### DIFF
--- a/packages/json-schema/README.md
+++ b/packages/json-schema/README.md
@@ -16,7 +16,7 @@ builder content schema:
 }
 ```
 
-## Setup
+## Local Setup
 
 ```
 $ npm install  # install the JS dependencies
--- a/packages/json-schema/README.md
+++ b/packages/json-schema/README.md
@@ -16,6 +16,14 @@ builder content schema:
 }
 ```
 
+## Setup
+
+```
+$ npm install  # install the JS dependencies
+$ mkdir dist  # we need to create the destination dir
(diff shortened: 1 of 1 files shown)


packages/json-schema/README.md
```diff
@@ -16,6 +16,14 @@ builder content schema:
 }
 ```
 
+## Setup
```
mrkoreye, line 19:
```suggestion
## Local Setup
```
I think we might want to point out that these steps are not necessary in order to use the schema, but only if you want to generate it locally from.

mrkoreye (changes requested):
Looking at this a little more, I think maybe we should actually add the `dist` file to this repo as well so the schema setup we describe actually works correctly without people having to download and generate it themselves.

mrkoreye:
@steve8708 I am not super familiar with this json schema stuff, was there a reason the `dist` directory with the `json` was removed in this PR: [packages/json-schema/dist/BuilderContent.json](https://github.com/BuilderIO/builder/pull/503/files#diff-23962119e4935dca4eb19a87cd9ff5b2f54179b6bf2fee762f04bcb1fe6d6dcd)?

mandx:
> was there a reason the dist directory with the json was removed

@mrkoreye maybe there was an assumption that running the Makefile will create the `dist/` directory or something like that, but as @sschwartz-builder found out, this is not the case.

PR #840, opened 2022-02-15
chore: Update versions to get Angular SDK to build

shyam-builder:
## Description

Get the Angular SDK for Builder IO to successfully `build` with the latest Angular version. Bump up dependencies as needed. 

Published a dev release, and tested against both the client side and the Universal application.

Had to rewrite the Angular Universal application as per the latest docs in Angular.


samijaber (commented):
Looks like you used both `npm install` and `yarn`. I think it would be best to pick one and remove the others' lock file to avoid any confusion for future developers 🙏🏽

packages/angular/src/app/foo.component.ts
```diff
@@ -0,0 +1,9 @@
+import { Component } from '@angular/core';
+
```
murtaza-builder, line 2: Is this component needed?

packages/angular/src/app/app.component.html
```diff
@@ -1 +1,5 @@
+<a routerLink="/locale-test">AAAAA</a>
```
murtaza-builder, line 1: these anchors would be just for testing right?

packages/angular/src/app/app-routing.module.ts
```diff
@@ -1,7 +1,11 @@
 import { NgModule } from '@angular/core';
 import { Routes, RouterModule } from '@angular/router';
+import { FooComponent } from './foo.component';
 
-const routes: Routes = [];
+const routes: Routes = [{
+  path: '**',
+  component: FooComponent
```
murtaza-builder, line 7: aah. here it is!

steve8708 (approved):

packages/angular/package.json
```diff
@@ -21,70 +21,71 @@
     "_install": "npm install",
     "ng": "ng",
     "start": "npm run develop:package",
-    "_build": "ng build --prod",
+    "_build": "ng build --configuration production",
```
sanyamkamat, line 24: Is the configuration part needed since the default is production?
Also, were both the configurations tested?

packages/angular/package.json
```diff
@@ -1,6 +1,6 @@
 {
```
sanyamkamat, line 1: Regarding SDK development, do we check if the SDK/examples work with different Node versions?

PR #852, opened 2022-02-16
refactor and simplify shopify plugin to onlly require domain / access…

teleaziz:
…token


plugins/shopify/src/plugin.ts
```diff
@@ -0,0 +1,88 @@
...
+        name: 'storefrontAccessToken',
+        type: 'string',
+        helperText:
+          'Required to sync, index, and cache your storefront product data to avoid rate limits',
+        required: true,
+      },
+      {
+        name: 'storeDomain',
+        type: 'text',
+        helperText: 'your-store.myshopify.com',
```
samijaber, line 23: In my experience, people sometimes get confused and only put the `your-store` part, so I'd maybe say

```suggestion
        helperText: 'Your entire store domain, such as "your-store.myshopify.com"',
```

or something like that

samijaber (approved):

PR #853, opened 2022-02-16
SDKs: Add feature support matrix

samijaber:
## Description

Wanted to consolidate knowledge for users on what is/isn't implemented in the beta SDKs.


packages/sdks/README.md
```diff
@@ -1,3 +1,36 @@
...
+| SSR                                       |                                                          | ✅  |             | N/A          |                      |
+| Children for Custom Components            |                                                          | 🏗   |             | 🏗            |                      |
+| Dynamic Data Bindings                     |                                                          | ⚠️  |             | ⚠️           |                      |
+| Animations                                |                                                          | ⚠️  |             | ⚠️           |                      |
+| View Current Draft                        |                                                          | ⚠️  |             | ⚠️           |                      |
+| Symbols                                   |                                                          | ⚠️  |             | ⚠️           |                      |
+| `+ add block` button when starting a page | (for now just drag your first block to the "layers" tab) | ⚠️  |             | ⚠️           |                      |
```
samijaber, line 36: @steve8708 Could you please take a look at this and let me know if there is anything you know of that I may have missed? 🙏🏽

steve8708: love it! I think the only wrong thing is we don't support Vue A/B testing the way we will want to. We'll want to do the [SSG supported way](https://www.builder.io/blog/high-performance-no-code#:~:text=static%20generated%20a%2Fb%20testing) we do with React. So I'd just change that to be a no for now and we can circle back on

Also @teleaziz is a good person to ping re: how the SSG supported a/b testing works

steve8708 (approved):

packages/sdks/README.md
samijaber: Thanks for the heads up, noted.

PR #900, opened 2022-02-24
fix(sdk): respect this.canTrack for cookie

samijaber:
## Description

- Core SDK: adds missing check for `this.canTrack` in one instance of setting the session cookie
- Vue SDK 1.0: adds a missing assignment for `builder.canTrack` within the web components instance of Builder's SDK


packages/core/src/builder.class.ts
```diff
@@ -1130,7 +1130,15 @@ export class Builder {
...
-      this.setCookie(sessionStorageKey, sessionId, datePlusMinutes(30));
+      setTimeout(() => {
+        try {
+          if (this.canTrack) {
+            this.setCookie(sessionStorageKey, sessionId, datePlusMinutes(30));
+          }
+        } catch (err) {
+          console.debug('Cookie setting error', err);
+        }
+      });
```
samijaber, line 1141: This is a similar approach to the one in this same file right below:

https://github.com/BuilderIO/builder/blob/1d88dd10836fa492704c2d6fce46d09c80ded6fe/packages/core/src/builder.class.ts#L1164-L1175

packages/vue/src/components/RenderContent.vue
```diff
@@ -167,6 +167,20 @@ export default {
...
+      }
+    },
+    async ensureWcLoadedAndUpdate() {
+      await this.loadWebComponents();
+      const { onBuilderWcLoad } = window;
+      if (onBuilderWcLoad) {
+        onBuilderWcLoad(BuilderWC => {
+          const builder = BuilderWC.builder;
+          builder.canTrack = !window.builderNoTrack;
+        });
```
samijaber, line 183: This closely follows the approach in the Angular SDK:

https://github.com/BuilderIO/builder/blob/1d88dd10836fa492704c2d6fce46d09c80ded6fe/packages/angular/src/app/modules/builder/components/builder-component/builder-component.component.ts#L146-L154

I did not reuse the entire logic from Angular since it involves other things that could cause breaking changes (such as API Key assignment)

mrkoreye (approved):

samijaber:
Loom of the fix working: https://www.loom.com/share/99a3052591a94931b917f9a6c13c8c2e

PR #923, opened 2022-03-02
[@builder.io/plugin-sfcc-headless] Add field types and data provider configurations for product and categories

teleaziz:


plugins/sfcc-headless/src/CategoriesPicker.tsx
```diff
@@ -0,0 +1,201 @@
...
+              >
+                No categories found
+              </Typography>
+            </div>
+          ))}
+      </div>
+    </div>
+  ));
+};
+
```
gustavocodess, line 167: remove comments?

gustavocodess (approved):

PR #931, opened 2022-03-08
Remove throttle

armgjoka:
## Description

Fix the bug that data from multiple sources are not displayed. Removed @Throttle, because it was invoking the function once every 100ms and we were missing some calls to the datasources (if two different calls within 100ms, the second one was skipped, hence data missing).

### DIFF
--- a/packages/react/src/components/builder-component.component.tsx
+++ b/packages/react/src/components/builder-component.component.tsx
@@ -121,8 +121,6 @@ function decorator(fn: Function) {
   };
 }
 
-const Throttle = decorator(throttle);
-
 const fetchCache: { [key: string]: any } = {};
 
 export interface BuilderComponentProps {
@@ -1192,7 +1190,6 @@ export class BuilderComponent extends React.Component<
   }
 
   // TODO: customizable hm
-  @Throttle(100, { leading: true, trailing: true })
   throttledHandleRequest(propertyName: string, url: string) {
     return this.handleRequest(propertyName, url);
   }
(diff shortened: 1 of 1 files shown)


samijaber (changes requested):
I assume the throttling was there to reduce the number of requests. I don't think we need to completely get rid of it, that might cause a lot more requests than we'd like! Instead, we should fix the throttling logic such that it grabs the latest value within the throttling time window.
I see that we manually wrote `throttle`, so it might be hard to do that. We could dig into alternative libraries that support what I described

samijaber:
https://lodash.com/docs/#throttle could be one solution to this problem. My guess is that our implementation of the `leading: true` option in our `throttle` function is flawed here. It's meant to grab the last value in the time window, but you said it's not. 🤔

armgjoka:
@samijaber we want to throttle the requests to an endpoint. How throttle worked here: 
Request1 (1ms), Request2 (30ms), Request3 (110ms). For this case we would miss the Request2 (throttle wait time is 100 ms). While that would be ok for cases all these three requests go to the same endpoint, the issue here is that these three requests go to different endpoints. So to implement that correctly we would have a 100ms throttle window for each endpoint, in this case all three requests would be executed because they are targeting different endpoints.

mrkoreye:
@steve8708 do you happen to remember what the throttle was solving in this case? Will it cause problems if we remove it?

steve8708:
I think it's safe to remove if we test it out and see if normal editing usage doesn't overly hammer these requests, as I think the editor does some throttling too which may do what we need

would also recommend that if we no longer throttle,then we shouldn't have this method called `throttledHandleRequest` that doesn't actually throttle, and instead just refactor those calls to purely do `this.handleRequest` instead

samijaber (approved):

samijaber:
Oh, one last thing @armgjoka! Would you be able to add this change to the `CHANGELOG.MD` doc? We're working on a way to automate changelogs very soon, but for now it needs to be done manually.

PR #945, opened 2022-03-22
fix(sdks): ENG-2365 pass content id as an unique key

sanyamkamat:
## Description

This PR passes a content id to options so that HMR takes the correct content id to re-render the BuilderContent component

_Loom Link_
https://www.loom.com/share/2cfb7ab0a4e04674844b5291483a40c0

### DIFF
--- a/packages/react/src/components/builder-content.component.tsx
+++ b/packages/react/src/components/builder-content.component.tsx
@@ -88,6 +88,9 @@ export class BuilderContent<ContentType extends object = any> extends React.Comp
     let options = {
       ...(this.props.options || ({} as GetContentOptions)),
     };
+    if (!options.key && this.props.content?.id) {
+      options.key = this.props.content.id;
+    }
     if (this.props.content && !options.initialContent?.length) {
       options.initialContent = [this.props.content];
     }
@@ -188,6 +191,7 @@ export class BuilderContent<ContentType extends object = any> extends React.Comp
         builder.queueGetContent(this.name, this.options).subscribe(
           matches => {
             const match = matches && matches[0];
+            console.log('matches', matches);
             this.setState({
               data: match,
               loading: false,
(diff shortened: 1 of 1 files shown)


packages/react/src/components/builder-content.component.tsx
```diff
@@ -90,6 +90,7 @@ export class BuilderContent<ContentType extends object = any> extends React.Comp
     };
     if (this.props.content && !options.initialContent?.length) {
       options.initialContent = [this.props.content];
+      options.key = this.props.content.id;
```
mrkoreye, line 93: Two possible concerns here:
1. What if there is `initialContent`, don't we still want to set `options.key` to fix for that case with two builder components on the page? It feels like maybe we don't need to tie setting `options.key` to `!options.initialContent?.length` like we do here
2. What if `options` already has a `key` set? We will override it here, which does not seem like the right behavior.

What if instead we added an additional condition like this:
```typescript
get options() {
  let options = {
    ...(this.props.options || ({} as GetContentOptions)),
  };
  if (!options.key && this.props.content?.id) {
    options.key = this.props.content.id;
  }
.....
    
```

sanyamkamat: Thanks @mrkoreye for your comment. I have updated the PR with the checks.

packages/react/src/components/builder-content.component.tsx
```diff
@@ -188,6 +191,7 @@ export class BuilderContent<ContentType extends object = any> extends React.Comp
         builder.queueGetContent(this.name, this.options).subscribe(
           matches => {
             const match = matches && matches[0];
+            console.log('matches', matches);
```
samijaber, line 194:
```suggestion
```

sanyamkamat: Oops. Thanks for the suggestions!

samijaber (approved):

mrkoreye (approved):

PR #959, opened 2022-04-01
fix starters for npm init builder.io

steve8708:
fixes errors causing it to not work due to specific three.js/react versions. also a bit of cleanup (add .gitignore, remove yarn.lock, fix the docs around adding your API key

### DIFF
--- a/starters/create-builder/create-react-app/package.json
+++ b/starters/create-builder/create-react-app/package.json
@@ -8,9 +8,9 @@
     "react-scripts": "4.0.3",
     "typescript": "^4.1.2",
     "@builder.io/react": "^1.1.47",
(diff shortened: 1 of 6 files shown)


ca136 (approved):

samijaber (approved):

samijaber:
our README still mentions installing `npm` btw: https://github.com/BuilderIO/builder/blob/main/starters/create-builder/nextjs/README.md#prerequisites

steve8708:
hm we should really probably say `install node.js` there instead, cc @kapunahelewong 

node.js comes with npm and is the real hard dependency. we have this readme in a couple spots

steve8708:
plus the npm site doesn't really give help to install npm, but node.js site is more helpful

PR #960, opened 2022-04-05
fixes for starters

steve8708:
- remove an empty div causing the visual editor to appear broken
- always default to `fallback: 'blocking'` so pages don't look broken while loading

### DIFF
--- a/examples/next-js-simple/pages/[[...page]].tsx
+++ b/examples/next-js-simple/pages/[[...page]].tsx
@@ -39,7 +39,7 @@ export async function getStaticPaths() {
 
   return {
     paths: pages.map((page) => `${page.data?.url}`),
-    fallback: true,
+    fallback: 'blocking',
   }
 }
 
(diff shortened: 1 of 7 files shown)


examples/next-js-simple/pages/[[...page]].tsx
```diff
@@ -39,7 +39,7 @@ export async function getStaticPaths() {
 
   return {
     paths: pages.map((page) => `${page.data?.url}`),
-    fallback: true,
+    fallback: 'blocking',
```
Comment (teleaziz) [inline examples/next-js-simple/pages/[[...page]].tsx:42]: Would't this 404 on new urls in draft content? ( prevents editing and previewing draft pages )

Comment (steve8708) [reply examples/next-js-simple/pages/[[...page]].tsx:42]: Hasn’t been an issue that I’ve seen anywhere, the only behavior change is not flashing a janky blank loading page that makes it appear broken

teleaziz (approved):

PR #961, opened 2022-04-05
Chore: SDKs: yarn 3

samijaber:
## Description

- use yarn 3 in SDKs
- fix yarn linking with Mitosis
- add SDKs build to CI (it had stopped running at some point when we migrated to yarn)


packages/sdks/package.json
```diff
@@ -41,5 +40,9 @@
     "commitizen": {
       "path": "git-cz"
     }
+  },
+  "resolutions": {
+    "@X/X: mitosis line below is removed in CI to allow for proper usage of Mitosis.": "0.0.1",
```
mandx, line 45: Does it have to have that syntax? How about something closer to actual comments?
```suggestion
    "// NOTE: mitosis line below is removed in CI to allow for proper usage of Mitosis.": "0.0.1",
```

packages/sdks/scripts/remove-mitosis-resolutions.js
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env node
+
+var fs = require('fs');
+var path = require('path');
+
+const packageJson = require('../package.json');
+delete packageJson.resolutions['@builder.io/mitosis-cli'];
```
mandx, line 7: If this runs in CI only, maybe we could replace this script with a `jq` invocation?

mandx (approved):

packages/sdks/package.json
samijaber: Yeah, it needs to confuse the key with an actual module name. I had your suggestion first and running `yarn` ends up deleting the line 😄.

packages/sdks/scripts/remove-mitosis-resolutions.js
samijaber: Short answer was that I never used `jq` before. But thanks for pushing me to try it! Pretty neat tool to add to my repertoire :) https://github.com/BuilderIO/builder/pull/961/commits/1524eaa264a6a75c9d3363f0128fc6b01412bbf2

PR #977, opened 2022-04-08
Add  example to show how to register action shortcuts in builder

teleaziz:
In this example code, we define an `action shortcut` to trigger events to Google Analytics on specified element events (`submit`, `click`, etc.)

## Getting started
```
git clone git@github.com:BuilderIO/builder.git
cd builder/plugins/example-action-plugin/
npm install
npm run start
```

then configure Builder to load this plugin from localhost:

- Go to to your space settings > plugins
  and add `http://localhost:1268/plugin.system.js` to the list
  Screen Shot 2022-04-08 at 1 42 16 PM

You should be able to see this example in action.

## What are action shortcuts?

Screen Shot 2022-04-08 at 1 34 34 PM

Action shortcuts in Builder are a great way to simplify some of the tasks that would otherwise require developer involvement, using action shortcuts a content creator can trigger tracking events, set up conditional values, or set specific values on state without having to write code, see `src/plugin.tsx` for a detailed example.

Screen Shot 2022-04-08 at 1 29 25 PM


mrkoreye (approved):

ca136:
fyi @tgaribal and @ancheetah this is a good example to show how to register custom actions

PR #988, opened 2022-04-14
Prep: Feature: Svelte SDK

samijaber:
## Description

Prep PR that introduces some required changes:

- fix `getTarget()` so that it works for all frameworks
- extract style rendering logic to a `RenderStyles` component


packages/sdks/src/constants/target.ts
```diff
@@ -0,0 +1,4 @@
+type Target = import('../types/targets').Target;
+
+/** This file should be overriden for each framework. Ideally this would be implemented in Mitosis.  */
+export const TARGET: Target = 'reactNative';
```
mrkoreye, line 4: I am confused by this file, is it manually overwritten? I see `TARGET` used in other files, so I assume it is not always `'reactNative'`

mrkoreye (approved):

packages/sdks/src/constants/target.ts
samijaber: yeah. It's overridden for every single output. I had to pick _something_ for this file even though it's technically never selected. I just didn't want to choose `undefined` or the type would be `Target | undefined` which would become a pain elsewhere.

samijaber: more info: https://github.com/BuilderIO/mitosis/blob/34428c48a9ed4ab3dd5a38c036d156dd61fd421d/docs/project-structure.md#project-structure

> `overrides/` contains a per-target folder that mimicks the structure of src, and will completely swap out any files with identical names. Example: if we have defined `overrides/react-native/src/functions/is-react-native.ts`, it will override `src/functions/is-react-native.ts` in `output/react-native/src/functions/is-react-native.js`

mrkoreye: great, thanks for the explanation!

PR #994, opened 2022-04-25
stop accordion from cutting off contents

steve8708:


mrkoreye (approved):
Lgtm but will this break the animation? Looks like there is a transition on max height

PR #995, opened 2022-04-28
fix inappropriate images showing in demo product results

steve8708:


samijaber (approved):
the lockfile reverted to an old version because of the `node`/`npm` version you are running locally. Not a big issue, but explaining why the huge package-lock json change is in this PR

mandx (approved):

PR #1002, opened 2022-04-29
astro example + ensure solidjs SDK works on server too

steve8708:


samijaber (commented):
One suggestion: Can we move all the solidjs examples under `examples/solidjs`? I started doing that for Vue ( https://github.com/BuilderIO/builder/tree/main/examples/vue ) and I prefer this format since we have so many examples. Categorizing a little bit helps clean up the directory

packages/sdks/package.json
```diff
@@ -54,5 +54,6 @@
       "path": "git-cz"
     }
   },
-  "packageManager": "yarn@3.2.0"
+  "packageManager": "yarn@3.2.0",
+  "resolutions": {}
```
samijaber, line 58: I think you manually removed the 2 lines here, which is not enough. You need to run `yarn run remove-symlinks`: that runs `yarn` at the end which removes the resolutions from `yarn.lock`

steve8708:
Ah, so I saw that and had a concern with it. We have kept things flat for a reason, it’s the same way vercel, firebase, and other projects I’ve seen with this type of directory do this.

I think the best reason to keep it flat is manu examples include many things - eg Astro + solid. Should it go in an Astro folder, or a solid folder, or leave it flat so you can more easily hit cmd+f and search for either thing and find it. 

I lean toward this, as otherwise I fear a lot of these combo examples will get lost

packages/sdks/package.json
steve8708: Oh I try to only run the add/remove symlinks command but perhaps made a mistake

samijaber (approved):

PR #1014, opened 2022-05-11
Use the "Required" attribute

LoganAffleck:
User reported error. Textarea was not preventing empty submission when set as Required in inputs. This was because the value was not being used in the render() method for the component.

### DIFF
--- a/packages/react/src/blocks/forms/TextArea.tsx
+++ b/packages/react/src/blocks/forms/TextArea.tsx
@@ -7,6 +7,7 @@ export interface FormTextAreaProps {
   value?: string;
   defaultValue?: string;
   placeholder?: string;
+  required?: boolean;
 }
 
 class TextAreaComponent extends React.Component<FormTextAreaProps> {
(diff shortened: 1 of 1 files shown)


packages/react/src/blocks/forms/TextArea.tsx
```diff
@@ -17,6 +18,7 @@ class TextAreaComponent extends React.Component<FormTextAreaProps> {
         name={this.props.name}
         value={this.props.value}
         defaultValue={this.props.defaultValue}
+        {if (this.props.required) 'required'}
```
mandx, line 21: Are you sure that syntax works? Don't you mean something like:
```suggestion
        required={!!this.props.required}
```
?

mandx: Another way of doing this
```suggestion
        {...( this.props.required ? { required: 'required' } : {} )}
```

mandx (approved):

PR #1016, opened 2022-05-11
fix UI hydration mismatch error when previewing/editing in react 18

teleaziz:


teleaziz:
reproduction repo: https://github.com/llun/builder-bug
forum post with details: https://forum.builder.io/t/builder-react-sdk-did-not-expect-server-html-to-contain-a-div-in-div/494/19

mrkoreye (approved):

PR #1019, opened 2022-05-13
Fix: registerComponent

samijaber:
## Description

Applies RFC https://github.com/BuilderIO/builder/discussions/1015

- remove all `index.ts` files for components
- store registered components in BuilderContext
- add deprecation warning for `registerComponent`
- add types from Core SDK for `registerComponent` component info


packages/sdks/src/types/element.ts
```diff
@@ -0,0 +1,58 @@
...
+    actions?: {
+      [key: string]: string;
+    };
+  };
+  repeat?: {
+    collection: string;
+    itemName?: string;
+  } | null;
+  animations?: any[]; // TODO: type the animation spec
+}
```
samijaber, line 58: TO-DO: investigate if all of these types carry over or not

packages/sdks/src/types/components.ts
```diff
@@ -0,0 +1,181 @@
...
+  };
+  advanced?: boolean;
+  onChange?: Function | string;
+  code?: boolean;
+  richText?: boolean;
+  showIf?: ((options: Map<string, any>) => boolean) | string;
+  copyOnAdd?: boolean;
+}
+
+export interface ComponentInfo {
```
samijaber, line 68: TO-DO: investigate if all of these types carry over or not

packages/sdks/src/types/components.ts
```diff
@@ -0,0 +1,181 @@
+import { BuilderElement } from './element';
+
+export interface Input {
```
samijaber, line 3: TO-DO: investigate if all of these types carry over or not

PR #1033, opened 2022-05-27
fix examples and docs

steve8708:
1. be sure to use the right hook names
2. always use "preview URL" at the model level, never use "site URL"

@teleaziz the last two places to update are the shopify starter repo and the theme UI example here - can you update those for me next week?


steve8708:
going to merge this in since we're actively running into confused users from the current instructions. but would still value a review to make sure my copy updates are ok next week @mrkoreye, I can send a followup PR to address any suggestions

PR #1056, opened 2022-06-02
more docs improvements

steve8708:


steve8708:
eager merging to share this out for feedback, happy to follow up with updates based on feedback

PR #1067, opened 2022-06-08
Fix: SDK styles

samijaber:
## Description

- create `RenderComponentAndStyles` abstraction
- reuse when `noWrap === false`, fixing lack of styles


samijaber:
@armgjoka The first commit has the changes in Mitosis. The second commit is just building those changes for every SDK individually. You can just check the first commit.

armgjoka (approved):

PR #1069, opened 2022-06-08
feat(cli): add a command to integrate with an existing site 

ca136:
This will be used in the UI to allow users to copy/pase a single line to integrate with their codebase. Initially it will support Next.js, potentially other frameworks down the line.


packages/cli/src/integrate.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+const NEXTJS_PAGES_TEMPLATE_PATH = './src/templates/nextjs-[pages].tsx';
+
+function verifyPagesDirectory() {
+  return fse.existsSync('pages');
+}
+
+function checkForPagesFile(prefix: string, extension: string) {
+  return fse.existsSync(`./pages/${prefix}/[pages].${extension}`);
```
samijaber, line 17: couldn't `[pages]` be any other word though? I think you need to do a glob check for any file in `./pages/${prefix}` with the glob `[(.*)].${extension}` (or something like that)

samijaber: https://nextjs.org/docs/routing/dynamic-routes

packages/cli/src/templates/nextjs-[pages].tsx
```diff
@@ -0,0 +1,62 @@
...
+    },
+  };
+}
+
+export async function getStaticPaths() {
+  // Get a list of all pages in builder
+  const pages = await builder.getAll('<<<MODEL_NAME>>>', {
+    // We only need the URL field
+    fields: 'data.url',
+    options: { noTargeting: true },
```
Comment (samijaber) [inline packages/cli/src/templates/nextjs-[pages].tsx:32]: PS: I believe you need a `limit: 0` here, the default limit is 30

https://github.com/BuilderIO/builder/blob/d7e477991aa5fdc2cee7a5b9ce033dba6fe0e8dc/packages/core/src/builder.class.ts#L2580

packages/cli/src/templates/nextjs-[pages].tsx
```diff
@@ -0,0 +1,62 @@
+import { useRouter } from 'next/router';
```
Comment (samijaber) [inline packages/cli/src/templates/nextjs-[pages].tsx:1]: Potentially, make this a `jsx` file so it works for JS only projects. No types are being declared anywhere here anyways

Comment (ca136) [reply packages/cli/src/templates/nextjs-[pages].tsx:1]: yup, good point, plan is to add both!

ca136:
@samijaber @adamdbradley updated based on your feedback 👍🏻

packages/cli/src/index.ts
```diff
@@ -33,4 +30,16 @@ program
...
   });
 
+program
+  .command('integrate')
+  .description('integrate Builder.io with an existing codebase, currently supports Next.js')
+  .option('-d,--debug', 'print debugging information')
+  .option('-s,--stack <stack>', 'currently supports nextjs', 'nextjs')
+  .option('-m,--model <model>', 'name of the model you want to integrate')
+  .option('-a,--apiKey <apiKey>', 'you can find your apiKey on builder.io/account/settings')
+  .option('-p,--pathPrefix <prefix>', 'ADD SOMETHING GOOD HERE', '/m/')
```
samijaber, line 40: I think you're missing the `language` flag you just added!

I would make sure that it's optional btw, and default to JS instead of requiring a value.

samijaber: Looks from the code that you did make it optional, so no worries there.

samijaber (approved):
Love this! 🎉 🚀

ca136:
And this is the success/next steps messaging:
image

packages/cli/src/integrate.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    3. Once you've deployed to a remote url, you'll want to edit the ${chalk.cyan(
+      'previewUrl'
+    )} property of your model here:
+       ${chalk.cyan(`https://builder.io/models/${options.model}`)}
+
+    --------------------------------------------------
+    --------------------------------------------------
+
+    ${chalk.green(`We'd love to here from you! Please give us feedback on any
```
samijaber, line 159: the sentence is slightly incomplete here

PR #1081, opened 2022-06-18
feat: 🎸 Initial Qwik SDK

mhevery:
Initial implementation of Qwik SDK generated by mitosis

## Description

_Screenshot_


packages/sdks/output/qwik/src/blocks/button/button.tsx
```diff
@@ -0,0 +1,202 @@
+// GENERATED BY MITOSIS
+
+import { component$, useStore, h, Fragment } from "@builder.io/qwik";
+export const COMPONENT = {
```
steve8708, line 4: What are these for? Usually this should compile away

packages/sdks/output/qwik/src/blocks/columns/columns.tsx
```diff
@@ -0,0 +1,258 @@
...
+  return typeof props.space === "number" ? props.space || 0 : 20;
+}
+function getColumns(props, state) {
+  return props.columns || [];
+}
+function getWidth(props, state, index: number) {
+  const columns = getColumns(props, state);
+  return columns[index]?.width || 100 / columns.length;
+}
+function getColumnCssWidth(props, state, index: number) {
```
steve8708, line 180: Do we need to extract these functions to the top of the file instead of having them in the function body?

steve8708: Aka the component function body. Just bc mitosis tries to generate idiomatic code and this feels less idiomatic compared to normal react code that would just be in there and have normal function signatures (vs this idea of always having the props and state args)

this is obv a mitosis related comment but wanted to put in ckntext

packages/sdks/output/qwik/src/blocks/button/button.tsx
mhevery: I keep them here for debugging purposes. They will be removed.

packages/sdks/output/qwik/src/blocks/columns/columns.tsx
mhevery: The issue here is that in all other frameworks there is a way to attach behavior (methods) to data. In Qwik functions can't be attached to data because they can't be serialized. This creates a problem, which I don't know how to solve. So for now I turn them into top lever functions. Because these functions capture props and state I need to pass these values directly.

@manucorporat do you have an idea on how to solve this?

Related issue: https://github.com/BuilderIO/qwik/issues/617#issuecomment-1157386259

steve8708: Now if Qwik can't take functions in the component body, then I think that is probably worth solving. But my understanding is the optimizer + useLexicalScope solves this

steve8708: also this is not well documented (my fault), but the idea of Mitosis outputs is they should be as idiomatic as possible (aka if someone "ejects" and decides to use the Qwik code directly, it should be as close to what people would have hand written as we can ideally get)

packages/sdks/output/qwik/src/blocks/submit-button/submit-button.tsx
```diff
@@ -0,0 +1,89 @@
...
+    "export interface ButtonProps {\n  attributes?: any;\n  text?: string;\n}",
+  ],
+  propsTypeRef: "ButtonProps",
+};
+export interface ButtonProps {
+  attributes?: any;
+  text?: string;
+}
+const SubmitButton = component$((props: ButtonProps) => {
+  const state = useStore({});
```
steve8708, line 82: nit pick: we don't need to generate state if no state is used. iirc in mitosis there is a helper function to detect if there actually is any state used (you can peek at how the other generators detect this)

packages/sdks/output/qwik/src/blocks/section/section.tsx
```diff
@@ -0,0 +1,101 @@
...
+export interface SectionProps {
+  maxWidth?: number;
+  attributes?: any;
+  children?: any;
+  builderBlock?: any;
+}
+const SectionComponent = component$((props: SectionProps) => {
+  const state = useStore({});
+  return (
+    <section
```
steve8708, line 87: realized we're not handling CSS currently from what I can tell. we prob would have the qwik generator generate whatever is idiomatic, IIRC that's a CSS file plus a `useScopedStyles$()` hook here?

packages/sdks/output/qwik/src/blocks/columns/columns.tsx
mhevery: Agree on "ejection"

So your suggestion will work for JSX because the function gets re-executed and hence the helper methods get re-created. But those methods will not be executable in say `useWatch$()` location or in `useMount$()` (Basically any place where they cross the `$` boundary. 

So from my point of view, we don't yet have an idiomatic way of doing that which can be used from all locations in Qwik, and that is what we need to solve. For now, it is the hack above, but I am not happy with the hack.

steve8708: Ah yes I see, functions in component body not currently supported, this works for me in the meantime 👍

packages/sdks/output/qwik/src/blocks/section/section.tsx
mhevery: this has been added now.

packages/sdks/src/components/render-block/render-block.lite.tsx
```diff
@@ -221,12 +225,20 @@ export default function RenderBlock(props: RenderBlockProps) {
            */}
           <For each={state.childrenWithoutParentComponent}>
             {(child) => (
-              <RenderBlock key={'render-block-' + child.id} block={child} />
+              <RenderBlock
+                key={'render-block-' + child.id}
+                block={child}
+                context={props.context}
```
steve8708, line 231: how important is this change (all the changes in here changing builder context to be a prob and not normal context)? I realized I was wrong and RenderBlock does get used by customers, and really should use normal context if possible

the main impact on users is they can define child blocks in their code, aka

```tsx
function myCustomComponent(props) {
  return <div>{props.someArray.map(block => <RenderBlock block={block} />)</div>
}
```

so we need to try to keep this the same

mhevery: 1.) I needed to convert this to lite-component so that we don't have extra wrapper divs.
2.) Now that we have hostless we could convert this back, but we would have a lot of extra comments, which is not ideal.
3.) I think we should rename `RenderBlock` to `RenderBlockInternal`, which just calles `RenderBlock` but which can be used by our customers. This way we can have same API and not have extra junk in HTML.

mhevery: OK, looked into it deeper and I don't think this is an issue. 
- `<RenderBlocks>` Is exported and public. Its props interface did not change.
- `<RenderBlock>` is never exported (internal use only) so we can change its props without implication. So I am going to leave this as is.

samijaber (approved):
Welcome to the SDKs 😄

PR #1092, opened 2022-06-27
Feature: initial vue3 SDK

samijaber:
## Description

- add vue-3 SDK output
- update SDKs to use latest mitosis (some breaking changes there)
- fix JSX types imports from using React to using Mitosis JSX types


examples/vue/vue-3/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "dev": "vite",
+    "build": "run-p type-check build-only",
+    "preview": "vite preview --port 4173",
+    "build-only": "vite build",
+    "type-check": "vue-tsc --noEmit",
+    "lint": "eslint . --ext .vue,.js,.jsx,.cjs,.mjs,.ts,.tsx,.cts,.mts --fix --ignore-path .gitignore"
+  },
+  "dependencies": {
+    "@builder.io/sdk-vue": "^0.0.1-55",
+    "vue": "^3.2.37",
```
kylecordes, line 14: Should possibly be a `peerDependency` (and likely also a `devDependency` for the bundle/build to work). I've not worked on this exact situation for Vue, but in the general case for other similar frameworks the goal is to avoid a library like this bringing in it own copy (or separate version) of the base library, rather to share with the consuming project.

samijaber: Good point! Will do that in a follow up PR

packages/sdks/mitosis.config.js
```diff
@@ -9,24 +9,40 @@ const getSeededId = () => {
...
+    vue2: vueConfig,
+    vue3: vueConfig,
+    reactNative: {
+      plugins: [
+        () => ({
+          code: {
+            pre: (code) => {
+              // workaround until we resolve
+              // https://github.com/BuilderIO/mitosis/issues/526
+              code.replace('srcset=', 'srcSet=');
```
mrkoreye, line 38: hacker status

mrkoreye (approved):

packages/sdks/mitosis.config.js
samijaber: Those Mitosis maintainers take too long to respond to me and fix their bugs 😠

PR #1104, opened 2022-07-01
Android sdk v0

steve8708:
Screen Shot 2022-07-01 at 1 23 41 PM


packages/android/sdk/app/src/main/java/com/example/myapplication/blocks/BuilderText.kt
```diff
@@ -0,0 +1,46 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.text.style.TextAlign
+import com.example.myapplication.*
+import kotlinx.serialization.json.contentOrNull
+import kotlinx.serialization.json.jsonPrimitive
+
+@Composable
+fun BuilderText(block: BuilderBlock, text: String) {
+    val stripped = text.replace(
+        "<.*?>".toRegex(),
```
mrkoreye, line 16: what is this doing?

packages/android/sdk/app/src/main/java/com/example/myapplication/ui/theme/Theme.kt
```diff
@@ -0,0 +1,47 @@
...
+    surface = Color.White,
+    onPrimary = Color.White,
+    onSecondary = Color.Black,
+    onBackground = Color.Black,
+    onSurface = Color.Black,
+    */
+)
+
+@Composable
+fun MyApplicationTheme(
```
mrkoreye, line 31: Why do we have this theme file if this is the SDK?

packages/swift/src/testing.xcodeproj/xcuserdata/stephensewell.xcuserdatad/xcschemes/xcschememanagement.plist
```diff
@@ -28,12 +28,12 @@
 		<key>testing (iOS).xcscheme_^#shared#^_</key>
 		<dict>
 			<key>orderHint</key>
-			<integer>36</integer>
+			<integer>34</integer>
```
mrkoreye, line 31: Why the changes in this file? It looks like maybe a local file that should not be committed?

mrkoreye (approved):
Why do we need all the `ic_launcher` images and xml files? Seems like there are a lot of files in here from like a project generator or something that maybe don't apply for an SDK?

packages/android/sdk/app/src/androidTest/java/com/example/myapplication/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,24 @@
+package com.example.myapplication
```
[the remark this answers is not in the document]
steve8708: ah yes this was generated with the project, the idea is to not worry about structure yet, a next step is to rework this into a proper package which will have a name we decide (e.g. maybe like `io.builder.androidSdk`), and shuffle the files, project name, etc accordingly

packages/android/sdk/app/src/main/java/com/example/myapplication/MainActivity.kt
```diff
@@ -0,0 +1,106 @@
+package com.example.myapplication
```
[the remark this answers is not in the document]
steve8708: same reason as above

packages/android/sdk/app/src/main/java/com/example/myapplication/blocks/BuilderText.kt
steve8708: strips HTML, will add comment

packages/swift/src/testing.xcodeproj/xcuserdata/stephensewell.xcuserdatad/xcschemes/xcschememanagement.plist
steve8708: no clue, I think these should probably be gitignored, will look into

packages/gatsby/src/batching/merge-queries.js
```diff
@@ -1,345 +0,0 @@
-const { visit, visitInParallel, Kind } = require(`gatsby/graphql`);
```
mrkoreye, line 1: Are these gatsby file removals supposed to be in here? Why are we removing them?

steve8708: I wiped out all gitignore files. I assume they are here by mistake but @teleaziz maybe let us know if that's a wrong assumption

teleaziz: those files looks needed, the gitignore is for batching folder at the root of the package which is being copied on build and publish https://github.com/BuilderIO/builder/blob/f247fa99562f5245f82cbf9dcea1143f4ebf67d3/packages/gatsby/src/constants.js#L5

PR #1105, opened 2022-07-05
render default variant when cookies are disabled

teleaziz:


packages/react/src/functions/try-eval.tsx
```diff
@@ -61,7 +61,7 @@ export const tryEval = (str?: string, data: any = {}, errors?: Error[]): any =>
       }).run(value.replace(/(^|;)return /, '$1'));
       // tslint:enable:comment-format
     }
-  } catch (error: any) {
+  } catch (error) {
```
teleaziz, line 64: merge error maybe, didn't change this file, looking 👀

PR #1124, opened 2022-07-13
native SDK fixes

steve8708:


steve8708:
Need to get in to unblock a demo/deal, happy to follow up to address feedback

PR #1125, opened 2022-07-13
[builder.io/react] move react/react-dom to peerdeps and fix npm insta…

teleaziz:
…ll errors

### DIFF
--- a/examples/next-js-simple/package.json
+++ b/examples/next-js-simple/package.json
@@ -44,8 +44,8 @@
     ]
   },
   "dependencies": {
-    "@builder.io/react": "^2.0.2-7",
-    "@builder.io/widgets": "^1.2.22-7",
+    "@builder.io/react": "^2.0.2",
+    "@builder.io/widgets": "^1.2.23",
     "next": "^12.2.2",
     "next-seo": "^5.4.0",
     "react": "^18.2.0",
--- a/examples/next-js-simple/package.json
+++ b/examples/next-js-simple/package.json
@@ -44,7 +44,7 @@
     ]
   },
   "dependencies": {
-    "@builder.io/react": "^2.0.2-6",
+    "@builder.io/react": "^2.0.2-7",
     "@builder.io/widgets": "^1.2.22-7",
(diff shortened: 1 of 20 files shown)


examples/next-js-simple/package.json
```diff
@@ -44,16 +44,12 @@
...
-    "@vercel/fetch": "^6.1.0",
-    "atob": "^2.1.2",
-    "jest": "^26.6.3",
-    "next": "^12.1.0",
-    "next-seo": "^4.11.0",
-    "react": "^16.14.0",
-    "react-dom": "^16.14.0",
-    "react-merge-refs": "^1.1.0"
+    "@builder.io/react": "^2.0.2-7",
+    "@builder.io/widgets": "^1.2.22-7",
```
teleaziz, line 48: I'll release patch once this pr is reviewed and update the dependencies in this example

packages/widgets/package.json
```diff
@@ -127,17 +127,13 @@
...
-    "inferno": "^7.1.9",
-    "inferno-compat": "^7.1.9",
     "preact": "^8.4.2",
     "preact-compat": "^3.18.4",
     "preact-context": "^1.1.3",
     "prop-types": "^15.7.2"
   },
   "peerDependencies": {
-    "@builder.io/react": ">=1.1.0"
+    "@builder.io/react": ">=1.1.0 ^2.0.2-5"
```
samijaber, line 136: isn't this syntax broken? It doesn't make sense to me

if you go to this calculator: https://semver.npmjs.com/

and put the values in, you see only ^2.0.2-5 as highlighted 
image

```suggestion
    "@builder.io/react": "^2.0.2-5"
```

samijaber (approved):

packages/widgets/package.json
teleaziz: It's valid, but it's only temporary, the original one `>=1.1.0` should be unchanged, this was needed because `>=` doesn't include dev versions, I'll remove it once a patch version of the react sdk is released

PR #1129, opened 2022-07-14
Fix: SDKs image block

samijaber:
## Description

- add srcset support
- add webp support


packages/sdks/src/blocks/image/image.helpers.ts
```diff
@@ -0,0 +1,75 @@
+// Taken from (and modified) the shopify theme script repo
```
samijaber, line 1: all this code is copied from the React SDK verbatim

samijaber: https://github.com/BuilderIO/builder/blob/16cbe4646ffd639d2d255affb410e7642cc5a49d/packages/react/src/blocks/Image.tsx#L61-L94

mrkoreye (approved):

PR #1168, opened 2022-07-25
improve withBuilder HOC types

charkour:
## Description
Hey Builder Team,

This PR adds improved TS definitions to the `withBuilder` HOC. Before the change, the function returned from `withBuilder` had the type `Function`. Now, the function returned matched the type of the passed component. If the component is `FC<Props>`, the returned function is `React.ComponentType<Props>`

Thanks!

### DIFF
--- a/packages/react/src/decorators/builder-block.decorator.ts
+++ b/packages/react/src/decorators/builder-block.decorator.ts
@@ -1,9 +1,10 @@
-import { Builder, Component } from '@builder.io/sdk';
+import { Builder, Class, Component } from '@builder.io/sdk';
+import React from 'react';
 
 interface ReactComponent extends Component {}
 
-export function BuilderBlock(options: ReactComponent) {
+export function BuilderBlock<P extends object>(options: ReactComponent) {
   options.type = 'react';
 
-  return Builder.Component(options);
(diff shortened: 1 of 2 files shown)


teleaziz:
thanks @charkour, appreciate this PR from you, we're actually moving towards deprecating this method in the next version (soft deprecation at the beginning but we will drop it entirely in our next major version), if you're using it please use `Builder.registerComponent` instead, for examples on registering custom component recommend checking the [react-design-system example](https://github.com/BuilderIO/builder/blob/main/examples/react-design-system)

charkour:
Hey @teleaziz, 

Thanks for getting back to me and for letting me know the future of the Builder SDK packages.

Do you, by chance, happen to have a rough timeline for when the next version will be released? Also, will the next version have better TS support and the ability to use different CSS-in-JS libs? Another big concern for my team is the Builder bundle size. Currently, the "lite" version is 35 kB and we would like it to be smaller if at all possible. I'd be willing to work with your engineers to make that happen. 

Thank you! And have a great week.

charkour:
For what it's worth, the [Adding Children to Custom Components](https://www.builder.io/c/docs/custom-components-children) documentation makes use of `withChildren` and there is no deprecation notice on the docs or within the code that I found.

What are the plans to deprecate? 

Thank you.

PR #1186, opened 2022-08-08
reduce widgets bundle size by 85%

steve8708:
from 161kb (before minification and gzipping) to 23.8kb by making all component async imported, so they are only included when used


packages/widgets/src/builder-widgets-sync.ts
```diff
@@ -0,0 +1,15 @@
...
+import { MasonryComponent } from './components/Masonry';
+import { carouselConfig } from './components/Carousel.config';
+import { tabsConfig } from './components/Tabs.config';
+import { accordionConfig } from './components/Accordion.config';
+import { masonryConfig } from './components/Masonry.config';
+
+Builder.registerComponent(CarouselComponent, carouselConfig);
+Builder.registerComponent(TabsComponent, tabsConfig);
+Builder.registerComponent(AccordionComponent, accordionConfig);
+Builder.registerComponent(MasonryComponent, masonryConfig);
```
samijaber, line 15: Doesn't seem like this new file is imported/used anywhere, is it?

samijaber (approved):

packages/widgets/src/builder-widgets-sync.ts
steve8708: it's meant to be directly imported from dist if desired, in case anyone still needs the old behavior we can point them to importing that directly (it's copy + paste from the old main path, and may still be needed by some if their bundler doesn't support dynamic imports)

mrkoreye (approved):

samijaber:
@steve8708 can we merge this or close it (since we're adding the widgets to v2 and they will be async loaded there eventually)?

PR #1188, opened 2022-08-09
use actual buttons for button blocks

steve8708:
this is best for a11y and semantic html, and with `all: unset` we can avoid styling issues from browser default styling

next thing to figure out is how to preserve the outline behavior on hover, ideally default styling would work for that. we could perhaps in the short term just add like `css={{ ':hover': { outline: '1px solid some-color' } }}`

@samijaber does mitosis support :hover in css={{}} currently?

### DIFF
--- a/packages/sdks/src/blocks/button/button.lite.tsx
+++ b/packages/sdks/src/blocks/button/button.lite.tsx
@@ -12,7 +12,7 @@ export default function Button(props: ButtonProps) {
     <>
       <Show
         when={props.link}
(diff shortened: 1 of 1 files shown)


samijaber:
> @samijaber does mitosis support :hover in css={{}} currently?

I can't say for sure. I think it wouldn't work in some places like SolidJS...probably works in React/Vue/Svelte which can handle normal CSS? Can add a test in Mitosis to find out for sure

samijaber (approved):

PR #1190, opened 2022-08-10
[plugins/example-seo-review] add an example plugin for sending html c…

teleaziz:
…ontent for SEO review

This is a demo and a skeleton of a plugin that extracts  html content of iframed preview areas to be sent to seo review services like Yoast


gustavocodess (approved):
I would remove those console.logs on plugin.tsx file. Other than that looks good.

teleaziz:
thanks Gustavo, console.logs are meant to be there for education purposes

PR #1208, opened 2022-08-16
update package-lock for compatibility with newer npm

gustavocodess:
## Description

update package-lock for compatibility with newer npm

_Screenshot_


teleaziz:
please update to use JS in this example

PR #1211, opened 2022-08-16
fix lazy loading in lite package

charkour:
## Description
This correctly fixes the issue in lite package where loading was always set to lazy.

Thank you. Please let me know if you need steps to reproduce or any other changes.


charkour:
I know your team is hard at work at v2, so I apologize for the extra ping. This seems like an important fix considering the lite version is used by clients in prod.

Thanks

charkour:
My team is no longer using Builder and I'd like to cleanup my GitHub PR tab. 

Hopefully, this will be fixed in the future! Thanks.

PR #1212, opened 2022-08-16
update sdks and add skip preflight

gustavocodess:
## Description

Added new builder sdks and added .env file with skip preflight to run with newer versions of node


teleaziz (approved):

gustavocodess:
Merge

PR #1214, opened 2022-08-17
test: 💍 Add sdk-qwik e2e tests

mhevery:
## Description

_Screenshot_


packages/sdks/e2e/qwik/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@builder.io/e2e-qwik",
+  "description": "Blank Qwik starter app. Vite.js tooling. Express.js server. Prettier code formatter.",
+  "private": true,
+  "type-disable-for-playwright": "module",
```
samijaber, line 5: Can we remove this then?

```suggestion
```

samijaber (approved):
:ok

packages/sdks/e2e/qwik/package.json
mhevery: I would like to keep this. My thinking is that in the future, someone may want to add it back and then will get confused as to what breaks. This way it is a form of commenting which tells the dev that this is intentionally disabled, and if you enable it you should expect to verify that playwright will break.

PR #1216, opened 2022-08-18
feat(utils): add translation to custom components

sanyamkamat:
## Description

This PR adds translation to custom component blocks.

### DIFF
--- a/packages/utils/src/translation-helpers.test.ts
+++ b/packages/utils/src/translation-helpers.test.ts
@@ -25,16 +25,16 @@ test('getTranslateableFields from content to match snapshot', async () => {
         },
         {
           '@type': '@builder.io/sdk:Element',
-          'id': 'builder-15cd8202b8a3423d9fce69d71ee63510',
-          'component': {
-            'name': 'Heading',
-            'options': {
-              'title': {
+          id: 'builder-15cd8202b8a3423d9fce69d71ee63510',
+          component: {
+            name: 'Heading',
+            options: {
+              heading: {
                 '@type': '@builder.io/core:LocalizedValue',
-                'Default': 'I am a heading!'
-              }
-            }
-          }
+                Default: 'I am a heading!',
+              },
+            },
+          },
         },
       ],
     },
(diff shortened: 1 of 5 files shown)


packages/utils/src/translation-helpers.test.ts
```diff
@@ -53,13 +66,27 @@ test('applyTranslation from content to match snapshot', async () => {
...
+          }
+        },
       ],
     },
   };
 
   const translations = {
     'metadata.title': { value: 'salut' },
     'blocks.block-id#text': { value: 'block-id' },
+    'blocks.1#component#options#title': { value: 'title translated' },
```
teleaziz, line 89: The format for this should be `blocks.${block id}${path.join('#')}`

packages/utils/src/translation-helpers.ts
```diff
@@ -35,7 +36,12 @@ export function getTranslateableFields(
 
   // blocks
   traverse(blocks).forEach(function (el) {
-    // TODO: localized custom components inputs
+    if (this.key && el && el['@type'] === localizedType) {
```
teleaziz, line 39: localized types can reference any time, not just `text`, we need to exclude that, maybe set a meta property on the element to flag the text inputs

packages/utils/src/translation-helpers.ts
```diff
@@ -68,7 +74,14 @@ export function applyTranslation(
   });
 
   traverse(blocks).forEach(function (el) {
-    // TODO: localized custom components inputs
+    const path = this.path?.join('#');
+    if (translation[`blocks.${path}`]) {
+      this.update({
+        ...el,
+        [locale]: translation[`blocks.${path}`].value,
+      });
```
teleaziz, line 82: This feels incorrect,  we should be only updating the bindings to point to the state.translations property, same as the next few lines

teleaziz (changes requested):

packages/utils/src/translation-helpers.ts
```diff
@@ -67,8 +79,21 @@ export function applyTranslation(
...
+    const blockPath = [...this.path];
+    const blockIndex = blockPath.shift();
+    if (blockIndex) {
+      const parsedBlockIndex = parseInt(blockIndex);
+      const blockId = blocks![parsedBlockIndex].id;
+      const path = blockPath.join('#');
+      if (translation[`blocks.${blockId}#${path}`]) {
+        this.update({
+          ...el,
+          [locale]: translation[`blocks.${blockId}#${path}`].value,
```
teleaziz, line 92: This should update the bindings not the el object

teleaziz: check this line for reference https://github.com/BuilderIO/builder/blob/main/packages/utils/src/translation-helpers.ts#L87

teleaziz:
not sure what happened, but we should've merged this one after release

teleaziz (approved):

PR #1230, opened 2022-08-29
React quill sample

gustavocodess:
## Description

Quick simple example of a nextjs app with builder integrate and React Quill as a Custom Rich text component input.

_Screenshot_


gustavocodess:
Private code to specific customer, closing PR.

PR #1238, opened 2022-09-02
Add get all content

gustavocodess:
## Description

This is a helper function to fetch content with nested refs such as direct model reference or reference list. 

Example: 
ModelC {
name: ''
}
Model B {
themeName: reference to model C
}

ModelA {
colorsList: [Reference to model B, Reference to model B]
}

This function when called should get all values for all nested refs. In the example above should get the values for the list of references to ModelB and and the actual data for ModelC

Added a new file to handle the entire function.


packages/utils/src/get-all-content-with-refs.ts
```diff
@@ -0,0 +1,37 @@
...
+            .catch(err => {
+              console.error(`Error fetching relationship ${child.id} (API key: ${apiKey}):`, err);
+            })
+        );
+      }
+    }
+  });
+
+  await Promise.all(promises);
+  return { content: result };
```
teleaziz, line 28: this should be return result to match the value assigned to `child.value`

packages/utils/src/get-all-content-with-refs.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export async function getContentWithAllReferences(builderInstance: Builder, modelName: string) {
+  const content: { results: BuilderContent[] } = await fetch(
+    `https://cdn.builder.io/api/v2/content/${modelName}?apiKey=${builderInstance.apiKey}`
```
teleaziz, line 33: please use builderInstance to fetch the content, and accept the same arguments as builder.get

teleaziz (changes requested):
few commnets, please add a test for this method

gustavocodess:
> few commnets, please add a test for this method

Done.

packages/utils/src/get-all-content-with-refs.ts
```diff
@@ -0,0 +1,32 @@
...
+  return Boolean((item as any)?.['@type'] === '@builder.io/sdk:Element');
+};
+
+async function resolveRefs(content: any, apiKey: string, builderInstance: Builder) {
+  const promises: Promise<void>[] = [];
+  const result = traverse(content).map(function (child) {
+    if (child && child['@type'] === '@builder.io/core:Reference') {
+      if (child.model) {
+        promises.push(
+          builderInstance.getAll(child.model, { query: { id: child.id } }).then(async value => {
```
teleaziz, line 14: this should be `builderInstance.get(child.model, { options: { query: {id : child.id} }`, ref.value should be a content object

packages/utils/src/get-all-content-with-refs.ts
```diff
@@ -0,0 +1,32 @@
...
+    }
+  });
+
+  await Promise.all(promises);
+  return result;
+}
+
+export async function getContentWithAllReferences(builderInstance: Builder, modelName: string) {
+  const content: BuilderContent[] = await builderInstance.getAll(modelName, {
+    options: { noTargeting: true },
```
teleaziz, line 28: wdyt of making this accept the same options as builder.get, this method could be more helpful for the user if they'd replace:
```
const content = await builder.get('model', options).promise();
```
with 
```
const contentWithAllRefs = await getContentWithAllRefs(builder, 'mode', options)
```

gustavocodess:
> 

done.

packages/utils/src/get-all-content-with-refs.ts
```diff
@@ -0,0 +1,36 @@
...
+  await Promise.all(promises);
+  return result;
+}
+
+export async function getContentWithAllReferences(
+  builderInstance: Builder,
+  modelName: string,
+  options: GetContentOptions
+) {
+  const content: BuilderContent[] = await builderInstance.getAll(modelName, options);
```
teleaziz, line 33: this should be `builder.get`

packages/utils/src/get-all-content-with-refs.ts
```diff
@@ -0,0 +1,36 @@
...
+};
+
+async function resolveRefs(content: any, builderInstance: Builder, options?: GetContentOptions) {
+  const promises: Promise<void>[] = [];
+  const result = traverse(content).map(function (child) {
+    if (child && child['@type'] === '@builder.io/core:Reference') {
+      if (child.model) {
+        promises.push(
+          builderInstance
+            .getAll(child.model, { options: options || { query: { id: child.id } } })
```
teleaziz, line 15: this should be `builder.get`

packages/utils/src/get-all-content-with-refs.test.ts
```diff
@@ -0,0 +1,123 @@
+import { builder } from '@builder.io/sdk';
+import { getContentWithAllReferences } from './get-all-content-with-refs';
+
+test('Adds async props', async () => {
```
teleaziz, line 4: test name should reference the method

packages/utils/src/get-all-content-with-refs.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  const content = [
+    {
+      createdBy: 'IqzTkBgwWTbzaXDAhTL6IpwiCah1',
+      createdDate: 1661750062934,
+      data: {
+        colors: {
+          '@type': '@builder.io/core:Reference',
+          id: '740c91c542b047e3965e2c0985cea3f4',
+          model: 'color-collection',
+          value: [
```
teleaziz, line 14: reference values are content objects and not arrays

packages/utils/src/get-all-content-with-refs.ts
```diff
@@ -0,0 +1,36 @@
...
+  return Boolean((item as any)?.['@type'] === '@builder.io/sdk:Element');
+};
+
+async function resolveRefs(content: any, builderInstance: Builder, options?: GetContentOptions) {
+  const promises: Promise<void>[] = [];
+  const result = traverse(content).map(function (child) {
+    if (child && child['@type'] === '@builder.io/core:Reference') {
+      if (child.model) {
+        promises.push(
+          builderInstance
```
teleaziz, line 14:
```
promises.push( getContentWithAreferences(builderInsatnce, child.model, { query: {id: child.id} }).then(value => child.value = value)```

File: packages/utils/src/get-all-content-with-refs.ts [hunk 10]
```diff
@@ -0,0 +1,34 @@
...
+
+async function resolveRefs(content: any, builderInstance: Builder, options?: GetContentOptions) {
+  const promises: Promise<void>[] = [];
+  const result = traverse(content).map(function (child) {
+    if (child && child['@type'] === '@builder.io/core:Reference') {
+      if (child.model) {
+        getContentWithAllReferences(
+          builderInstance,
+          child.model,
+          options || { query: { id: child.id } }
```
teleaziz, line 16:
```
{ query: { id: child.id } }
```

packages/utils/src/get-all-content-with-refs.ts
```diff
@@ -0,0 +1,34 @@
...
+  await Promise.all(promises);
+  return result;
+}
+
+export async function getContentWithAllReferences(
+  builderInstance: Builder,
+  modelName: string,
+  options: GetContentOptions
+) {
+  const content: BuilderContent = await builderInstance.get(modelName, options);
```
teleaziz, line 31:
```
builderInstance.get(modelName, options).promise()
```

packages/utils/src/get-all-content-with-refs.ts
```diff
@@ -0,0 +1,34 @@
+import { BuilderContent, BuilderElement, Builder, GetContentOptions } from '@builder.io/sdk';
+import traverse from 'traverse';
+
+export const isBuilderElement = (item: unknown): item is BuilderElement => {
```
teleaziz, line 4: not used

gustavocodess:
The test won't work because it will always return the same prepared content on the json, so it enters a loop, how should we proceed? @teleaziz

packages/utils/src/get-all-content-with-refs.ts
```diff
@@ -0,0 +1,31 @@
+import { BuilderContent, Builder, GetContentOptions } from '@builder.io/sdk';
+import traverse from 'traverse';
+
+async function resolveRefs(content: any, builderInstance: Builder) {
+  const promises: Promise<void>[] = [];
+  const result = traverse(content).map(function (child) {
+    if (child && child['@type'] === '@builder.io/core:Reference') {
+      if (child.model) {
```
ersinakinci, line 8: Should this also maybe check if `child.model && !child.value` to avoid unnecessary fetches if the ref has already been resolved (e.g., from the content API)?

samijaber:
Closing due to being stale. Feel free to reopen if there is still interest in adding this helper.

PR #1240, opened 2022-09-05
ENG-2956 Honor custom breakpoints, if available, when rendering content

murtaza-builder:
## Description
When custom break-points are available in content data, re-initialize size constants, from builder-component 's constructor with supplied values. This is so that the corresponding rendering flow can proceed w/o any further changes.

+ A content json can have custom breakpoints specified at `content.data.breakpoints`
    content.data.breakpoints = {small: <number>, medium: <number>}

[React SDK: Honor custom breakpoints when rendering content](https://builder-io.atlassian.net/browse/ENG-2956)

Note: App level changes to save breakpoint data in content is yet to be done.


packages/react/src/constants/device-sizes.constant.ts
```diff
@@ -36,3 +36,21 @@ export const sizes = {
     return 'large';
   },
 };
+
+export const useSizes = ({ small, medium }: { small?: number; medium?: number }) => {
```
mrkoreye, line 40: I think a name like `setBreakpoints` is a little more descriptive of what is happening
```suggestion
export const setBreakpoints = ({ small, medium }: { small?: number; medium?: number }) => {
```

murtaza-builder: just removed this one and created another one for diff purpose

packages/react/src/constants/device-sizes.constant.ts
```diff
@@ -36,3 +38,39 @@ export const sizes = {
...
+    max: medium,
+    min: newSizes.small.max + 1,
+    default: newSizes.medium.min + 1,
+  };
+
+  newSizes.large = {
+    max: 2000, // TODO: decide upper limit
+    min: newSizes.medium.max + 1,
+    default: newSizes.large.min + 1,
+  };
```
samijaber, line 73: the way you calculate the `default` values is giving the old values. You want to read `newSizes.small.min`, but you want the new value that still has not been set. You are instead reading the values that exist from cloning `sizes`.

See this example:

image

As you can see, `bar` uses the old `foo` value, not the new one being set in the same expression.

You're going to want to first set the `max` values, then the `min` values, and then set the `default` values afterwards to make sure to read them correctly. Or use the `small`/`medium` variables instead:

```ts
// this doesn't work
  newSizes.large = {
    max: 2000, // TODO: decide upper limit
    min: newSizes.medium.max + 1,
    default: newSizes.large.min + 1,
  };

// this works
  newSizes.large = {
    max: 2000, // TODO: decide upper limit
    min: newSizes.medium.max + 1,
  };
newSizes.large.default = newSizes.large.min + 1;

// this also works
  newSizes.large = {
    max: 2000, // TODO: decide upper limit
    min: newSizes.medium.max + 1,
    default: newSizes.medium.max + 1 + 1,
  };
```

I recommend the middle approach to avoid duplicating the calculations

samijaber (changes requested):
I think I found an issue in how `default` values are computed. Looks good otherwise

packages/react/src/constants/device-sizes.constant.ts
murtaza-builder: That's a great catch. Thanks.

packages/react/src/constants/device-sizes.constant.ts
```diff
@@ -36,3 +38,39 @@ export const sizes = {
     return 'large';
   },
 };
+
+interface Breakpoints {
+  small: number;
+  medium: number;
```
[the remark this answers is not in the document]
murtaza-builder: done

packages/react/src/components/builder-component.component.tsx
```diff
@@ -361,6 +361,7 @@ export class BuilderComponent extends React.Component<
   private _asyncRequests?: RequestOrPromise[];
   private _errors?: Error[];
   private _logs?: string[];
+  private sizes: any;
```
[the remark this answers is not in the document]
murtaza-builder: done

packages/react/src/constants/device-sizes.constant.ts
```diff
@@ -36,3 +38,39 @@ export const sizes = {
...
+}
+
+export const getSizesForBreakpoints = ({ small, medium }: Breakpoints) => {
```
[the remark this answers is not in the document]
murtaza-builder: couldn't do this since I had to use typeof sizes to address comment in 
image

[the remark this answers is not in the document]
samijaber: You can do so by only exporting the type:

```ts
const sizes = //...

export const Sizes = typeof sizes
```

packages/react/src/constants/device-sizes.constant.ts
```diff
@@ -36,3 +38,42 @@ export const sizes = {
...
+    max: medium,
+    min: newSizes.small.max + 1,
+    default: 0, // Note: For TS
+  };
+  newSizes.medium.default = newSizes.medium.min + 1; // TODO: handle negative values?
+
+  newSizes.large = {
+    max: 2000, // TODO: decide upper limit
+    min: newSizes.medium.max + 1,
+    default: 0, // Note: For TS
```
samijaber, line 74: so we don't need to use the `default` values at all?

samijaber: I see that `default` is not being used in the SDK anywhere. In that case, let's just remove it from the new sizes, the old sizes, and from the types. Otherwise its presence will cause confusion

samijaber: Are you sure we can set these to `0`? I am looking at where `sizes` gets used: it gets fed to `this.state.state` which is provided to every custom component and custom code block that is evaluated.

This makes me worry that we cannot set this to `0`. We should probably stick with the old way of setting it to be safe

packages/react/src/constants/device-sizes.constant.ts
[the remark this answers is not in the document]
murtaza-builder: done

packages/react/src/constants/device-sizes.constant.ts
murtaza-builder: done, as per our conversation

samijaber (approved):

packages/react/src/functions/safe-dynamic-require.ts
```diff
@@ -5,6 +5,5 @@ import { Builder } from '@builder.io/sdk';
 const noop = () => null;
 // Allow us to require things dynamically safe from webpack bundling
 
-export let safeDynamicRequire: typeof require;
+export let safeDynamicRequire: typeof require = noop as any;
 if (typeof globalThis.require === 'function') safeDynamicRequire = eval('require');
-safeDynamicRequire ??= noop as any;
```
samijaber, line 10: Your change here will impact the behavior of how `safeDynamicRequire` is set. I would prefer to rollback as it is unrelated to this PR

more info on this operator: https://www.joshwcomeau.com/operator-lookup?match=logical-nullish-assignment

murtaza-builder: does the new change look good?

samijaber (approved):

PR #1242, opened 2022-09-05
Fixes for patch updates

steve8708:
The copying of objects had too many issues, so in this latest (and most effective to date) approach, we only patch update primitives, and apply them directly to props - effectively updating the core content tree as needed

This doesn't work when there are child blocks modified in the tree, but works well for primitives

The Builder.io editor has been updated accordingly to only patch primitives, which is the bulk of what needs to be fast too (like updating a style, text, etc)

### DIFF
--- a/packages/react/src/components/builder-component.component.tsx
+++ b/packages/react/src/components/builder-component.component.tsx
@@ -98,30 +98,6 @@ const sizeMap = {
   mobile: 'small',
 };
 
-function decorator(fn: Function) {
-  return function argReceiver(...fnArgs: any[]) {
-    // Check if the decorator is being called without arguments (ex `@foo methodName() {}`)
-    if (fnArgs.length === 3) {
-      const [target, key, descriptor] = fnArgs;
-      if (descriptor && (descriptor.value || descriptor.get)) {
-        fnArgs = [];
-        return descriptorChecker(target, key, descriptor);
-      }
-    }
-
-    return descriptorChecker;
-
-    // descriptorChecker determines whether a method or getter is being decorated
-    // and replaces the appropriate key with the decorated function.
-    function descriptorChecker(target: any, key: any, descriptor: any) {
-      const descriptorKey = descriptor.value ? 'value' : 'get';
-      return {
-        ...descriptor,
-        [descriptorKey]: fn(descriptor[descriptorKey], ...fnArgs),
-      };
-    }
-  };
-}
 
 const fetchCache: { [key: string]: any } = {};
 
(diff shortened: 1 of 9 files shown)


packages/react/src/components/builder-component.component.tsx
```diff
@@ -1391,47 +1357,27 @@ export class BuilderComponent extends React.Component<
               const builderModelRe = /builder\.io\/api\/v2\/([^\/\?]+)/i;
               const builderModelMatch = url.match(builderModelRe);
               const model = builderModelMatch && builderModelMatch[1];
-              if (false && Builder.isEditing && model && this.builder.editingModel === model) {
```
mrkoreye, line 1394: nice cleanup

packages/react/src/functions/apply-patch-with-mutation.test.ts
```diff
@@ -0,0 +1,55 @@
+import { applyPatchWithMinimalMutationChain } from './apply-patch-with-mutation';
+
+describe('applyPatchWithMinimalMutationChain', () => {
```
mrkoreye, line 3: 😄

packages/react/src/components/builder-content.component.tsx
```diff
@@ -140,17 +140,17 @@ export class BuilderContent<ContentType extends object = any> extends React.Comp
...
         }
+        let newData = this.state.data as any;
         for (const patch of patches) {
-          applyPatchWithMinimalMutationChain(this.state.data, patch);
+          newData = applyPatchWithMinimalMutationChain(newData, patch, false);
         }
         this.setState({
           updates: this.state.updates + 1,
-          data: this.state.data ? { ...this.state.data } : this.state.data,
+          data: newData,
```
mrkoreye, line 149: It looks like the only change here besides putting `this.state.data` into a variable (`newData`) is that now we don't make data a new object with `{ ...this.state.data }`. Was that causing come of the performance issue?

mrkoreye (approved):

packages/react/src/components/builder-content.component.tsx
steve8708: oh no this is unrelated, just was funky code. I first suspected something was wrong here, then realized it's actually ok, but I think this is the more "proper" functional approach

packages/react/src/functions/apply-patch-with-mutation.test.ts
steve8708: wrote these tests bc I didn't trust my implementation from 2 years ago. turns out it worked as expected, couldn't find any bugs, but hey we have more tests now :D

PR #1244, opened 2022-09-06
[packages/react] add support for locale prop to autoresolve localized text and component inputs

teleaziz:


packages/react/src/components/builder-block.component.tsx
```diff
@@ -11,13 +11,29 @@ import { BuilderStoreContext } from '../store/builder-store';
...
 const kebabCaseToCamelCase = (str = '') =>
   str.replace(/-([a-z])/g, match => match[1].toUpperCase());
 
+const localizedType = '@builder.io/core:LocalizedValue';
+const resolveLocalizedOptions = (options?: any, locale?: string) => {
+  if (!locale || !options) {
+    return options;
+  }
+
+  const result = traverse(options).map(function (el) {
```
teleaziz, line 28: investigating if this is possible on the editor side instead of here

packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@builder.io/sdk",
-  "version": "1.1.27",
+  "version": "1.1.28-5",
```
shyam-builder, line 3: should this be pointing to a dev release instead of a 1.1.29 ?

shyam-builder (approved):

packages/core/package.json
teleaziz: Correct, was waiting for review before cutting a stable release, I'll do this next, thanks Shyam

Issue #1224, opened 2022-08-23
'model' does not exist in type 'Input'


joelwilsonmt:
**Describe the bug**
When registering a component using `Builder.registerComponent()` and defining the inputs, the `type` `model` is missing from `Input`:
<img width="808" alt="image" src="https://user-images.githubusercontent.com/43355957/186251895-c42456f9-e5d3-43f9-b85d-720586f660ee.png">

**To Reproduce**

1. Use `"@builder.io/react": "^2.0.2"` version
2. Create a custom component, and register using `Builder.registerComponent()`
3. Include something like:
```
inputs:  [{
    name: 'test',
    type: 'reference',
    model: 'test',
  }]
```
and see typescript compiler highlighting it as problem "Object literal may only specify known properties, and 'model' does not exist in type 'Input'.ts(2322)"

**Expected behavior**
`model` should work on an `Input` field

**Additional context**
When `model` is added, the visual editor restricts references to the correct model, but typescript will not allow it.

PR #1251, opened 2022-09-10
fix(core): TypeError - Fixes type error when using `Builder.registerComponent` with type of `reference` and restricting to specific `model`.

eligeske:
## Description

When using a custom component with a `reference` input restricted to a model, as per the docs https://www.builder.io/c/docs/custom-components-input-types, a type error is thrown due to missing Input property.

`
Type '{ name: string; friendlyName: string; helperText: string; type: string; model: string; }' is not assignable to type 'Input'.
 Object literal may only specify known properties, and 'model' does not exist in type 'Input'.ts(2322)
`

## Changes
- Adds `model` property to `Input` interface in Builder Core

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -587,6 +587,10 @@ export interface Input {
   showIf?: ((options: Map<string, any>) => boolean) | string;
   /** @hidden */
   copyOnAdd?: boolean;
+  /**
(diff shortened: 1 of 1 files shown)


eligeske:
https://github.com/BuilderIO/builder/issues/1224

teleaziz (approved):
Thanks @eligeske

PR #1252, opened 2022-09-12
[packages/react] fix editing/previewing custom fields in a builder content when it contains a buildercomponent of same model

teleaziz:


packages/react/src/components/builder-content.component.tsx
```diff
@@ -199,6 +199,11 @@ export class BuilderContent<ContentType extends object = any> extends React.Comp
               data: match,
               loading: false,
             });
+            const isPreviewing = (builder.editingModel || builder.previewingModel) === this.name;
+            if (!this.props.inline && this.firstLoad && isPreviewing) {
+              this.firstLoad = false;
+              this.subscribeToContent();
+            }
```
steve8708, line 206: I think we still need this, otherwise I think this opens a new bug that we had in the way past, where say you have

```tsx
<BuilderComponent model="header" />
<BuilderComponent model="page" />
```

unless I'm mistaken the header will also subscribe to changes of page when editing and crazy stuff ensues (suddenly you see your page twice - where the header should be and where the page should be, etc)

packages/react/src/components/builder-content.component.tsx
```diff
@@ -199,6 +199,11 @@ export class BuilderContent<ContentType extends object = any> extends React.Comp
               data: match,
               loading: false,
             });
+            const isPreviewing = (builder.editingModel || builder.previewingModel) === this.name;
+            if (!this.props.inline && this.firstLoad && isPreviewing) {
```
mrkoreye, line 203: Can you add a comment here to clarify the intention behind this change? To me it is not clear just from the code

teleaziz:
This looks like it's closed by mistake , reopening and adding more comments next

mrkoreye (approved):

PR #1253, opened 2022-09-13
fix(qwik): dont pull in `node-fetch` in bundles

mhevery:
## Description

_Screenshot_


packages/sdks/src/functions/get-fetch.ts
```diff
@@ -1,12 +1,21 @@
 import { getGlobalThis } from './get-global-this.js';
 
+/**
+ * The issue is that when building standalone bundle the bundler includes `node-fetch` as
```
samijaber, line 4: which bundler are we talking about here? Is this a qwik optimizer/bundler issue in particular? or a Vite issue? can we add that information for clarity?

mhevery: clarified

samijaber (approved):
Thanks for clarifying. To make sure I understand though: the "bundler" in this case is the Vite bundler used when we build the qwik SDK?

Also, was this causing any crashes/issues, or is it purely a bundle size optimization?

samijaber (changes requested):
Actually, I'm not sure this hack is necessary. Instead, we need to exclude node modules from our vite bundles as a cleaner solution to this problem

samijaber:
See how we do it for our Vue2 SDK:

https://github.com/BuilderIO/builder/blob/74f8c6d49b044aa52a43c3970bbaf5a5cf9bd30d/packages/sdks/output/vue/vue2/vite.config.ts#L14-L18

packages/sdks/src/functions/get-fetch.ts
```diff
@@ -1,14 +1,36 @@
...
-    const nodeFetch = import('node-fetch').then((d) => d.default) as Promise<
+    const nodeFetch = import(`node-fetch`).then((d) => d.default) as Promise<
       typeof global.fetch
     >;
-    return (nodeFetch as any).default || nodeFetch;
+    return (fetch = (nodeFetch as any).default || nodeFetch);
   }
 
-  return (globalFetch as any).default || globalFetch;
+  return (fetch = (globalFetch as any).default || globalFetch);
```
samijaber, line 35: can we undo all the changes to this file now?

mhevery: done

packages/sdks/src/functions/get-fetch.ts
```diff
@@ -1,14 +1,17 @@
...
```
samijaber, line 16: Why is this new mutable `fetch` needed? I don't fully understand

samijaber: (what happens if we don't have it in here)

samijaber (approved):

PR #1261, opened 2022-09-16
fix: 🐛 Qwik event handling

mhevery:
## Description

_Screenshot_


examples/svelte/svelte-vite/yarn.lock
```diff
@@ -0,0 +1,1924 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
samijaber, line 1: Can you delete this yarn file? We want our examples to be as simple as possible, so we only use `npm install` in them

samijaber (approved):
🏆

samijaber:
Ah...like I suspected, this breaks the Vue2 build as you can see in the e2e tests 😕

PR #1266, opened 2022-09-20
[packages/core] Support repeating scrollInView animation, add support for threshold input

teleaziz:
https://www.loom.com/share/2db96a054bc541d2abbe6c921fae79f5


packages/core/src/classes/animator.class.ts
```diff
@@ -174,17 +178,33 @@ export class Animator {
...
+              }
+            }, (animation.duration + (animation.delay || 0)) * 1000 + 100);
           });
+        } else if (
+          animation.repeat &&
+          triggered &&
+          !pendingAnimation &&
+          !isScrolledIntoView(element)
+        ) {
+          // we want to repeat the animation everytime the the element is out of vide and back again
```
mrkoreye, line 205:
```suggestion
          // we want to repeat the animation every time the the element is out of view and back again
```

mrkoreye (approved):

PR #1269, opened 2022-09-21
Add `override` property for overriding built-in component to types and generated docs

steve8708:


packages/react/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### 2.0.6
+
+- added types for `override` option
```
steve8708, line 3: awesome

PR #1272, opened 2022-09-22
fix: 🐛 Columns styling was incorrect

mhevery:
BREAKING CHANGE: 🧨 no

## Description

_Screenshot_

### DIFF
--- a/packages/sdks/package.json
+++ b/packages/sdks/package.json
@@ -40,7 +40,7 @@
   },
   "dependencies": {
     "@builder.io/mitosis": "^0.0.63",
-    "@builder.io/mitosis-cli": "^0.0.23",
+    "@builder.io/mitosis-cli": "^0.0.24",
     "@playwright/test": "^1.24.0",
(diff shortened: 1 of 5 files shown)


samijaber (approved):
Excellent 👌🏽

packages/sdks/package.json
```diff
@@ -39,7 +39,7 @@
     "upgrade-example:all": "bash ./scripts/loop-command.sh upgrade-example \"svelte react-native solid qwik react vue\""
   },
   "dependencies": {
-    "@builder.io/mitosis": "^0.0.62",
+    "@builder.io/mitosis": "^0.0.63",
     "@builder.io/mitosis-cli": "^0.0.23",
```
samijaber, line 43: you need to update the CLI, since that has a dependency on mitosis and is still pointing to the older version I believe

mhevery: Opps, thanks for the tip

PR #1275, opened 2022-09-27
Android SDK cleanup

steve8708:


shyam-builder (approved):
So I'm assuming you are just removing all the `isEditing` flow related code?

steve8708:
exactly, to reduce scope (and given some feedback) I want to just remove all editing features and just focus on two remaining needs of SDK

1. make it a package that people can install into their Android apps
2. make sure we can render a good set of Builder content (working with Sami on that example space)

PR #1276, opened 2022-09-27
Fix bug where SSR was not working for data binding

shyam-builder:
## Description

SSR with data-binding was not working because the way we were recognizing the server had changed and was not working as expected. Revert back to the old way of recognizing the server for data binding cases.

_Screenshot_

Before:
image

After:
image

### DIFF
--- a/packages/react/src/functions/safe-dynamic-require.ts
+++ b/packages/react/src/functions/safe-dynamic-require.ts
@@ -6,5 +6,11 @@ const noop = () => null;
 // Allow us to require things dynamically safe from webpack bundling
 
 export let safeDynamicRequire: typeof require;
-if (Builder.isServer) safeDynamicRequire = eval('require');
+
+/*
+ * The if condition below used to be if (typeof globalThis.require === "function"
+ * That broke in case Builder was running on the server (Next, SSR use-cases) where
+ * globalThis.require was undefined. Avoiding use of Builder.isServer for Cloudflare worker cases
+ */
+if (typeof require === 'function') safeDynamicRequire = eval('require');
 safeDynamicRequire ??= noop as any;
--- a/packages/react/src/functions/safe-dynamic-require.ts
+++ b/packages/react/src/functions/safe-dynamic-require.ts
@@ -6,5 +6,5 @@ const noop = () => null;
 // Allow us to require things dynamically safe from webpack bundling
 
 export let safeDynamicRequire: typeof require;
-if (typeof globalThis.require === 'function') safeDynamicRequire = eval('require');
+if (Builder.isServer) safeDynamicRequire = eval('require');
 safeDynamicRequire ??= noop as any;


packages/react/src/functions/safe-dynamic-require.ts
```diff
@@ -6,5 +6,5 @@ const noop = () => null;
 // Allow us to require things dynamically safe from webpack bundling
 
 export let safeDynamicRequire: typeof require;
-if (typeof globalThis.require === 'function') safeDynamicRequire = eval('require');
+if (Builder.isServer) safeDynamicRequire = eval('require');
```
samijaber, line 9: Can we: 
- Understand why the change was made in the first place? Was it when @Dkendal was trying to get the React SDK working in the Hydrogen environment?
- Document this reversal and the difference between both attempts so that we have more information in the future about t these 2 choices?
- 
```suggestion
/* 
 * We initially tried to check the existence of `typeof globalThis.require === 'function'` but that didn't work because of X.
 * This is why we make sure to use `Builder.isServer` to determine whether we can use `require` or not
 */
if (Builder.isServer) safeDynamicRequire = eval('require');
```

samijaber (changes requested):
I believe reverting this change breaks React SDK in Hydorgen environments. Would love some more context on the tradeoff between both approaches

packages/react/src/functions/safe-dynamic-require.ts
Dkendal: `Builder.isServer` is not reliable enough to decide to use `eval(...)` - eval is not available in cloudfront workers.

I would also like to know why the typecheck on `globalThis.require` doesn't work as intented, my understanding was that it should function in all environments.

The only way that you can programatically detect that the server code is running in cloudflare workers is if `caches.default` is defined, which is ... less that ideal.

shyam-builder: Which is super weird, because if I run `globalThis.require` in a Node.JS command prompt, I do get it...

Dkendal: where'd the report come from?

Dkendal: image
should work assume they are nodejs v12 >

shyam-builder: Updated as per our discussion in the thread. Added comments for history

Dkendal (approved):

samijaber (approved):

ersinakinci:
Hey @shyam-builder I've been working on this actively, as well. Some notes from my findings:

- I also found the change to `safeDynamicRequire` was broken on Next, and I worked around by reverting the change. Glad that a solution could be found that's also Hydrogen-compatible. 🎉 
- The function that runs in the browser when evaluating dynamic bindings has access to many more variables than the VM sandbox that runs on the server. Probably the most important among these is `context`, since `context.builderContent.data.*` is a common data binding to populate blocks from model fields. Can we populate the sandbox in such a way that there's parity between what's available in the browser vs. what's available on the server?
- If a field is dynamically bound to a state property that itself is populated from a content input, very often this results in a hydration error. Happens with symbols a _lot_. For example, if I have a symbol with a text box bound to `state.label`, and I have a `label` content input, during SSR `state.label` will often be empty. If my content field has a default value (e.g., `"Click me!"`), then I get a hydration error. It seems that state gets populated in the browser in a way that accounts for default values, but the same doesn't happen during SSR.
- Another common source of problems is the "repeat for each" feature in the Visual Editor. If you bind to an array--for example, `state.menuItems`--then you should be able to bind to each array member's value by appending `Item`: `state.menuItemsItem`. This works well in the browser, but during SSR, the sandbox has no concept of iterables. If you try to access a property on the iterable, such as `state.menuItemsItem.label`, the VM throws an error. At the very least, the docs should be updated to make sure that people use a conditional to avoid errors during SSR.
- Any errors thrown by the VM seem to be swallowed. I see that they get put into an errors array, but as an end user of the SDK, I have no idea how to see those errors, and if I could, I would have no way of understanding that they're occurring during SSR. Adding a simple message like `Builder error on the server: ______` and updating the docs so that we know how to access that errors array would be 💥.

ersinakinci:
@shyam-builder created a PR on your branch: https://github.com/shyam-builder/builder/pull/1

PR #1289, opened 2022-09-28
Add context in SSR flow

shyam-builder:
## Description

The function that runs in the browser when evaluating dynamic bindings has access to many more variables than the VM sandbox that runs on the server. Probably the most important among these is context, since context.builderContent.data.* is a common data binding to populate blocks from model fields. 

This PR populates the sandbox in such a way that there's parity between what's available in the browser vs. what's available on the server


ersinakinci:
I actually authored this change. Let me know if you have questions.

Context: https://github.com/BuilderIO/builder/pull/1276#issuecomment-1260191993

mrkoreye (approved):

PR #1291, opened 2022-09-29
Better data binding execution errors

ersinakinci:
Data bindings silently failing to execute during SSR is major cause of hydration issues. These failures are almost impossible to debug without some deep knowledge of the React SDK.

Fortunately, the React SDK logs these failures. Unfortunately, the errors are not very helpful and are hidden behind an undocumented `DEBUG=true` environment variable:

https://github.com/BuilderIO/builder/blob/c2663237c83435c04d92e6b81d658e5759e9396b/packages/react/src/functions/string-to-function.ts#L143-L153

The errors aren't helpful because:

1. They misleadingly state that there is a `Builder custom code error`. I think that most users would understand "custom code" to refer to the Visual Editor's custom/content JS feature, but the code that's executed here actually comes from data bindings.
2. The errors log the code that failed to execute but without any context as to where it's from.

My PR does the following:

1. Gives a better description of the error. An example on the server:

```
The Builder React SDK failed to execute the following data binding on the server:

  var _virtual_index=context.builderContent;return _virtual_index

The error was

  ReferenceError: context is not defined

You can inspect the binding and its code by visiting https://www.builder.io/content/87aaf4bc20b140f98ba3db2aba959453/edit?activeDesignerTab=3&selectedBlock=builder-a79fc1ddb9c443a6a9b1b69d77048aa8 and entering the JSON view for the selected block. You can enter the JSON view by right clicking on the block in the Layers tab or by pressing Cmd/Ctrl+E with the block selected.

For a more detailed error message, restart your server with DEBUG=true set in your environment variables. If you're using npm, yarn, or a similar tool to run your development server, be sure to add DEBUG=true to your script in package.json so that your server can access the environment variable.
```

And if `DEBUG=true` is set, the above example looks like:

```
The Builder React SDK failed to execute the following data binding on the server:

  var _virtual_index=context.builderContent;return _virtual_index

The error was

  ReferenceError: context is not defined

You can inspect the code by visiting https://www.builder.io/content/87aaf4bc20b140f98ba3db2aba959453/edit?activeDesignerTab=3&selectedBlock=builder-a79fc1ddb9c443a6a9b1b69d77048aa8 and entering the JSON view for the selected block. You can enter the JSON view by right clicking on the block in the Layers tab or by pressing Cmd/Ctrl+E with the block selected.

***BEGIN DEBUG INFO***

Block type: Text
Block ID: builder-a79fc1ddb9c443a6a9b1b69d77048aa8
Content item ID: 87aaf4bc20b140f98ba3db2aba959453
Content item name: Blank page

Server VM sandbox:

{
  "deviceSize": "large",
  "location": {
    "path": "",
    "query": {}
  },
  "isBrowser": false,
  "isServer": true,
  "device": "desktop",
  "state": {
    "deviceSize": "large",
    "location": {
      "path": "",
      "query": {}
    },
    "isBrowser": false,
    "isServer": true,
    "device": "desktop"
  },
  "event": null
}

Error stack trace:

ReferenceError: context is not defined
    at vm.js:1:20
    at Script.runInContext (node:vm:139:12)
    at VM.runScript (/Users/ersin/projects/hydration-test/node_modules/vm2/lib/vm.js:285:18)
    at /Users/ersin/projects/hydration-test/node_modules/vm2/lib/vm.js:507:16
    at timeout_bridge.js:1:1
    at Script.runInContext (node:vm:139:12)
    at doWithTimeout (/Users/ersin/projects/hydration-test/node_modules/vm2/lib/vm.js:132:29)
    at VM.run (/Users/ersin/projects/hydration-test/node_modules/vm2/lib/vm.js:506:10)
    at final (webpack-internal:///../builder/packages/react/dist/builder-react.cjs.js:349:33)
    at BuilderBlock.getElement (webpack-internal:///../builder/packages/react/dist/builder-react.cjs.js:775:35)

***END DEBUG INFO***
```

Whether `DEBUG=true` is set or not, the original error is logged separately for good measure, although it's not terribly helpful since it's being eval'd from a string inside of a block. This behavior is similar to how React logs the raw stack trace and then provides a more helpful error immediately afterward.

A similar error is logged in the browser:

CleanShot 2022-09-28 at 17 30 30

2. My PR makes these errors visible by default, not just behind `DEBUG=true`, and logs them as `console.error`, not `console.warn` or `console.debug`. If a data binding doesn't execute, that's really important info and we should consider it as breaking the page!

3. The `DEBUG` environment variable is now used to toggle some verbose info on the server (see above). It has no effect on browser errors.

Note that while the original intention was to improve errors on the server where there was almost no visibility into what was going wrong, my PR also improves the browser errors.

### DIFF
--- a/packages/react/package-lock.json
+++ b/packages/react/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@builder.io/react",
-  "version": "2.0.7",
+  "version": "2.0.8-4",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@builder.io/react",
-      "version": "2.0.7",
+      "version": "2.0.8-4",
       "license": "MIT",
       "dependencies": {
         "@builder.io/sdk": "^1.1.30",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@builder.io/react",
-  "version": "2.0.7",
+  "version": "2.0.8-4",
   "description": "",
   "keywords": [],
   "main": "dist/builder-react.cjs.js",
--- a/packages/react/src/components/builder-block.component.tsx
+++ b/packages/react/src/components/builder-block.component.tsx
@@ -516,7 +516,7 @@ export class BuilderBlock extends React.Component<
     // tslint:disable-next-line:comment-format
     ///REACT15ONLY finalOptions.className = finalOptions.class
 
-    if (Builder.isEditing) {
+    if (Builder.isEditing || !Builder.isBrowser) {
       // TODO: removed bc JS can add styles inline too?
       (finalOptions as any)['builder-inline-styles'] = !(options.attr && options.attr.style)
         ? ''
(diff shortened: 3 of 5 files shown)


ersinakinci:
cc: @samijaber, you might want to add this to the v2 SDKs.

PR #1293, opened 2022-09-30
Fix issue with require getting triggered on client side

shyam-builder:
## Description

The fix for getting SSR to work broke client side. This fix tries to work around that.

More details in the comments of the file, but basically,

1. Client side - `if (typeof require === 'function')` is over-optimized to `if ('function' === 'function')`
2. Node - `if (typeof globalThis.require === 'function')` does not work since `globalThis.require` is undefined
3. Cloudflare edge workers - Either `if (typeof globalThis.require === 'function')` or `if (typeof require === 'function')` will work, but `Builder.isServer` will not work

Hence the final condition that I think should work across all cases

### DIFF
--- a/packages/react/src/functions/safe-dynamic-require.ts
+++ b/packages/react/src/functions/safe-dynamic-require.ts
@@ -5,26 +5,26 @@ import { Builder } from '@builder.io/sdk';
 const noop = () => null;
 // Allow us to require things dynamically safe from webpack bundling
 
-export let safeDynamicRequire : typeof require;
+export let safeDynamicRequire: typeof require;
 
 /*
- * The if condition below used to be 
- * 
+ * The if condition below used to be
+ *
  *     if (typeof globalThis.require === "function")
- * 
+ *
  * That broke in case Builder was running on the server (Next, SSR use-cases) where
  * globalThis.require was undefined. Avoiding use of Builder.isServer for Cloudflare worker cases
- * That said, if change it to 
- * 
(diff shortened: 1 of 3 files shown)


packages/react/src/functions/safe-dynamic-require.ts
```diff
@@ -8,9 +8,29 @@ const noop = () => null;
...
+ * In Summary:
+ *
+ * 1. Node -> globalThis.require does not work
+ * 2. Client -> typeof require === 'function' does not work because of overoptimization by the compiler
+ * 3. Cloudflare edge -> only globalThis.require works
  */
-if (typeof require === 'function') safeDynamicRequire = eval('require');
+if (typeof globalThis.require === 'function' || Builder.isServer) {
+  safeDynamicRequire = eval('require');
+}
```
samijaber, line 34: One suggestion for the future that might save us headaches for issues like this one: getting rollup to generate 2 bundles: one for the browser & another for the server: https://stackoverflow.com/a/62471802

We could pass an env var from the rollup config like `process.env.IS_ROLLUP_CLIENT` and then add a toggle here based on that. We can even toggle to import completely separate files:

```js
// packages/react/src/functions/safe-dynamic-require/index.ts

if (process.env.IS_ROLLUP_CLIENT) {
  require('./client-require.ts')
} else {
  require('./server-require.ts')
}
``

samijaber (approved):

PR #1296, opened 2022-09-30
[packages/sdks] support localization

teleaziz:


examples/svelte/localized-sveltekit/src/hooks.server.js
```diff
@@ -0,0 +1,48 @@
+import { getLocaleFromPathname } from './utils';
+const defaultLocale = 'en';
+// Match this with the locales defined in your builder space
+const supportedLocales = ['en', 'fr', 'de'];
+const routeRegex = new RegExp(/^\/[^.]*([?#].*)?$/);
```
samijaber, line 5: Would be great to explain this regex with a comment, and a link to a regex fiddle like https://regexr.com.

To an outsider (like me), I have no idea what this is looking up, it's somewhat cryptic.

examples/svelte/localized-sveltekit/README.md
```diff
@@ -0,0 +1,41 @@
+# Builder.io example with SvelteKit
```
samijaber, line 1:
```suggestion
# Builder.io example with SvelteKit (+ localization)
```

samijaber (approved):
Sweet. Just 1 regex that should be explained for end-users.

samijaber:
Looks like this works correctly in the dev version you released? If so, let's bring it to all SDKs via `yarn run release:all` in the root of `packages/sdks`!

PR #1305, opened 2022-10-05
SDKs: Add external `track` function

samijaber:
## Description

- export our tracking function for users
- fix A/B test tracking by adding variation ID to track call


packages/sdks/src/functions/track.ts
```diff
@@ -38,12 +44,19 @@ const getTrackingEventData = async ({
...
-  orgId: Event['data']['ownerId'];
-  contentId: Event['data']['contentId'];
-  [index: string]: any;
-};
+type EventProperties = Pick<Event, 'type'> &
+  Pick<Event['data'], 'ownerId' | 'contentId'> & {
+    /**
+     * This is the ID of the space that the content belongs to.
+     */
+    orgId: Event['data']['ownerId'];
```
steve8708, line 52: this should always be called `apiKey` externally and referenced as such in the jsDocs (e.g.the jsDoc should say "this is your public API key")

PR #1309, opened 2022-10-07
RSC SDK beta

steve8708:


samijaber (commented):
@steve8708 can we delete all the redundant stuff in `e2e/rcs/src/*`? I assume it came from a starter and we don't need any of it.

steve8708:
oh so that is actually incredibly useful right now for testing things thoroughly. if it's ok perhaps I can keep wrangling with that and then when things are looking good can distill it down and send a future PR with the most minimal e2e flow

steve8708:
the amount of merge conflicts keeps growing, @samijaber can you give this a 👍 or 👎

samijaber:
Oh sorry, I totally forgot to come back to this PR 🤦🏽

packages/sdks/src/blocks/util.ts
```diff
@@ -1,6 +1,5 @@
-export function markSerializable<FN extends (...args: any[]) => any>(
```
samijaber, line 1: cc @mhevery I think you had some special logic here but stringifying the functions does the trick instead

packages/sdks/src/functions/get-block-styles.ts
```diff
@@ -23,25 +23,8 @@ const getStyleForTarget = ({
...
-        ...(styles.small
-          ? {
-              [getMaxWidthQueryForSize('small')]: convertStyleObject(
-                styles.small
-              ),
-            }
-          : {}),
-      };
+      // By default we do not want inline styles
+      return {};
```
samijaber, line 27: I simplified all this logic recently to only run on React Native, so you can take my changes from the merge conflict.

packages/sdks/src/scripts/init-editing.ts
```diff
@@ -23,7 +23,12 @@ export const registerInsertMenu = () => {
   });
 };
 
+let isSetupForEditing = false;
 export const setupBrowserForEditing = () => {
+  if (isSetupForEditing) {
+    return;
+  }
+  isSetupForEditing = true;
```
samijaber, line 31: Why is this needed? Asking because these side-effects might not consistently work for all framework environments, so if preferable I'd rather not rely on them (or know exactly what they're needed for at least)

packages/sdks/src/blocks/util.ts
```diff
@@ -1,6 +1,5 @@
...
+export function changeFunctionToString<FN extends (...args: any[]) => any>(
   fn: FN
-): FN {
-  (fn as any).__qwik_serializable__ = true;
-  return fn;
+): string {
+  return `return (${fn.toString()}).apply(this, arguments)`;
```
samijaber, line 4: note to self: this is the same as `serializeFn` in `src/functions/register-component.ts`, can dedupe.

samijaber (approved):

packages/sdks/src/blocks/util.ts
samijaber: Actually, this function is simply making sure we serialize the fn ahead of time, `serializeFn` is no longer necessary

packages/sdks/src/scripts/init-editing.ts
steve8708: we could move to a window var - the point though is just that we want to be sure this only ever runs one time, not multiple, or we could face issues

samijaber: `init-editing` runs in the `onMount` hook in every framework besides RSC though, so it's guaranteed to only run once per `RenderContent`. If this is RSC-specific, can probably add a piece of boolean state in `BuilderEditingWrapper` to make sure it only runs once.

steve8708: that's exactly the problem, many people will have multiple <RenderContent /> components on one page, we don't want all this stuff to run multiple times (e.g. the message event binding for code execution - bad for perf and reliability to have these register multiple times)

steve8708: also tmk this would not be an issue with bundlers, the issue with those is tree shaking, but the "side effect" is just a local var, I've never seen a case where this is problem with even the most aggressive tree shakers

samijaber: Fair enough, we can go with this approach. 👍🏽

PR #1311, opened 2022-10-11
[packages/react] fix hydrogen error

teleaziz:
Screen Shot 2022-10-11 at 12 31 07 PM

To reproduce:

```
git clone https://github.com/BuilderIO/builder-shopify-hydrogen
cd builder-shopify-hydrogen
yarn
yarn add @builder.io/react@latest
yarn build
yarn preview
```


mrkoreye (commented):
Can you update the comment above the condition to accommodate the new logic?

packages/react/src/functions/safe-dynamic-require.ts
```diff
@@ -26,10 +26,12 @@ export let safeDynamicRequire: typeof require;
  * In Summary:
  *
  * 1. Node -> globalThis.require does not work
- * 2. Client -> typeof require === 'function' does not work because of overoptimization by the compiler
- * 3. Cloudflare edge -> only globalThis.require works
+ * 2. Cloudflare edge -> only globalThis.require works
  */
-if (typeof globalThis.require === 'function' || Builder.isServer) {
+if (
+  typeof globalThis.require === 'function' ||
```
mrkoreye, line 32: Do we need to change this to 
```suggestion
  typeof globalThis?.require === 'function' ||
```
in case `globalThis` is undefined?

PR #1317, opened 2022-10-17
[customBreakpoints] feat: Support for custom breakpoints for Mitosis based sdks

murtaza-builder:
## Description
Added support for custom breakpoints feature.
Added corresponding E2E tests.


packages/sdks/e2e/tests/src/tests/main.spec.ts
```diff
@@ -57,4 +70,54 @@ test.describe(targetContext.name, () => {
       text: 'Mattel Certified by Great Place to Work and Named to Fast Company’s List of 100 Best Workplaces for Innovators',
     });
   });
+
+  test.describe('custom-breakpoints', () => {
```
shyam-builder, line 74: would it be possible to add one set of tests for where the custom breakpoints are not set, so we test the default flow as well and ensure nothing is broken there?

shyam-builder (commented):
Great stuff

packages/sdks/src/types/builder-content.ts
```diff
@@ -12,6 +12,10 @@ export interface BuilderContentVariation {
     jsCode?: string;
     tsCode?: string;
     httpRequests?: { [key: string]: string };
+    breakpoints?: {
```
mrkoreye, line 15: Why are we adding it to content data here? I thought we decided to use meta property for this, since in general `data` is for customer added properties and data?

mrkoreye: https://builder-internal.slack.com/archives/C01V9EH4T1N/p1663611462024529?thread_ts=1663610512.642029&cid=C01V9EH4T1N

packages/sdks/e2e/tests/src/tests/main.spec.ts
murtaza-builder: done

packages/sdks/src/types/builder-content.ts
murtaza-builder: Thankyou! I keep confusing myself. Fixed it now.

shyam-builder (approved):

samijaber (approved):
Beautiful & perfect 🎉

PR #1332, opened 2022-10-27
Button Rendered and made it clickable

krishnavivekesmagico:
Core:Button is interpreted and rendered. Made it clickable as Button.


samijaber:
@shyam-builder i assume we can close this PR?

PR #1333, opened 2022-10-29
fix: Support bindings to style attribute

mhevery:
## Description

_Screenshot_

### DIFF
--- a/packages/sdks/e2e/tests/src/tests/main.spec.ts
+++ b/packages/sdks/e2e/tests/src/tests/main.spec.ts
@@ -70,6 +70,16 @@ test.describe(targetContext.name, () => {
       text: 'Mattel Certified by Great Place to Work and Named to Fast Company’s List of 100 Best Workplaces for Innovators',
     });
   });
+  test('data-binding-styles', async ({ page }) => {
+    await page.goto('/data-binding-styles');
+
+    await expect(
+      await getElementStyleValue({
+        locator: page.locator(`text="This text should be red"`).first(),
+        cssProperty: 'color',
+      })
+    ).toBe('red');
+  });
 
   test('image', async ({ page }) => {
     await page.goto('/image');
--- a/packages/sdks/e2e/tests/src/tests/main.spec.ts
+++ b/packages/sdks/e2e/tests/src/tests/main.spec.ts
@@ -75,7 +75,7 @@ test.describe(targetContext.name, () => {
 
     await expect(
       await getElementStyleValue({
-        locator: page.locator(`.builder-text`).first(),
+        locator: page.locator(`text="This text should be red..."`),
         cssProperty: 'color',
       })
     ).toBe('rgb(255, 0, 0)');
--- a/packages/sdks/e2e/tests/src/tests/main.spec.ts
+++ b/packages/sdks/e2e/tests/src/tests/main.spec.ts
@@ -75,13 +75,13 @@ test.describe(targetContext.name, () => {
 
     await expect(
       await getElementStyleValue({
-        locator: page.locator(`text="This text should be red"`).first(),
+        locator: page.locator(`.builder-text`).first(),
         cssProperty: 'color',
       })
-    ).toBe('red');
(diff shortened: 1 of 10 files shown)


packages/sdks/src/functions/get-block-bindings.ts
```diff
@@ -0,0 +1,21 @@
...
+        const styleKey = key.substring('style.'.length);
+        const style = props.style || (props.style = {});
+        style[styleKey] = evaluate({ code: bindings[key], state, context: {} });
+      } else {
+        props[key] = evaluate({ code: bindings[key], state, context: {} });
+      }
+    }
+  }
+  return props;
+}
```
samijaber, line 21: This code should be evaluating all bindings already: https://github.com/BuilderIO/builder/blob/2d495aba29af37791b283d254f39dc1d5fa6f437/packages/sdks/src/functions/get-processed-block.ts#L8-L33

And it's called due to this flag being `true`: https://github.com/BuilderIO/builder/blob/2d495aba29af37791b283d254f39dc1d5fa6f437/packages/sdks/src/components/render-block/render-block.lite.tsx#L76


so `props.useBlock` should have all bindings evaluated correctly. Why is this function needed? If there's a bug in `evaluateBindings`, the bug should be fixed there directly

samijaber (changes requested):

packages/sdks/src/functions/get-block-bindings.ts
mhevery: hey @samijaber so I need some help figuring out what is going on here.

```typescript
    get component(): Nullable<RegisteredComponent> {
      const componentName = getProcessedBlock({
        block: props.block,
        state: props.context.state,
        context: props.context.context,
        shouldEvaluateBindings: false,  // <=== THIS LINE
      }).component?.name;
 ```

So you added `false` at this location, and that is the reason why the bindings are not being processed. Any idea as to why?

Comment (mhevery) [reply packages/sdks/src/functions/get-block-bindings.ts]: I have resolved your suggestions. PTAL

File: packages/sdks/src/functions/get-block-properties.ts [hunk 2]
```diff
@@ -4,8 +4,22 @@ export function getBlockProperties(block: BuilderBlock) {
   return {
     ...block.properties,
     'builder-id': block.id,
+    style: toStyleAttribute(block.style),
```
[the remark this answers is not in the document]
samijaber: you'll need to import `TARGET` at the top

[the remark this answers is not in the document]
mhevery: done

packages/sdks/e2e/tests/src/tests/main.spec.ts
```diff
@@ -75,13 +75,13 @@ test.describe(targetContext.name, () => {
 
     await expect(
       await getElementStyleValue({
-        locator: page.locator(`text="This text should be red"`).first(),
+        locator: page.locator(`.builder-text`).first(),
```
samijaber, line 78: This might not work in react-native, since react native does not have CSS classes. Selection by text will be more accurate

mhevery: But this is what fails the test in Solid.

samijaber:
Vue tests are now failing. I would recommend doing the following locally to make sure everything passes:

- `yarn run build:mitosis`: runs mitosis
- `yarn run e2e:build:all`: runs all build commands needed for e2e tests
- `yarn run e2e:run:all`: runs all e2e tests


and then testing each e2e test that may be having issues.

packages/sdks/e2e/tests/src/tests/main.spec.ts
samijaber: weird that this would be an issue, but regardless: if you need to do something just for react native, this is how:

```ts
const isRNSDK = process.env.SDK === 'reactNative';
```

and then check things in a different way for React Native.

samijaber (approved):
Yesss 🏆

PR #1399, opened 2022-11-22
fix: 🐛 Updated columns component to use custom breakpoint value

murtaza-builder:
## Description

Made Columns component to use custom breakpoints when available instead of using older hardcoded values

### DIFF
--- a/packages/sdks/src/blocks/columns/columns.lite.tsx
+++ b/packages/sdks/src/blocks/columns/columns.lite.tsx
@@ -1,8 +1,10 @@
 import RenderBlocks from '../../components/render-blocks.lite';
 import { For, useStore } from '@builder.io/mitosis';
 import type { BuilderBlock } from '../../types/builder-block';
-import { getSizesForBreakpoints, SizeName } from '../../constants/device-sizes';
-import { Breakpoints } from '../../types/builder-content';
+import { getSizesForBreakpoints } from '../../constants/device-sizes';
+import type { SizeName } from '../../constants/device-sizes';
+import type { Breakpoints } from '../../types/builder-content';
+import RenderInlinedStyles from '../../components/render-inlined-styles.lite';
 
 type CSS = {
   [key: string]: string;
@@ -78,6 +80,7 @@ export default function Columns(props: ColumnProps) {
         '--column-margin-left-tablet': state.maybeApplyForTablet(marginLeft),
       };
     },
+
     getWidthForBreakpointSize(size: SizeName) {
       const breakpointSizes = getSizesForBreakpoints(
         props.customBreakpoints || {}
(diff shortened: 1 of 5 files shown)


mrkoreye (commented):
LGTM but let's wait for @samijaber to take a look before merging

packages/sdks/src/blocks/columns/columns.lite.tsx
```diff
@@ -74,24 +80,55 @@ export default function Columns(props: ColumnProps) {
...
+
+    getWidthForBreakpointSize(size: SizeName) {
+      const breakpointSizes = getSizesForBreakpoints(
+        props.customBreakpoints || {}
+      );
+
+      return breakpointSizes[size].max;
+    },
+
+    get columnsStyles(): string {
```
shyam-builder, line 92: can you add a comment with some info on why we moved this from `css` to an interpolated string `styles`

shyam-builder (approved):
Would wait for a look-over from @samijaber before merging this in

packages/sdks/src/blocks/columns/columns.lite.tsx
murtaza-builder: sure, will do

murtaza-builder: done, @ln #130

packages/sdks/src/components/render-block/render-component.lite.tsx
```diff
@@ -3,12 +3,14 @@ import BlockStyles from './block-styles.lite';
...
 import { For, Show, useMetadata } from '@builder.io/mitosis';
 import type { BuilderContextInterface } from '../../context/types.js';
+import type { Breakpoints } from '../../types/builder-content.js';
 
 export interface RenderComponentProps {
   componentRef: any;
   componentOptions: any;
   blockChildren: BuilderBlock[];
   context: BuilderContextInterface;
+  customBreakpoints?: Breakpoints;
```
samijaber, line 13: I'm confused about this prop. It's not being used anywhere in this component. Can it be removed?

packages/sdks/src/blocks/columns/columns.lite.tsx
```diff
@@ -24,6 +28,8 @@ export interface ColumnProps {
   stackColumnsAt?: StackColumnsAt;
   // TODO: Implement this when support for dynamic CSS lands
   reverseColumnsWhenStacked?: boolean;
+
+  customBreakpoints?: Breakpoints;
```
samijaber, line 32: Ok, I think I see how this works!

It's a good solution, but I'm wondering if we can avoid the need for an extra prop altogether. We should be able to read this value from the Builder context! We already consumer the context in other blocks like Symbol, Form.

```
export default function Columns(props: ColumnProps) {
  const builderContext = useContext(BuilderContext);

// inside getWidthForBreakpointSize
  const breakpointSizes = getSizesForBreakpoints(
        builderContext.content?.meta?.breakpoints || {}
      );
```

This way, you don't have to pass `customBreakpoints` around in 3 different components. What do you think?

murtaza-builder: We could have used it, but when I was working on it, I didn't see context being used in columns.lite. so came up with this approach. I needed it to be cleaner and didn't want to stuff things in context again.
But this can be done, provided it works without surprises, but in another PR? later pls??

samijaber (approved):
LGTM.

Don't forget to reuse the `BuilderContext` in a future PR 😄

shyam-builder (approved):

PR #1401, opened 2022-11-24
Use breakpoints passed from editor on priority inside React SDK

murtaza-builder:
When react sdk is used inside the editor, we need to use the breakpoints values coming from editor on priority vs the breakpoints values available in passed content entry.
This is for cases when fresh breakpoints are set in org settings but they haven't yet persisted in content and to ensure editor to render content iframe based on this fresh breakpoints values (over the one persisted aka stale in content.meta.breakpoints)

Editor PR: [https://github.com/bridgeproject/builder-internal/pull/3709](https://github.com/bridgeproject/builder-internal/pull/3709)

### DIFF
--- a/packages/react/src/components/builder-block.component.tsx
+++ b/packages/react/src/components/builder-block.component.tsx
@@ -1,5 +1,6 @@
 /** @jsx jsx */
-
+/* ts-ignore */
+// @ts-ignore
 import { Builder, builder, BuilderElement, Component } from '@builder.io/sdk';
 import { ClassNames, jsx } from '@emotion/core';
 import React from 'react';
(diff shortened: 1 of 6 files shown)


mrkoreye (approved):

packages/react/src/components/builder-block.component.tsx
```diff
@@ -1,5 +1,6 @@
 /** @jsx jsx */
-
+/* ts-ignore */
+// @ts-ignore
```
samijaber, line 3: Is this still needed? Ideally we'd not persist the ts-ignore, completely removes type-checking

murtaza-builder: done, removed.

shyam-builder (approved):

PR #1407, opened 2022-12-06
fix: Adding missing pixel only when some blocks are available

murtaza-builder:
## Description

Updated the missing-pixel-block-addition logic to apply only when there is atleast a block already available in `data.blocks`

_Screenshot_
on local dev server - 
image

### DIFF
--- a/packages/react/test/basic.test.tsx
+++ b/packages/react/test/basic.test.tsx
@@ -184,7 +184,7 @@ describe('Content changes when new content provided', () => {
 });
 
 describe('Builder Pixel', () => {
-  it('Should be added automatically if missing in blocksString', () => {
+  it('Should NOT be added if missing in blocksString', () => {
     const renderedBlock = reactTestRenderer.create(
       <BuilderPage
         model="page"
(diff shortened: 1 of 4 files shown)


mrkoreye:
This approach makes sense, but I think either in addition to this or instead of we want to prevent adding the pixel to the blocks array if Builder is previewing. The reason is to prevent adding the pixel in the Builder preview.

samijaber (approved):
Sweet, thanks for adding tests too 👍🏽

packages/react/test/basic.test.tsx
```diff
@@ -199,7 +199,7 @@ describe('Builder Pixel', () => {
     expect(renderedBlock).toMatchSnapshot();
   });
 
-  it('Should be added automatically if missing in blocks array', () => {
+  it('Should NOT be added if missing in blocks array', () => {
```
teleaziz, line 202: Should this be : `Should not add pixel if emptry blocks array` ?

packages/react/test/basic.test.tsx
```diff
@@ -184,7 +184,7 @@ describe('Content changes when new content provided', () => {
 });
 
 describe('Builder Pixel', () => {
-  it('Should be added automatically if missing in blocksString', () => {
+  it('Should NOT be added if missing in blocksString', () => {
```
teleaziz, line 187: I think the text of the test could use a revisit

teleaziz (approved):

PR #1698, opened 2023-01-23
[packages/react] donot write variation id to dom, causes hydration er…

teleaziz:
…rors in react 18

### DIFF
--- a/packages/react/src/components/builder-component.component.tsx
+++ b/packages/react/src/components/builder-component.component.tsx
@@ -1113,9 +1113,6 @@ export class BuilderComponent extends React.Component<
                           <div
                             data-builder-component={this.name}
                             data-builder-content-id={fullData.id}
-                            data-builder-variation-id={
-                              fullData.testVariationId || fullData.variationId || fullData.id
-                            }
                           >
                             {!codegen && this.getCss(data) && (
                               <style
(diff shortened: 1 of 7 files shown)


packages/react/src/components/builder-component.component.tsx
```diff
@@ -1113,9 +1113,12 @@ export class BuilderComponent extends React.Component<
                           <div
                             data-builder-component={this.name}
                             data-builder-content-id={fullData.id}
-                            data-builder-variation-id={
-                              fullData.testVariationId || fullData.variationId || fullData.id
-                            }
+                            {...(this.isPreviewing
+                              ? {
+                                  'data-builder-variation-id':
```
mrkoreye, line 1118: Are we sure we do not rely on this attribute in any scripts? Also, do we have any documentation that references this attribute anywhere, in case a customer is relying on it being there?

packages/react/src/components/variants-provider.component.tsx
```diff
@@ -29,6 +29,7 @@ const variantsScript = (variantsString: string, contentId: string) =>
     variants.forEach(function (template) {
       document.querySelector('template[data-template-variant-id="' + template.id + '"]').remove();
     });
+    document.getElementById('variants-script-${contentId}').remove();
```
mrkoreye, line 32: Is this just cleanup or is it solving an issue?

teleaziz: solving an issue that happens only on prod builds, the script was being kept in ssr causing hydration issues in react 18

packages/react/src/components/builder-component.component.tsx
teleaziz: it's only used by editor

mrkoreye (approved):

PR #1800, opened 2023-02-08
Upgraded to use latest webcomponents version to get support for custom breakpoints

murtaza-builder:
## Description
Upgraded the webcomponents sdk script version used in the sdk to recent version, that uses the latest react sdk (supporting custom breakpoints feature)

Note: This change needs to be deployed after the webcomponents change with v1.3.49 is deployed.

### DIFF
--- a/packages/angular/src/app/modules/builder/utils/constants.ts
+++ b/packages/angular/src/app/modules/builder/utils/constants.ts
@@ -1,0 +1,3 @@
+// TODO remove hardcoded version, maybe a release tag?
+const ANGULAR_LATEST_VERSION = '1.3.49';
+const SCRIPT_ID = 'builder-wc-script';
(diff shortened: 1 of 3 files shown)


packages/angular/src/app/modules/builder/utils/constants.ts
```diff
@@ -0,0 +1,3 @@
+// TODO remove hardcoded version, maybe a release tag?
+const ANGULAR_LATEST_VERSION = '1.3.49';
+const SCRIPT_ID = 'builder-wc-script';
```
samijaber, line 3: don't we need to export these and import them in the `builder-component.component.ts` file?

murtaza-builder: yes indeed. Somehow my editor didnt complain at that time.

fixed it.

shyam-builder (approved):
I'm guessing this can be merged only after 
* We release the official 1.3.49 version of WC

I'm assuming that the logic we need for custom breakpoints for Angular will be a follow-up PR? So we do the Angular SDK release after that is done correct?

PR #1801, opened 2023-02-08
Upgrade to use latest react sdk, with custom breakpoints support

murtaza-builder:
This is required for angular sdk that use this webcomponents script, to get custom breakpoints feature support as well

### DIFF
--- a/packages/webcomponents/package.json
+++ b/packages/webcomponents/package.json
@@ -143,7 +143,7 @@
   },
   "//": "TODO: make all of these optional dependencies, simple mode doesn't require",
   "dependencies": {
-    "@builder.io/react": "^2.0.4",
+    "@builder.io/react": "^2.0.13",
     "@builder.io/sdk": "^1.1.29",
     "@builder.io/shopify": "^1.3.42-3",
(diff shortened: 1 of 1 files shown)


mrkoreye (approved):

shyam-builder (commented):
Is there any changes needed here? If not, lets do a non-dev release, and check it in with the version updated as well please

packages/webcomponents/package.json
```diff
@@ -143,7 +143,7 @@
   },
   "//": "TODO: make all of these optional dependencies, simple mode doesn't require",
   "dependencies": {
-    "@builder.io/react": "^2.0.4",
+    "@builder.io/react": "^2.0.13",
     "@builder.io/sdk": "^1.1.29",
```
shyam-builder, line 147:
```suggestion
    "@builder.io/sdk": "^1.1.30",
```
should we bump this up as well?

murtaza-builder: yep, will do. react sdk also uses 1.1.30 now.

murtaza-builder: done.

PR #1827, opened 2023-02-14
Use latest core for angular sdk & version upgrade after release

murtaza-builder:


shyam-builder (approved):

samijaber (commented):
Note: I see both `package-lock.json` and `yarn.lock` files. This is dangerous and can cause inconsistent builds: we should only use one of `npm`/`yarn`.

PR #1834, opened 2023-02-16
[packages/utils] add utility method to set alt prop on pixels

teleaziz:
## setPixelProperties

`setPixelProperties` allows you to add more pixel element properties, for example , `alt`
### Example

```ts
import { setPixelProperties } from '@builder.io/utils';

export async function getServerSideProps({
  params,
}) {
  const page =
    (await builder
      .get(mode, ....)
      .toPromise()) || null
     
  setPixelProperties(page, { alt: 'pixel tag from builder' })
  
  return {
    props: {
      page,
    },
  }
}
```


packages/utils/src/set-pixel-alt.ts
```diff
@@ -0,0 +1,20 @@
+import { BuilderContent, BuilderElement } from '@builder.io/sdk';
+import traverse from 'traverse';
+
+const isBuilderPixel = (item: unknown): item is BuilderElement => {
+  return (item as any).id?.startsWith('builder-pixel');
+};
+
+export function setPixelAlt(content: BuilderContent, alt: string): BuilderContent {
```
mrkoreye, line 8: Can we make this more generic (or add another function) and accept an object with properties to set? For example, another value people want to set on pixels is `'aria-hidden': 'true',`

packages/utils/src/set-pixel-properties.ts
```diff
@@ -0,0 +1,24 @@
...
+  properties: Record<string, string>
+): BuilderContent {
+  return traverse(content).forEach(function (item) {
+    if (isBuilderPixel(item)) {
+      this.update({
+        ...item,
+        properties: {
+          ...item.properties,
+          ...properties,
+          src: item.properties!.src,
```
mrkoreye, line 19: Nice catch here overriding the src

mrkoreye (approved):

PR #1845, opened 2023-02-22
Next.js localized starter

ancheetah:
This example walks you through setting up i18n internationalization with Next.js, creating localized custom components, setting up custom targeting attributes for localization in Builder, and a video walkthrough of the starter and field level localization


samijaber (commented):
We usually strive to make our examples as simple as possible, with only the bare minimum needed.

Because of that, can we remove TypeScript and replace it with plain JS code everywhere?

mrkoreye:
@samijaberI don't entirely agree with removing Typescript here. Why would that make it simpler? Also we currently have many Nextjs examples using Typescript, for example: https://github.com/BuilderIO/builder/tree/main/examples/next-js-simple, https://github.com/BuilderIO/nextjs-shopify, https://github.com/BuilderIO/builder/tree/main/examples/next-js-builder-site

samijaber:
@mrkoreye Sorry, I should have linked the `CONTRIBUTING.md` doc with more context! It was put together by @steve8708 once he realized our examples involve technologies that are not strictly necessary for getting the code working: 

https://github.com/BuilderIO/builder/blob/main/examples/CONTRIBUTING.md#use-javascript-not-typescript

All of the examples you linked were made before we put together the contribution guidelines, and we actually want to actually hopefully update them to reflect these new policies.


The idea is to have the absolute bare minimum amount of files, code, tools and dependencies so that the examples are as straightforward as possible without the possibility of confusing newcomers that might not know certain tools.

steve8708:
Fwiw most of the sentiment online lately is that people overwhelmingly prefer typescript and nextjs even now defaults to typescript. When it comes to more complex examples I think typescript is ok and we can tweak the guidelines accordingly

ancheetah:
Hey all, so what I'm hearing is let's leave the typescript in this starter. I also think it helps emphasize in the custom Heading component that the localized value will resolve to a string and not return an object of localized values

Also, I spoke to @teleaziz about the Readme and I will be scraping it and replacing it with the next-js-simple readme with the addition of instructions for adding custom targeting attributes. The current information in the readme will be moved into a forum post.

teleaziz (approved):
LGTM, thanks AJ!

Issue #1679, opened 2023-01-16
getPersonalizedURL not compatible with NextJS v13


MikeBLambert:
**Describe the bug**
When running NextJS v13, `getPersonalizedURL` causes a `request.cookies.entries is not a function` error.
`getPersonalizedURL` relies on the `NextRequest`, which has changed in NextJS 13. Cookies now have a `name` and a `value` prop, and there is no longer `request.cookies.entries`. 
**To Reproduce**
1. On an application running NextJS v13, run middleware using `getPersonalizedURL` following instruction in the documentation.

**Expected behavior**
A personalized URL should be generated.

**Additional context**

Replacing

```
 const allCookies = Array.from(request.cookies.entries()).reduce(
    (acc, [key]) => ({
      ...acc,
      [key]: tryJsonParse(request.cookies.get(key)!),
    }),
    {}
  );
```

with
 
```
  const allCookies: Record<string, any> = {};
  for (const cookie of request.cookies) {
    const key = cookie[0];
    const value = request.cookies.get(key)?.value;
    if (value) {
      const parsedCookie = tryJsonParse(value);
      allCookies[key] = parsedCookie;
    }
  }
```

appears to solve the issue for us.

kylehotchkiss:
@Ross-s can you please release this change in an updated NPM package? 2.0.0 on NPM still refers to the .entries() syntax (https://www.npmjs.com/package/@builder.io/personalization-utils?activeTab=code)

Ross-s:
> @Ross-s can you please release this change in an updated NPM package? 2.0.0 on NPM still refers to the .entries() syntax (https://www.npmjs.com/package/@builder.io/personalization-utils?activeTab=code)

@teleaziz

PR #1847, opened 2023-02-23
changed cookie parsing method with suggestion from https://github.com…

Ross-s:
changed cookie parsing method with suggestion from https://github.com/BuilderIO/builder/issues/1679 and prevented undefined from being in final oputput if the sanitised version is undefined

## Description
As per https://github.com/BuilderIO/builder/issues/1679 the same issue was hapening for me, so i have changed the code to use the suggested method. I have also prevented undefined from being allowed in the outputed attributes as it is also possible for the sanitizedKey to be undefined


packages/personalization-utils/src/next.ts
```diff
@@ -22,13 +22,15 @@ export function getPersonalizedURL(
...
   const query = Object.fromEntries(url.searchParams);
-  const allCookies = Array.from(request.cookies.entries()).reduce(
-    (acc, [key]) => ({
-      ...acc,
-      [key]: tryJsonParse(request.cookies.get(key)!),
-    }),
-    {}
-  );
+  const allCookies: Record<string, any> = {};
+    for (const cookie of request.cookies) {
```
teleaziz, line 26: indentation, you can run from root:
```
yarn prettier -- write ./packages/personalization-utils/src/*.ts
```

teleaziz (commented):
LGTM, thanks for the pr, left one comment

Ross-s:
I have cleaned up the formating @teleaziz

teleaziz (approved):

Ross-s:
strange, my pr has nothing to do with styling, but a styling unit test failed?

samijaber:
@Ross-s These unit tests had some flakiness issues that were resolved last week. I merged the latest `main` back into your PR to address the issue.

teleaziz:
changed released in `@builder.io/personalization-utils@2.0.1-0` thanks again @Ross-s , I'll make a major version release next as this will break for users on older next versions

PR #1849, opened 2023-02-24
Removed yarn lock in favor of package-lock already being used

murtaza-builder:
## Description

Package.json and angular packager relies on package-lock.json. having yarn.lock simulatenously could be trouble some. Thus removing yarn-lock.


samijaber (approved):
🙏🏽 thanks!

Issue #1387, opened 2022-11-16
[TypeScript] BuilderContent cannot be JSON-serialized


jamesarosen:
**Describe the bug**
The result of `builder.get(…).promise()` can be serialized to JSON, but when deserialized, the result is no longer a `BuilderContent` and cannot be used with `<BuilderComponent content={…} />` according to TypeScript (though it does work).

The TypeScript error from the latest version of [Remix](https://github.com/remix-run/remix), which uses the serialization and deserialization to improve page-load performance:
```
Type 'SerializeObject<UndefinedToOptional<Article>>' is not assignable to type 'BuilderContent'.
  Types of property 'variations' are incompatible.
    Type 'SerializeObject<UndefinedToOptional<{ [id: string]: BuilderContentVariation | undefined; }>> | undefined' is not assignable to type '{ [id: string]: BuilderContentVariation | undefined; } | undefined'.
      Type 'SerializeObject<UndefinedToOptional<{ [id: string]: BuilderContentVariation | undefined; }>>' is not assignable to type '{ [id: string]: BuilderContentVariation | undefined; }'.
        'string' index signatures are incompatible.
          Type 'SerializeObject<UndefinedToOptional<BuilderContentVariation>>' is not assignable to type 'BuilderContentVariation'.
            Types of property 'data' are incompatible.
              Type 'SerializeObject<UndefinedToOptional<{ [key: string]: any; blocks?: BuilderElement[] | undefined; inputs?: Input[] | undefined; state?: { [key: string]: any; } | undefined; }>> | undefined' is not assignable to type '{ [key: string]: any; blocks?: BuilderElement[] | undefined; inputs?: Input[] | undefined; state?: { [key: string]: any; } | undefined; } | undefined'.
                Type 'SerializeObject<UndefinedToOptional<{ [key: string]: any; blocks?: BuilderElement[] | undefined; inputs?: Input[] | undefined; state?: { [key: string]: any; } | undefined; }>>' is not assignable to type '{ [key: string]: any; blocks?: BuilderElement[] | undefined; inputs?: Input[] | undefined; state?: { [key: string]: any; } | undefined; }'.
                  Types of property 'blocks' are incompatible.
                    Type 'SerializeObject<UndefinedToOptional<BuilderElement>>[] | undefined' is not assignable to type 'BuilderElement[] | undefined'.
                      Type 'SerializeObject<UndefinedToOptional<BuilderElement>>[]' is not assignable to type 'BuilderElement[]'.
[...]

PR #1874, opened 2023-03-02
fix style declarations for remix

steve8708:
I noticed we have an ugly workaround in our docs and examples

The problem with the workaround is it completely omits `content.data` which is a bad idea - people need to use `data` for custom fields, so this is the actual fix I [mentioned previously](https://github.com/BuilderIO/builder/issues/1387#issuecomment-1397442797)


mrkoreye (approved):

steve8708:
@mrkoreye can you make sure we get this released sometime soon-ish (would be great to have someone verify too that it fixes the remix issue - I don't have a full local setup but it should be the fix)

PR #1875, opened 2023-03-02
feat(queryV3): ENG-3180 add optional apiVersion param while generating content url

sanyamkamat:
## Description

This PR aims to add `apiVersion` as an optional param which will helps our customers choose between v2 and v3 query API.

For now the default is `v2`

We will merge the PR after following things are tested/done

- [x] Test at least one SDK locally by passing `v3` as the apiVersion

### DIFF
--- a/packages/sdks/src/functions/get-content/generate-content-url.test.ts
+++ b/packages/sdks/src/functions/get-content/generate-content-url.test.ts
@@ -1,4 +1,5 @@
 import { generateContentUrl } from './generate-content-url';
+import type { GetContentOptions } from './types';
 
 const testKey = 'YJIGb4i01jvw0SRdL5Bt';
 const testModel = 'page';
@@ -63,9 +64,7 @@ describe('Generate Content URL', () => {
         model: testModel,
         query: { id: testId },
         options,
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore Need this ignore as TS itself was complaining about invalid type
-        apiVersion: 'v1',
+        apiVersion: 'v1' as GetContentOptions['apiVersion'],
       });
     }).toThrow(`Invalid apiVersion: expected 'v2' or 'v3', received 'v1'`);
   });
@@ -77,9 +76,7 @@ describe('Generate Content URL', () => {
         model: testModel,
         query: { id: testId },
         options,
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore Need this ignore as TS itself was complaining about invalid type
-        apiVersion: 'INVALID_API_VERSION',
+        apiVersion: 'INVALID_API_VERSION' as GetContentOptions['apiVersion'],
       });
     }).toThrow(
       `Invalid apiVersion: expected 'v2' or 'v3', received 'INVALID_API_VERSION'`
--- a/packages/sdks/src/functions/get-content/types.ts
+++ b/packages/sdks/src/functions/get-content/types.ts
@@ -45,7 +45,9 @@ export interface GetContentOptions {
   locale?: string;
 
   /**
-   * If provided, we will use that version of query API for fetching content.
+   * If provided, sets the Builder API version used to fetch content.
+   *
+   * Defaults to `v2`.
    */
   apiVersion?: 'v2' | 'v3';
 }


packages/sdks/src/functions/get-content/generate-content-url.ts
```diff
@@ -15,14 +15,19 @@ export const generateContentUrl = (options: GetContentOptions): URL => {
...
+    apiVersion = 'v2',
   } = options;
 
   if (!apiKey) {
     throw new Error('Missing API key');
   }
 
+  if (!['v1', 'v2', 'v3'].includes(apiVersion)) {
+    throw new Error('Invalid apiVersion value');
+  }
```
samijaber, line 27: Awesome that you did this! Good call, useful for users that don't rely on TypeScript.

However, since the `packages/sdks` does not support v1, I think we can remove `v1` from this list:

```suggestion
  if (!['v2', 'v3'].includes(apiVersion)) {
    throw new Error('Invalid apiVersion value');
  }
```

and add a test that throws for `v1`

packages/sdks/src/functions/get-content/generate-content-url.ts
```diff
@@ -15,14 +15,19 @@ export const generateContentUrl = (options: GetContentOptions): URL => {
...
     locale,
+    apiVersion = 'v2',
   } = options;
 
   if (!apiKey) {
     throw new Error('Missing API key');
   }
 
+  if (!['v1', 'v2', 'v3'].includes(apiVersion)) {
+    throw new Error('Invalid apiVersion value');
```
samijaber, line 26: I would also improve this message to be explicit about what is supported

```suggestion
    throw new Error(`Invalid apiVersion: expected 'v2' or 'v3', received '${apiVersion}'`);
```

shyam-builder (approved):
LGTM!

One worry I have is whether we might have different query params based on version and how we will handle it, but I'm ok with kicking that problem down the road for now.

packages/sdks/src/functions/get-content/generate-content-url.test.ts
```diff
@@ -32,4 +33,56 @@ describe('Generate Content URL', () => {
...
+  test('throw error when trying to generate content url with apiVersion as v1', () => {
+    expect(() => {
+      generateContentUrl({
+        apiKey: testKey,
+        model: testModel,
+        query: { id: testId },
+        options,
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore Need this ignore as TS itself was complaining about invalid type
+        apiVersion: 'v1',
```
samijaber, line 68: `@ts-ignore` is excessive because it will blindly ignore any TS error in the line below. A better approach is to type-cast the variable to what you want it to be:
```suggestion
        apiVersion: 'v1' as GetContentOptions['apiVersion'],
```

packages/sdks/src/functions/get-content/generate-content-url.test.ts
```diff
@@ -32,4 +33,56 @@ describe('Generate Content URL', () => {
...
+  test('throw error when trying to generate content url with an invalid apiVersion value', () => {
+    expect(() => {
+      generateContentUrl({
+        apiKey: testKey,
+        model: testModel,
+        query: { id: testId },
+        options,
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore Need this ignore as TS itself was complaining about invalid type
+        apiVersion: 'INVALID_API_VERSION',
```
samijaber, line 82:
```suggestion
        apiVersion: 'INVALID_API_VERSION' as GetContentOptions['apiVersion'],
```

packages/sdks/src/functions/get-content/types.ts
```diff
@@ -43,4 +43,9 @@ export interface GetContentOptions {
    * If provided, the API will auto-resolve localized objects to the value of this `locale` key.
    */
   locale?: string;
+
+  /**
+   * If provided, we will use that version of query API for fetching content.
```
samijaber, line 48:
```suggestion
   * If provided, sets the Builder API version used to fetch content.
   * Defaults to `v2`.
```

samijaber (commented):
Nice & easy!

samijaber:
@sanyamkamat I blanked on the fact that we still need to pass down `apiVersion` to the `getContent` call made in the Symbol block:

https://github.com/BuilderIO/builder/blob/f7f04f2b254ad460d360faca1d05fff29b4c7ad8/packages/sdks/src/blocks/symbol/symbol.lite.tsx#L72-L78

You're going to most likely need to pass `apiVersion` as a prop to `RenderContent`, and store it in the `BuilderContext`. Look at how we do that for the `apiKey` here: 

https://github.com/BuilderIO/builder/blob/f7f04f2b254ad460d360faca1d05fff29b4c7ad8/packages/sdks/src/context/types.ts#L20

It should be done very similarly. Let me know if you have any questions

PR #1883, opened 2023-03-06
Fixes: eval issue in shopify-hydrogen using remix

akshatgarg12:
## Description

fixes https://github.com/BuilderIO/builder-shopify-hydrogen/issues/12
Removed use of eval from safeDynamicRequire and directly assign require to safeDynamicRequire in case require is defined.

_Screenshot_
Not a visual change


akshatgarg12:
https://www.loom.com/share/218a1e19934e4689bc86995d998260ab

packages/react/src/functions/safe-dynamic-require.ts
```diff
@@ -32,7 +32,11 @@ if (
   typeof globalThis?.require === 'function' ||
   (Builder.isServer && typeof require === 'function')
 ) {
-  safeDynamicRequire = eval('require');
+  if (globalThis?.require) {
+    safeDynamicRequire = globalThis.require;
+  } else {
+    safeDynamicRequire = require;
```
teleaziz, line 38: The reason for the `eval` was to make sure requiring `vm2` doesn't include it in the webpack bundle , more info here https://github.com/webpack/webpack/issues/4175

akshatgarg12: Would it be better to move this into a `try-catch block` with eval method in try block and fallback to this method if a particular environment does not allow the use of eval?

teleaziz: This will continue to throw a build error, as `eval` would still be in the bundle, we need to somehow replace the `eval('require')` only for server builds,

maybe add a `replace` for `esm` and `cjs` bundles, in this line https://github.com/BuilderIO/builder/blob/main/packages/react/rollup.config.ts#L94
```
  plugins: options.plugins
      .filter(plugin => plugin !== resolvePlugin)
      .concat([
        resolve({
          only: [/^\.{0,2}\//],
        }),
        replace({
          "eval('require')": "require",
        })
      ]),

```

akshatgarg12: This is not working for me, but this does : 
```js
try{
    safeDynamicRequire = eval('require');
  }catch(e){
    if (globalThis?.require) {
      safeDynamicRequire = globalThis.require;
    } else {
      safeDynamicRequire = require;
    }
  }
```

teleaziz: Worth trying, I'd do :
```
try{
    safeDynamicRequire = eval('require');
  }catch(e){
    if (globalThis?.require) {
      safeDynamicRequire = globalThis.require;
    } else {
      safeDynamicRequire = noop;
    }
  }
 ```

akshatgarg12: changes made!

mrkoreye (approved):

terza98:
Can you guys merge this?

teleaziz:
Fix is released on @builder.io/react@2.0.16:

PR #1887, opened 2023-03-07
add new react js example with vite

sanyamkamat:
## Description

Adding a working example of react-js app with Builder components.

_Screenshot_
image

### DIFF
--- a/examples/react-js/README.md
+++ b/examples/react-js/README.md
@@ -1,20 +1,138 @@
-## Builder.io example with React
+# Builder.io example with React
 
-### Quick start
+This example walks you through using Builder.io with a minimal React application.
 
-[Open this example in Codesandbox](https://codesandbox.io/s/github/BuilderIO/builder/tree/main/examples/react-js)
+## Prerequisites
 
-<a target="_blank" href="https://codesandbox.io/s/github/BuilderIO/builder/tree/main/examples/react-js">
-  <img width="597" height="375" src="https://i.imgur.com/zue72Q0.jpg">
-</a>
+Before using this example, make sure you have the following:
 
-### To run the example Locally
+- A [Builder.io](builder.io) account. Check out the [plans](https://www.builder.io/m/pricing), which range from our free tier to custom.
+- [npm](https://www.npmjs.com/)
 
-```bash
-git clone https://github.com/BuilderIO/builder.git
-cd examples/react-js
-npm install
-npm run dev
-```
+A basic understanding of the following is helpful too:
 
-See src/main.jsx for the source
+- The [command line](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Understanding_client-side_tools/Command_line)
+- JavaScript and [React](https://reactjs.org/)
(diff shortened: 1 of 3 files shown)


samijaber (commented):
I think the `package-lock.json` is missing here

examples/react-js/README.md
```diff
@@ -0,0 +1,20 @@
...
+### To run the example Locally
+
+```bash
+git clone https://github.com/BuilderIO/builder.git
+cd examples/react-js
+npm install
+npm run dev
+```
+
+See src/main.jsx for the source
```
samijaber, line 20: We need better instructions than that for new examples. I know you copied the one for the react one, but this is an old one and we want more details for our users.

Some examples:
- https://github.com/BuilderIO/builder/blob/main/examples/svelte/sveltekit/README.md
- https://github.com/BuilderIO/builder/blob/main/examples/next-js-simple/README.md


You can probably copy/paste a bunch of the stuff from the READMEs and replace command names.

examples/react-js/index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Builder.io Plain JS example</title>
```
samijaber, line 4:
```suggestion
    <title>Builder.io React JS example</title>
```

samijaber (commented):
Thanks a lot for working on this @sanyamkamat . Only thing missing is a slightly improved README and then this is ready to go 🚀

examples/react-js/src/main.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+class CatchallPage extends React.Component {
+  state = { notFound: false };
+
+  render() {
+    return !this.state.notFound ? (
+      <BuilderComponent
+        apiKey="bb209db71e62412dbe0114bdae18fd15"
+        model="page"
+        contentLoaded={content => {
```
teleaziz, line 66: Please replace the old react example, and fetch the content manually here, same as our quick start example on https://www.builder.io/c/docs/quickstart
```jsx
import { BuilderComponent, builder, useIsPreviewing } from '@builder.io/react';
import { useEffect, useState } from "react";

// Put your API key here
builder.init(YOUR_API_KEY)

// set whether you're using the Visual Editor,
// whether there are changes,
// and render the content if found
export default function CatchAllRoute() {
  const isPreviewingInBuilder = useIsPreviewing();
  const [notFound, setNotFound] = useState(false);
  const [content, setContent] = useState(null);

  // get the page content from Builder
  useEffect( () => {
    async function fetchContent() {
      const content = await builder.get('page', {
        url: window.location.pathname
      }).promise();

      setContent(content);
      setNotFound(!content);
    }
    fetchContent();
  }, []);
  // if no page is found, return a 404 page
  if (notFound && !isPreviewingInBuilder) {
    return <div>Not Found 404 Error</div>
  }

  // return the page when found
  return (
    <>
      {/* Render the Builder page */}
      <BuilderComponent model="page" content={content} />
    </>
  );
}

```

samijaber (approved):

PR #1893, opened 2023-03-13
ENG-3180 add support of apiVersion in Mitosis generated SDKs

sanyamkamat:
## Description

This PR makes sure the Symbols respect the apiVersion passed to it via `RenderContent` block.

_Loom_
https://www.loom.com/share/88feb0f77a254d3fb5310bec3fef3f74

### DIFF
--- a/packages/sdks/src/context/builder.context.lite.ts
+++ b/packages/sdks/src/context/builder.context.lite.ts
@@ -7,7 +7,7 @@ export default createContext<BuilderContextInterface>({
   state: {},
   setState: () => {},
   apiKey: null,
-  apiVersion: 'v2',
+  apiVersion: undefined,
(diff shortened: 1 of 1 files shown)


packages/sdks/src/context/builder.context.lite.ts
```diff
@@ -7,6 +7,7 @@ export default createContext<BuilderContextInterface>({
   state: {},
   setState: () => {},
   apiKey: null,
+  apiVersion: 'v2',
```
samijaber, line 10: This can actually be `undefined` too. Since `getContent` accepts undefined values, it's better to only have one default there instead of another one here.

```suggestion
  apiVersion: undefined,
```

samijaber (approved):
👍🏽

PR #1907, opened 2023-03-15
fix(react): fix accessing node-fetch on server and continue remix typ…

teleaziz:
…e fixes, safe access to process for remix


samijaber (approved):
looks good. You can easily add a Remix e2e test in this folder: https://github.com/BuilderIO/builder/tree/main/packages/react-tests

and import the React SDK, and we will be testing Remix moving forward

mrkoreye (approved):

PR #1933, opened 2023-03-21
ENG-3473 add support of apiVersion in Core React SDK

sanyamkamat:
## Description

PR to add support for apiVersion in Core React SDK

https://www.loom.com/share/aeddb6df721d4a2f9cc708077389fb81

### DIFF
--- a/packages/sdks/package.json
+++ b/packages/sdks/package.json
@@ -80,7 +80,7 @@
     "pinst": "^3.0.0",
     "prettier": "^2.5.0",
     "ts-jest": "^27.1.4",
-    "typescript": "4.1.5",
+    "typescript": "^4",
     "watch": "^1.0.2"
   },
   "config": {
--- a/packages/yarn.lock
+++ b/packages/yarn.lock
@@ -2384,7 +2384,7 @@ __metadata:
     seedrandom: ^3.0.5
     traverse: ^0.6.6
     ts-jest: ^27.1.4
-    typescript: 4.1.5
+    typescript: ^4
     watch: ^1.0.2
   languageName: unknown
   linkType: soft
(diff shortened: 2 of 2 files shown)


samijaber (changes requested):
I think I have a good way to build an e2e test for the `apiVersion` for all SDKs, so we can make sure it works according to plan. I’ll share a loom later today explaining how we can do that. We should include it in that PR to make sure all the SDKs work as expected

samijaber:
https://www.loom.com/share/827e2359f08a401782f52470f1b150ee

packages/sdks-tests/src/tests/blocks.spec.ts
```diff
@@ -287,4 +289,120 @@ test.describe('Blocks', () => {
...
+          return route.fulfill({
+            status: 200,
+            body: x === 0 ? FIRST_SYMBOL_CONTENT : SECOND_SYMBOL_CONTENT,
+          });
+        },
+        { times: 2 }
+      );
+
+      await testSymbols(page);
+    });
```
samijaber, line 349: for the `v3` test, we don't need 2 separate tests for oldReact and the new SDKs. You can just run `test()` normally and have it work on all of them at once

```suggestion
```

packages/sdks-tests/src/tests/blocks.spec.ts
```diff
@@ -287,4 +289,120 @@ test.describe('Blocks', () => {
...
+
+    testExcludeOldReact('apiVersion in new SDKs is set to v3', async ({ page }) => {
```
samijaber, line 389: You can delete the other test and have this one run on all using `test()`

```suggestion
    test('apiVersion in SDKs is set to v3', async ({ page }) => {
```

packages/sdks-tests/src/tests/blocks.spec.ts
```diff
@@ -287,4 +289,120 @@ test.describe('Blocks', () => {
...
+
+      await testSymbols(page);
+    });
+
+    testExcludeOldReact('apiVersion in new SDKs is not set', async ({ page }) => {
+      await page.goto('/api-version-default');
+
+      let x = 0;
+      await page.route(
+        /.*cdn\.builder\.io\/api\/v2\/content\/symbol.*/,
```
samijaber, line 356: Instead of having 2 separate tests for the default version, we can toggle this URL based on which SDK is being tested. Something like

```suggestion
       SDK === 'oldReact' ?  /.*cdn\.builder\.io\/api\/v1\/content\/symbol.*/ : /.*cdn\.builder\.io\/api\/v2\/content\/symbol.*/,
```

samijaber (commented):
Sweet! A few comments on combining some of the tests together

shyam-builder (approved):

samijaber (approved):
🚀

PR #1934, opened 2023-03-21
Update README note on SDK perf size

samijaber:
## Description

Fix perf section about React SDK, make it clear that there is an old and a new React SDK

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,8 @@ Some example SDK sizes from Bundlephobia:
 
 |SDK|Size|
 |----|----|
-|React|![React Bundle Size](https://img.shields.io/bundlephobia/minzip/@builder.io/sdk-react)|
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -169,9 +169,12 @@ Some example SDK sizes from Bundlephobia:
 
 |SDK|Size|
 |----|----|
-|React|![React Bundle Size](https://img.shields.io/bundlephobia/minzip/@builder.io/sdk-react)|
+|React|![React Bundle Size](https://img.shields.io/bundlephobia/minzip/@builder.io/react)|
```
shyam-builder, line 172: should we put the package name along with the SDK? so that it is absolutely clear?

samijaber: Good call, done

PR #2032, opened 2023-04-11
use this.apiVersion on server for getAll

steve8708:
I believe this is needed - correct me of I'm wrong @sanyamkamat


samijaber (approved):
Makes sense. Our e2e tests are only able to test `apiVersion` via loading symbol data client-side, as it's hard to test SSR API calls. Thanks for catching this one

PR #2096, opened 2023-04-13
Fix: CI

samijaber:
## Description

webcomponents project is failing on builds after dependabot merges
- update `/packages/shopify` to stop including TS files in npm package (causing TS issues in webcomponents)
- revert https://github.com/BuilderIO/builder/pull/1555

### DIFF
--- a/packages/webcomponents/src/elements.ts
+++ b/packages/webcomponents/src/elements.ts
@@ -538,8 +538,7 @@ if (Builder.isBrowser && !customElements.get(componentName)) {
       const getReactPromise = importReact(); // TODO: only import what needed based on what comes back
       const getWidgetsPromise = importWidgets();
       const getShopifyPromise = isShopify ? importShopify() : null;
-      // TO-DO: double-check return type of this promise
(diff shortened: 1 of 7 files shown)


mrkoreye (approved):

packages/webcomponents/src/elements.ts
```diff
@@ -1,7 +1,7 @@
 import { Builder, builder } from '@builder.io/react';
 const importReact = () => import('@builder.io/react');
-const importShopify = () => import('@builder.io/shopify/react');
-const importShopifyJs = () => import('@builder.io/shopify/js');
+const importShopify = () => import('@builder.io/shopify/dist/react/bundle.esm.js');
+const importShopifyJs = () => import('@builder.io/shopify/dist/js');
```
samijaber, line 4: That's what the original non-dist files were re-exporting, so this should be OK:

https://github.com/BuilderIO/builder/blob/5019f07c2cc718de230713daa0e5986771220104/packages/shopify/react/index.js#L1

https://github.com/BuilderIO/builder/blob/5019f07c2cc718de230713daa0e5986771220104/packages/shopify/js.js#L1

PR #2122, opened 2023-04-18
Feature: SDKs: A/B testing + SSR

samijaber:
## Description

SDKs:
- Add support for SSR'd A/B testing 
  - add `RenderContentVariants` wrapper around `RenderContent` that renders all variations side-by-side when needed (`canTrack: true` + variations exist + is in SSR)
  - we now export `RenderContentVariants` instead of `RenderContent` to the end-user (but keep `RenderContent` as the export's name). It has the same props, so there is no breaking change or compatibility issue.
    - NOTE: this excludes React Native, which has no notion of SSR (even though `react-native-web` could technically be used to SSR, `react-native-web`'s rendering ignores `<script>` and `<style>` tags, which is problematic as those are needed for the SSR'd A/B tests to work correctly)
  - explore `processContentResult` for internal purposes (mainly having e2e tests correctly process mocked data)


packages/sdks/e2e/svelte/vite.config.ts
```diff
@@ -3,11 +3,5 @@ import { svelte } from '@sveltejs/vite-plugin-svelte';
 
 // https://vitejs.dev/config/
 export default defineConfig({
-  plugins: [
-    svelte({
-      // this seems to be needed for the SDK's CSS to be passed down properly
-      // not actually sure what this flag does internally.
-      emitCss: false,
```
samijaber, line 10: TODO: is this no longer needed?

PR #2205, opened 2023-04-25
support enrich flag

sanyamkamat:
## Description

This PR adds support for `enrich` flag in the SDKs

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -274,8 +274,8 @@ export interface UserAttributes {
 
 type AllowEnrich =
   | { apiVersion?: Extract<ApiVersion, 'v1'> }
-  | { apiVersion?: Extract<ApiVersion, 'v3'>; enrich?: boolean; }
-  | { apiVersion?: never; enrich?: boolean; };
+  | { apiVersion?: Extract<ApiVersion, 'v3'>; enrich?: boolean }
+  | { apiVersion?: never; enrich?: boolean };
 
 export type GetContentOptions = AllowEnrich & {
   /**
@@ -454,7 +454,7 @@ export type GetContentOptions = AllowEnrich & {
    * content thinking they should updates when they actually shouldn't.
    */
   noEditorUpdates?: boolean;
-}
+};
 
 export type Class = {
   name?: string;


samijaber (changes requested):
Requesting changes as per the conversation we had in Slack around deprecating the old params as well. 🙏🏽

samijaber:
@sanyamkamat Thanks. I just checked the conversation we had internally and it sounds like maybe we want to:

- `v3` to only allow `enrich`
- `v1` and `v2` to only allow


So we need to update our types to look something like this:

```ts
type GetContentOptions = 
  | { apiVersion: 'v3'; enrich?: boolean }
  | { apiVersion: 'v2'; noTraverse?: boolean; includeRefs?: boolean }
```


So that it's impossible for the user to accidentally use a wrong property.

samijaber:
> * We want users to smoothly move from `v1` to `v3`. By making strict TS checks which allow only `enrich` and not `includeRefs` or `noTraverse` could break their existing build steps where they are passing the now deprecated options.

The TS checks I'm describing should not break any existing build: I only want to forbid impossible scenarios that will confuse our users and cause a lot of questions ("what is the difference between `enrich`, `noTraverse` and `includeRefs`? Why does `noTraverse` do nothing when I set it along with `enrich`?")

> * `includeRefs`, `noTraverse`, and `enrich` have been part of previous content API for more than a year.
> What this PR aims to do is let users pass `enrich` field in the options of BuilderComponent regardless of apiVersion being `v1` or `v3`

What I'm saying is that we should _not_ let user pass `enrich` if they are on `v1`, because `v1` does not support the `enrich` flag.


I'm OK with still allowing `includeRefs` and `noTraverse` in `v3` and marking them as deprecated for now, but I think `enrich` should not be allowed for the React SDK in `v1` as it isn't supported there.

packages/core/src/builder.class.ts
```diff
@@ -298,6 +303,7 @@ export interface GetContentOptions {
   /**
    * Follow references. If you use the `reference` field to pull in other content without this
    * enabled we will not fetch that content for the final response.
+   * @deprecated use `enrich` instead
```
samijaber, line 306: I think we also need a similar `@deprecated` tag for `noTraverse`.

packages/sdks/src/functions/get-content/generate-content-url.ts
```diff
@@ -15,6 +15,7 @@ export const generateContentUrl = (options: GetContentOptions): URL => {
     model,
     apiKey,
     includeRefs = true,
+    enrich,
```
samijaber, line 18: Two things:

- I can see here that `generateContentUrl` sets default values for `noTraverse` and `includeRefs`. We end up providing those two along with `enrich`. Do we still want to do that? Would that cause any issues? Or do we want to not provide those defaults if `enrich` is provided for us?
- shouldn't `enrich` have a default of `true`, since `includeRefs` has a default of `true`?

packages/core/src/builder.class.ts
samijaber: Oh, didn't realize this was only in the new SDK. Gotcha 👍🏽

samijaber (approved):
👍🏽

PR #2219, opened 2023-05-06
fix(sdk): properly write to the root state (instead of repeated state)

mhevery:
When adding an action to a repeated item, the action code assignment would write to the local state. This is incorrect as local state is pseudo-state created for the purposes of the repeater. The correct behavior is to write to the actual state.

This is achieved by having a Proxy that intercepts the write and correctly writes to the root state.

## Description

_Screenshot_


samijaber:
@mhevery I think the react native test is failing due to this `mitosis.config.js` plugin missing the new properties you added:

https://github.com/BuilderIO/builder/blob/7fd5be5a0cfc806a90225124dfbae7c8305d2e6a/packages/sdks/mitosis.config.js#L405-L437

samijaber (approved):

PR #2220, opened 2023-05-08
make v1/v2 as the default apiVersion for next minor release

sanyamkamat:
## Description

The last upgrade of minor version in our SDKs made `v3` as the default apiVersion. After that change there are reports of issues from customers complaining of things not working correctly(eg Zapier)

This PR will make sure that current Major versions will keep `v1`/`v2` as the default apiVersion. While the next major version will make `v3` as the default version. We will also update the documentation respectively.


packages/sdks/src/types/api-version.ts
```diff
@@ -1,2 +1,2 @@
 export type ApiVersion = 'v2' | 'v3';
-export const DEFAULT_API_VERSION: ApiVersion = 'v3';
+export const DEFAULT_API_VERSION: ApiVersion = 'v2';
```
samijaber, line 2: The new SDKs are still on `0.x` versions and have not reached `1.0`. For them, we would want to stick with `v3` as a default because we don't want to do a major version bump yet (they are still beta SDKs and we are not ready to bump to `v1.0`)

samijaber (changes requested):
Looks good for the React SDK, but I don't think we want to make that change for the new SDKs

samijaber (approved):

PR #2238, opened 2023-05-10
Update README.md

steve8708:
to include builder playground


steve8708:
@samijaber FYI I think these SDK tests are failing on main

PR #2290, opened 2023-05-17
release next minor version with v1 as default

sanyamkamat:
## Description

This PR commits the files which were modified when I released the next **minor** versions of core and react SDK with `v1` as the default `apiVersion`.


samijaber (approved):
Perfect 👍

PR #2292, opened 2023-05-19
release next major versions of core and react SDK with v3 as default

sanyamkamat:
## Description

This PR commits the files which were modified when I released the next major versions of core and react SDK with `v3` as the default apiVersion.

### DIFF
--- a/packages/react/CHANGELOG.md
+++ b/packages/react/CHANGELOG.md
@@ -1,6 +1,12 @@
 ### 3.0.0
 - `apiVersion` property now defaults to `v3`.
 
+### 2.2.0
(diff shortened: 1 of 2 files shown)


packages/react/CHANGELOG.md
```diff
@@ -1,8 +1,5 @@
-### 2.2.0
-- Sets the default `apiVersion` to `v1`.
-
-### 2.1.0
```
samijaber, line 4: I don't think it makes sense to hide old information. It's better to keep it in case someone's using that version, so they aren't confused about what's going on

PR #2327, opened 2023-06-06
Fix: SDKs tests

samijaber:
## Description

flaky flaky race condition


packages/sdks-tests/src/tests/ab-test.spec.ts
```diff
@@ -58,7 +58,7 @@ test.describe.configure({ retries: 0 });
 test.describe('A/B tests', () => {
   const TRIES = 10;
   // loop 10 times to check for flakiness
-  Array.from({ length: TRIES }).forEach((_, i) => {
```
samijaber, line 61: Can't have async code inside of `forEach`. Technically the async code is an argument to `test()`, so maybe this doesn't help me, but worth trying anyways

mrkoreye (approved):

PR #2359, opened 2023-06-27
ENG-3807 remove process keyword to unblock remix hydrogen users

sanyamkamat:
## Description

There is a customer who was saying that react SDK fails for them. They are using remix/hydrogen setup which is against using `process` keyword. We released different dev versions with `process?.env?.DEBUG` and `process && process.env && process.env.DEBUG` and they both didnt work for them. Finally when we removed the lines, they reported it worked for that dev release.
So this PR will make sure it is added in our next patch releases, so that the same customer can use the next versions of our SDK

Tagging @teleaziz as he added the line.

_JIRA ticket_
https://builder-io.atlassian.net/browse/ENG-3807

### DIFF
--- a/examples/astro-solidjs/package-lock.json
+++ b/examples/astro-solidjs/package-lock.json
@@ -8,7 +8,7 @@
       "name": "@example/basics",
       "version": "0.0.1",
       "dependencies": {
-        "@builder.io/sdk-solid": "^0.4.0",
+        "@builder.io/sdk-solid": "^0.4.4",
         "solid-styled-components": "^0.27.6"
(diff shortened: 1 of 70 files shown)


teleaziz (approved):

samijaber (approved):

samijaber (commented):
@sanyamkamat You missed one more btw: https://github.com/BuilderIO/builder/blob/4352bd8ad1f12a4efc3ed57d1acd340fff1a80b6/packages/react/src/functions/try-eval.tsx#L73

PR #2363, opened 2023-06-29
Fix: SDKs a/b test SSR hydration

samijaber:
## Description

SDKs:
- Vue: stop asyncly importing components (except those absolutely necessary to avoid circular dependencies). This was causing the A/B test inline styles not to render/update properly.
- A/B testing: fix hydration target check

### DIFF
--- a/packages/sdks/src/components/render-content-variants/render-content-variants.lite.tsx
+++ b/packages/sdks/src/components/render-content-variants/render-content-variants.lite.tsx
@@ -75,7 +75,7 @@ export default function RenderContentVariants(props: VariantsProviderProps) {
               customComponents={props.customComponents}
               hideContent
               parentContentId={props.content?.id}
-              isSsrAbTest={state.shouldRenderVariants}
+              isSsrAbTest
             />
           )}
         </For>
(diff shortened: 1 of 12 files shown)


packages/sdks/src/components/render-content-variants/render-content-variants.lite.tsx
```diff
@@ -89,7 +89,7 @@ export default function RenderContentVariants(props: VariantsProviderProps) {
         customComponents={props.customComponents}
         classNameProp={`variant-${props.content?.id}`}
         parentContentId={props.content?.id}
-        isSsrAbTest={state.shouldRenderVariants}
+        isSsrAbTest={false}
```
mrkoreye, line 92: Is this change intentional or leftover from testing something?

mrkoreye (approved):

packages/sdks/src/components/render-content-variants/render-content-variants.lite.tsx
samijaber: It's a mistake which is causing the build to fail! Thanks for pointing this out to me 🙏🏽

PR #2396, opened 2023-07-12
Adding preload html attribute to video component

anaghav2023:
## Description
Refer to this Jira Ticket to understand the issue: https://builder-io.atlassian.net/browse/SUPP-24

Or this loom: https://www.loom.com/share/1d3b0696a6d84c3cac2d7273641d7e20

**Loom**
Watch this loom to see the sdk changes working
https://www.loom.com/share/27f904ae6c18410bbe3c90aaaed523bb

### DIFF
--- a/packages/sdks/src/blocks/video/component-info.ts
+++ b/packages/sdks/src/blocks/video/component-info.ts
@@ -60,7 +60,7 @@ export const componentInfo: ComponentInfo = {
     {
       name: 'preload',
       type: 'text',
-      defaultValue: 'auto',
+      defaultValue: 'metadata',
       enum: ['auto', 'metadata', 'none'],
     },
     {
--- a/packages/sdks/src/blocks/video/component-info.ts
+++ b/packages/sdks/src/blocks/video/component-info.ts
@@ -58,6 +58,12 @@ export const componentInfo: ComponentInfo = {
       enum: ['contain', 'cover', 'fill', 'auto'],
     },
     {
+      name: 'preload',
+      type: 'text',
+      defaultValue: 'auto',
+      enum: ['auto', 'metadata', 'none'],
+    },
+    {
       name: 'fitContent',
       type: 'boolean',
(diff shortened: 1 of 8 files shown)


packages/sdks/src/blocks/video/video.lite.tsx
```diff
@@ -47,6 +48,7 @@ export default function Video(props: VideoProps) {
   return (
     <video
       {...state.spreadProps}
+      preload={props.preload}
```
teleaziz, line 51:
```suggestion
      preload={props.preload || 'metadata'}
```

Default values on inputs do not apply retroactively, meaning that content that is already published will have this as undefined

packages/sdks/src/blocks/video/component-info.ts
```diff
@@ -57,6 +57,12 @@ export const componentInfo: ComponentInfo = {
       defaultValue: 'cover',
       enum: ['contain', 'cover', 'fill', 'auto'],
     },
+    {
+      name: 'preload',
+      type: 'text',
+      defaultValue: 'auto',
```
teleaziz, line 63:
```suggestion
      defaultValue: 'metadata',
```

Wdyt if we match the spec, default value in browsers is `metadata` https://developer.mozilla.org/en-US/docs/Learn/Performance/video#video_preload

teleaziz (changes requested):

packages/sdks/src/blocks/video/video.lite.tsx
anaghav2023: Oh right ! Thanks. So Ill also add it to the Video.tsx file in react sdk

packages/sdks/src/blocks/video/component-info.ts
anaghav2023: okay cool

teleaziz (approved):

PR #2402, opened 2023-07-14
Fix: SDK symbols SSR A/B tests

samijaber:
## Description

- add SSR'd A/B test to Symbols by reusing `RenderContentVariants`.
- add missing props not being passed down from `RenderContentVariants` to `RenderContent`
- fix duplicate global A/B test scripts when multiple Builder components are rendered on the same page.
- add `testVariationId` back to all content variants for consistency.
- fix a small issue with `Text` block rendering non-strings


packages/sdks/src/blocks/text/text.lite.tsx
```diff
@@ -1,8 +1,8 @@
-export default function Text(props: { text: string }) {
+export default function Text(props: { text?: string }) {
   return (
     <span
       class="builder-text"
-      innerHTML={props.text}
+      innerHTML={props.text?.toString()}
```
mrkoreye, line 5: if `innerHtml` always has to be a string this could be changed to
```suggestion
      innerHTML={String(props.text || '')}
```

packages/sdks/src/components/render-content/render-content.lite.tsx
```diff
@@ -435,7 +436,7 @@ export default function RenderContent(props: RenderContentProps) {
               },
             }
           : {})}
-        {...(props.hideContent ? { hidden: true, 'aria-hidden': true } : {})}
+        {...(props.showContent ? {} : { hidden: true, 'aria-hidden': true })}
```
mrkoreye, line 439: 😎

mrkoreye (approved):

packages/sdks/src/blocks/text/text.lite.tsx
samijaber: Thanks for the tip! Will change it in my next SDK PR

PR #2473, opened 2023-07-19
fix: 🐛 incorrect handling of computed property bindings

mhevery:
Computed properties/attributes were not applied to the rendering (only static values were applied)

BREAKING CHANGE: 🧨 no

## Description

_Screenshot_


samijaber (approved):

samijaber (changes requested):
(undoing my approval after I noticed tests failing). 

In fact, I'm surprised that even the gen1 react SDK fails on this test. Any ideas why? https://github.com/BuilderIO/builder/actions/runs/5602976068/jobs/10248953144?pr=2473

PR #2479, opened 2023-07-20
Added fix for translating data models issue

gustavocodess:
Added fix for translating data models issue, on both functions getTranslatableFields and applyTranslation

### DIFF
--- a/examples/next-app-directory/package-lock.json
+++ b/examples/next-app-directory/package-lock.json
@@ -151,6 +151,41 @@
         "node": ">=6.9.0"
       }
     },
+    "node_modules/@builder.io/react": {
+      "version": "3.0.7",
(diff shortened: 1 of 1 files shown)


packages/utils/src/translation-helpers.test.ts
```diff
@@ -55,6 +55,24 @@ test('getTranslateableFields from content to match snapshot', async () => {
...
+      },
+    },
+  };
+  const result = getTranslateableFields(
+    content,
+    'en-US',
+    'Visit https://builder.io/fiddle/... for more details'
+  );
+  expect(result).toMatchSnapshot();
+});
```
teleaziz, line 74: also please add a test for `applyTranslation on plain data models`

PR #2485, opened 2023-07-20
Fix for translation helpers for data models

gustavocodess:
Added fix for translating data models issue, on both functions getTranslatableFields and applyTranslation

Added test for getTranslatableFields and applyTranslation for data models


teleaziz (approved):
Great catch, thanks Gustavo

PR #2486, opened 2023-07-21
[packages/react] replace vm2 with isolated vm

teleaziz:
also fix an issue with default values for content inputs are not set when SSR

### DIFF
--- a/packages/react/src/functions/string-to-function.ts
+++ b/packages/react/src/functions/string-to-function.ts
@@ -140,6 +140,7 @@ export function stringToFunction(
               ? new ivm.Reference(
                   index === 3
                     ? {
+                        // workaround: methods with default values for arguments is not being cloned over
                         ...arg,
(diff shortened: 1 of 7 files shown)


packages/react/src/functions/string-to-function.ts
```diff
@@ -118,21 +119,36 @@ export function stringToFunction(
...
+          }
+        });
+
+        const ivm = safeDynamicRequire('isolated-vm') as typeof import('isolated-vm');
+        return isolateContext.evalClosureSync(
+          makeFn(str, useReturn),
+          args.map((arg, index) =>
+            typeof arg === 'object'
+              ? new ivm.Reference(
+                  index === 3
```
mrkoreye, line 141: What is the significance of the `3` here? Could be worth a comment or define it as a constant with a descriptive name

mrkoreye (approved):

PR #2495, opened 2023-07-25
make sure nextjs doesn't aggressively cache fetches

steve8708:
currently, without setting `next` - nextjs will cache fetches and in production content will take forever to update

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -2263,7 +2263,10 @@ export class Builder {
     url: string,
     options?: { headers: { [header: string]: number | string | string[] | undefined } }
   ) {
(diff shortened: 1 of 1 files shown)


packages/core/src/builder.class.ts
```diff
@@ -2263,7 +2263,10 @@ export class Builder {
     url: string,
     options?: { headers: { [header: string]: number | string | string[] | undefined } }
   ) {
-    return getFetch()(url, options as SimplifiedFetchOptions).then(res => res.json());
+    return getFetch()(url, {
+      next: { revalidate: 1 },
```
samijaber, line 2267: We probably wanna allow propagating `next` flags from users:

```suggestion
      next: { revalidate: 1, ...options?.next },
```

samijaber (approved):

PR #2503, opened 2023-07-28
fix(pacakges/react): fix bug with evaluting repeat/collection in SSR …

teleaziz:
…with isolate-vm

### DIFF
--- a/packages/react/src/functions/string-to-function.ts
+++ b/packages/react/src/functions/string-to-function.ts
@@ -150,6 +150,7 @@ export function stringToFunction(
           )
         );
         try {
+          // returning objects throw errors in isolated vm, so we stringify it and parse it back
           const res = JSON.parse(resultStr);
           return res;
         } catch (_error: any) {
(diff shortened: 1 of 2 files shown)


shyam-builder (approved):

packages/react/src/functions/string-to-function.ts
```diff
@@ -195,6 +201,9 @@ const makeFn = (code: string, useReturn: boolean) => {
   }
   return new Proxy({}, {
       get(target, key) {
+          if (key === 'copySync') {
+            return () => obj.copySync();
+          }
```
samijaber, line 206: where is that `copySync` key coming from exactly? I assume I might need the same workaround in the new SDKs when I add isolate vm

teleaziz: it's coming from the `ivm.Reference` https://github.com/laverdet/isolated-vm#referencecopysync , I refactored a bit to improve readability and added more comments

PR #2531, opened 2023-08-04
Feature: SDKs: fetchBuilderProps

samijaber:
## Description

Lots of feedback from @manucorporat on making DX better for the SDKs
 
SDKs:
- deprecate `RenderContent` and `RenderBlocks`, exporting them under new names `Content` and `Blocks`.
- deprecate `getContent` and `getAllContent`, exporting them under new names `fetchOneEntry` and `fetchEntries` (to be more explicit about data-fetching happening)
- create a new export: `fetchBuilderProps`, which wraps `fetchContent` and also sets sensible defaults. It allows you to pass either a `url: URL` _or_ a `{ path: string; searchParams: URLSearchParams | object }`, which it uses to set the `userAttributes.urlPath` and extract builder preview from search params.


packages/sdks/src/functions/fetch-builder-props.ts
```diff
@@ -0,0 +1,72 @@
...
+        url?: undefined;
+        path?: undefined;
+        searchParams?: undefined;
+      }
+  );
+
+/**
+ * Fetches builder content, and returns it along with sensible defaults for other props that `Content` needs to render.
+ */
+export const fetchBuilderProps = async (
```
samijaber, line 48: This is the new helper export

steve8708:
we've never really had any user feedback saing RenderContent or RenderBlocks were confusing or anything. I personally would vote that I like `Render` in there because it is a dynamic rendering system, but don't feel too strongly. not sure it's worth a breaking change tbh tho, unless we just create an alias and use the new one in docs but not break peoples old/current integrations (still export the old name too just flag as deprecated)

if we are going to rename getContent and getAllContent I would suggest addressing the main feedback we've heard - which is that people aren't clear that getContent = get one and they assume getAllContent = get _everything_ in one fetch (no limit). what we mean is get one vs get a list - maybe @adamdbradley or @manucorporat hav eideas but this has been repeated feedback so if we're going to do breaking name changes I would suggest addressing that if we have ideas

i'm slightly confused on `fetchBuilderProps`, can you give an example on the old vs new ways of using this? it suggests to me we are dynamically fetching props (aka component inputs) which I don't think this does

steve8708:
just caught the end of your loom that explains fetchBuliderProps. the only confusing part still for me si the props themselves aren't fetched, it feels to me like we are getting props from the backend which is not true, the props are constructed client side we are just dynamically pulling the value of `content`. maybe it' sjust me that has that confusion tho

glad to see from your loom we are still exporting the old names as deprecated so that def helps for my main concern on the first point above

packages/sdks/src/functions/get-content/index.ts
```diff
@@ -87,19 +95,24 @@ export const processContentResult = async (
   return content;
 };
 
-export async function getAllContent(options: GetContentOptions) {
+export async function fetchAllEntries(options: GetContentOptions) {
```
steve8708, line 98: i would suggest calling this `fetchEntries` since that makes it clear it's plural and removes `all` since we saw repeat confusion that people thought `all` means actually give me everything, vs just give me multiple. we could call it `fetchMultipleEntries` or `fetchEntriesList` but feels like `fetchEntries` is nicer

PR #2538, opened 2023-08-07
feat(packages/react): add `customComponents` prop 

teleaziz:


samijaber (approved):
Nice, I take it we're backfilling the new approach into the gen1 SDK?

PR #2541, opened 2023-08-08
Change element type of text components to div

STRd6:
## Description

Since the text components often contain block level elements like `p` I think they should be `div` rather than `span`. This will also mollify some html validation tools.

If there is some strong reason why we need to use `span` let me know so I can document it. 

Thanks!

### DIFF
--- a/packages/sdks-tests/src/specs/css-nesting.ts
+++ b/packages/sdks-tests/src/specs/css-nesting.ts
@@ -6,7 +6,7 @@ export const CONTENT = {
     title: 'css-nesting',
     themeId: false,
     cssCode:
-      "/*\n* Custom CSS styles\n*\n* Global by default, but use `&` to scope to just this content, e.g.\n*\n*   & .foo {\n*     color: 'red'\n*   }\n*/\n\n& > div > div > div {\n  color: rgb(0, 0, 255)\n}\n& > div > div > div > div {\n  color: rgb(65, 117, 5)\n}\n",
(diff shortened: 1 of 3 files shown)


samijaber:
@STRd6 This could have substantial impacts on styling and break user expectations. Could you double check in both gen1 and gen2 SDKs what the before/after of text blocks looks like? Does this have styling impact?

BTW: the e2e tests are failing here: https://github.com/BuilderIO/builder/blob/48783b29c1734c3c3323e7a2f85d863d2a62f8cb/packages/sdks-tests/src/tests/styles.spec.ts#L107-L114

In that case, the test is checking if arbitrary user-provided CSS with nesting operators works. I had written the CSS here: https://github.com/BuilderIO/builder/blob/48783b29c1734c3c3323e7a2f85d863d2a62f8cb/packages/sdks-tests/src/specs/css-nesting.ts#L9

with the assumption that it would be a `span` 😄. That's not a big deal, and we can alter the test accordingly, but we should check internally if this is an acceptable change to make.

STRd6:
@samijaber It could have styling impact if people were depending on the internal element structure of the text component for their custom css. We may want to wait for the next major release since it is technically a breaking change.

packages/react/src/blocks/Text.tsx
```diff
@@ -55,7 +55,7 @@ class TextComponent extends React.Component<TextProps> {
           return (
             <React.Fragment>
               {/* TODO: <BuilderEditableText component that wraps this for other components with text */}
-              <span
+              <div
```
teleaziz, line 58: what do you think of limiting this PR changes to the gen2 sdks only?

STRd6: I think that's a good idea, we can revisit the idea in gen1 before the next major release.

STRd6:
@samijaber @teleaziz is this good to merge now that it is new sdks only?

samijaber (approved):

PR #2559, opened 2023-08-23
fix inherit state for symbols

harmeet-builder:
## Description

Omit copying of children when inherit state is enabled for symbols in react sdk


samijaber:
@harmeet-builder Is there any way to show that this fix works as intended? Can we add an e2e test? You can see the steps to add a test here: https://github.com/BuilderIO/builder/blob/main/packages/sdks/DEVELOP.md#test-an-sdk

Let me know if you have any questions.

samijaber:
@harmeet-builder I just noticed: we actually still don't have the `Slot` Block in gen2 SDKs: https://github.com/BuilderIO/builder/tree/main/packages/sdks/#builder-blocks

That's why I think all the gen2 SDK tests are failing. You will need to disable the test for all of those as well, with a note that they should be re-enabled when `Slot` is implemented in gen2 SDKs.

packages/sdks-tests/src/tests/blocks.spec.ts
```diff
@@ -450,9 +450,19 @@ test.describe('Blocks', () => {
...
   });
 
-  test('nested symbols with inherit', async ({ page }) => {
+  test('nested symbols with inherit', async ({ packageName, page }) => {
     await page.goto('/nested-symbols');
 
+    // Skipping the test as v2 sdks currently don't support Slot
+    // gen1-remix and gen1-next are also skipped because React.useContext is not recognized
+    if (
+      ['react-native', 'vue2', 'vue3', 'svelte', 'qwik-city', 'gen1-remix', 'gen1-next'].includes(
```
samijaber, line 459: There are actually a lot more packages to exclude:

https://github.com/BuilderIO/builder/blob/a899abbc2aa195dc9e2ee756d72210c00912806b/packages/sdks-tests/src/tests/sdk.ts#L26-L44

```
export type PackageName =
  | 'react-native'
  | 'solid'
  | 'solid-start'
  | 'qwik-city'
  | 'next-pages-dir'
  | 'next-app-dir-client'
  | 'next-app-dir'
  | 'react'
  | 'vue2'
  | 'vue3'
  | 'nuxt3'
  | 'nuxt2'
  | 'svelte'
  | 'sveltekit'
  | 'gen1-react'
  | 'gen1-remix'
  | 'gen1-next';
```

All need to be skipped, except for `gen1-react`.

harmeet-builder: Added all except gen1-react

samijaber (approved):
Awesome, thanks for adding the test 👍🏽

PR #2571, opened 2023-09-05
fixed exclude from translation for non text elements

gustavocodess:
Removed comparison to text element to exclude elements from translation

How to test:

1 - Install smartling or phrase plugin
2 - Enable localization
3 - Right click on a non text element and check if the exclude from translation options is displayed

### DIFF
--- a/packages/utils/package-lock.json
+++ b/packages/utils/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@builder.io/utils",
-  "version": "1.1.15",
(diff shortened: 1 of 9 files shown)


samijaber:
I am going over our repo and closing very stale PRs. [@gustavocodess](https://github.com/gustavocodess) Feel free to reopen this if it is still needed.

Heads up: this PR seemed to have published its changes to 2 packages, so it's possible we might need to re-add them in the future if necessary

PR #2584, opened 2023-09-22
[Sdk] Set 'noTraverse' to true as default when using getAll builder api

murtaza-builder:
## Description
Make 'noTraverse' option as true by default for getAll api invocation. It still can be overridden if user passes it.

### DIFF
--- a/packages/sdks/src/functions/get-content/index.ts
+++ b/packages/sdks/src/functions/get-content/index.ts
@@ -47,6 +47,16 @@ type ContentResponse =
     };
 
 const _fetchContent = async (options: GetContentOptions) => {
+  if (options.limit === undefined || options.limit > 1) {
+    if (!options.options) {
+      options.options = {
+        enrich: false,
+      };
+    } else if (!('enrich' in options.options)) {
+      options.options.set('enrich', false);
+    }
+  }
+
   const url = generateContentUrl(options);
 
   const res = await fetch(url.href);
--- a/packages/sdks/src/functions/get-content/index.ts
+++ b/packages/sdks/src/functions/get-content/index.ts
@@ -47,13 +47,6 @@ type ContentResponse =
     };
 
 const _fetchContent = async (options: GetContentOptions) => {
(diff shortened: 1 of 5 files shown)


packages/sdks/src/functions/get-content/index.ts
```diff
@@ -47,6 +47,13 @@ type ContentResponse =
     };
 
 const _fetchContent = async (options: GetContentOptions) => {
+  // Set noTraverse=true if NOT already passed by user, for query performance
+  if (options.limit === undefined || options.limit > 1) {
```
samijaber, line 51: so if I understand correctly, when `limit === =`, then we don't want to set `noTraverse = true`? I'm surprised this condition is in gen2 but not gen1. Any reason for that?

shyam-builder: in Gen1, there is a separate `getAll` call, which is where the change is made. In Gen2, from my understanding, there is a common `getContent` call used for both individual and multiple fetches.

samijaber: I see. Can we move this logic to `generateContentUrl`? This is where we set the default values for all of our `options`. It is a more ideal place to have this logic in

murtaza-builder: I understand the benefit, but I cant move it coz `generateContentUrl` assumes a default limit of 30 which affect my condition to apply `noTraverse`. It should only apply in case of `fetchEntries` and not to the single `fetchEntry` case.

murtaza-builder: nvmd, found a way where it was possible. Moved it. pls have a look

murtaza-builder: @samijaber @shyam-builder

samijaber: Awesome, thanks @murtaza-builder! Your build is failing because you need to update the snapshot tests for generateContentUrl

samijaber (approved):

PR #2586, opened 2023-09-28
support subFields showIf

youleapnir:
## Description

This PR tries to support `showIf` in builder `subFields`.

## Notice
This has not been checked yet, because I don't know how to run builder.

## Motivation
I want `subFields` will be able to be shown conditionally like regular fields with `showIf`.
Here is an example of `showIf`:
https://github.com/BuilderIO/builder/assets/104710778/1037361d-b6d7-4862-94dc-7629d5eb72c3

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -14,14 +14,13 @@ import { omit } from './functions/omit.function';
 import { getTopLevelDomain } from './functions/get-top-level-domain';
 import { BuilderContent } from './types/content';
 import { uuid } from './functions/uuid';
-import { parse as urlParse } from './url';
+import { emptyUrl, parse as urlParse, UrlLike } from './url';
 
 // Do not change this to a require! It throws runtime errors - rollup
 // will preserve the `require` and throw runtime errors
 import hash from 'hash-sum';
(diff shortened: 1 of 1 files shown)


packages/core/src/builder.class.ts
```diff
@@ -1027,27 +1026,38 @@ export class Builder {
...
+    input: Input;
+    keysToConvert: string[];
+  }) {
+    for (const key of keysToConvert) {
+      if (input[key] && typeof input[key] === 'function') {
+        const fn = input[key];
+        input = {
+          ...input,
+          [key]: `return (${fn.toString()}).apply(this, arguments)`,
+          ...(input.subFields == null
```
teleaziz, line 1042: there might be a lint rule type coercion with double equeals, maybe we do this instead?
```
  ...(Array.isArray(input.subFields)
    ? {
        subFields: input.subFields.map((input) => {
          return this.convertFnToString({ keysToConvert, input });
        }),
      }
    : {}),
```

teleaziz (commented):
Thanks for contributing! left one comment, otherwise looks good

samijaber:
Closing due to being stale. Please reopen if you are still interested in contributing this code!

PR #2835, opened 2023-11-13
Add meta field to Component interface

adamdbradley:
Add `meta` to the `Component` interface. This is will be useful to when linking the Figma main component to a registered component. This is only for types, and no changes to the runtime.

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -563,6 +563,11 @@ export interface Component {
    * for more information on permissions in builder check https://www.builder.io/c/docs/guides/roles-and-permissions
    */
   requiredPermissions?: Array<Permission>;
+
+  /**
+   * Used to store additional component information.
+   */
+  meta?: Record<string, any>;
 }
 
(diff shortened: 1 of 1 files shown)


samijaber (approved):

packages/core/src/builder.class.ts
```diff
@@ -563,6 +563,11 @@ export interface Component {
    * for more information on permissions in builder check https://www.builder.io/c/docs/guides/roles-and-permissions
    */
   requiredPermissions?: Array<Permission>;
+
+  /**
+   * Used to store additional component information.
+   */
+  meta?: Record<string, any>;
```
samijaber, line 570: Oh, can you add this to the `ComponentInfo` in the new SDKs as well? Your change currently only applied to the gen1 React SDK

https://github.com/BuilderIO/builder/blob/main/packages/sdks/src/types/components.ts#L4

samijaber (commented):
@adamdbradley I think your branch is a very old one, hence the CI vercel deploys failing. You might want to rebase against latest `main`.

PR #2859, opened 2023-11-29
added data plugin for bigcommerce that lets products and categories b…

snowdogg:
## Description

Added support for a big commerce data plugin that has a dropdown to select products and categories which are added to state when selected.

### DIFF
--- a/plugins/bigcommerce/src/data-plugin.ts
+++ b/plugins/bigcommerce/src/data-plugin.ts
@@ -7,140 +7,126 @@ import { BuilderStore, BuilderStoreContext } from '@builder.io/react';
 import { useContext, useEffect } from 'react';
 
 const buildHeaders = (headers: any) => {
-
-    return Object.entries(headers)
-        .map(([key, value]) => `headers.${key}=${value}`)
-        .join('&');
+  return Object.entries(headers)
+    .map(([key, value]) => `headers.${key}=${value}`)
+    .join('&');
 };
 
+function buildBigCommerceUrl({ resource, resourceId, query, limit, headers }) {
+  const base = `https://cdn.builder.io/api/v1/proxy-api?url=https%3A%2F%2Fapi.bigcommerce.com%2Fstores%2Fpdfuqrrtbk%2Fv3%2Fcatalog`;
+  let path = resource;
+  if (path === 'product') {
+    path = 'products';
+  } else if (path === 'category') {
+    path = 'categories';
+  }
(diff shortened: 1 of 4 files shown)


plugins/bigcommerce/src/data-plugin.ts
```diff
@@ -0,0 +1,160 @@
...
+                        {
+                            friendlyName: 'Search',
+                            name: 'query',
+                            type: 'string',
+                        },
+                    ],
+
+                    toUrl: ({ entry, query, limit }) =>
+                        //this request fails due to lack of auth headers but doesn't seem to affect the functionality
+                        buildBigCommerceUrl({
```
teleaziz, line 112: The url has to work for the data to be fetched and included in content, you can append headers to proxied calls same as this function https://github.com/BuilderIO/builder/blob/539f435a3a10f8b44e517e6f794dfc2c6261784d/plugins/contentstack/src/plugin.ts#L218

plugins/bigcommerce/src/data-plugin.ts
```diff
@@ -0,0 +1,160 @@
...
+            if (entry) {
+                //get the product or category's data
+                const entryObj = await service[resourceTypeId].findById(entry);
+
+                //filter out undefined properties
+                const newObj = filterOutUndefinedProperties(entryObj)
+
+                //set the product or category data to state
+                appState.designerState.editingContentModel.data.set('state', { [resourceTypeId]: newObj })
+
```
teleaziz, line 134:
```suggestion
```

this is not the way to add data objects to the state, you'd want to fix the HTTP request url above

teleaziz (changes requested):

plugins/bigcommerce/src/data-plugin.ts
```diff
@@ -0,0 +1,129 @@
+import { APIOperations, ResourceType } from '@builder.io/data-plugin-tools';
+import appState from '@builder.io/app-context';
+import { CommerceAPIOperations } from '@builder.io/commerce-plugin-tools';
+import pkg from '../package.json';
+type BigCommerceResourceType = 'product' | 'category';
+import { BuilderStore, BuilderStoreContext } from '@builder.io/react';
+import { useContext, useEffect } from 'react';
```
teleaziz, line 7: some of those import are not used here, please remove, ( strange that the linter didn't detect it )

teleaziz (approved):
Great work Andreas, left one comment

PR #2860, opened 2023-11-29
Dependency Updates

JuliusDorfman:
## Description
Removed eslint-plugin-svelte3 - No longer supported per https://github.com/sveltejs/eslint-plugin-svelte3
Replaced with Official package: eslint-plugin-svelte https://github.com/sveltejs/eslint-plugin-svelte

Other dependency updates


samijaber (changes requested):
- You need to run `npm install` and update the `package-lock.json` file with your changes too
- Can you share a loom showing that the example works with your updates?

samijaber (approved):

PR #2863, opened 2023-12-04
Update CloudinaryImageEditor component buttons to make the `set credentials` button less prominent once credentials are set

jonathan-connecteam:
Update CloudinaryImageEditor component buttons to make the `set credentials` button less prominent once credentials are set

This reduces accidental credential changing by non-technical users

## Description

This is a simple UX improvement to the Cloudinary plugin that makes changes the `Set credentials` button text to `...` after they are set, and makes the button smaller appropriately. Since there isn't a constant need to change the credentials for each image, it makes sense to make the button smaller.

Screenshot 2023-12-04 at 18 03 49


mrkoreye (approved):
@jonathan-connecteam sorry we missed this one! Thank you for the improvement. cc @gustavocodess

PR #2880, opened 2023-12-12
SDKs: fix video element discrepancy in gen2 sdk

sidmohanty11:
## Description

In gen2 SDKs we spread out all attributes which lead to addition of extra attributes to the video element like:
- `builder-id`
-  class=`builder-id`
that causes duplication of styles as the styles already gets applied in the root block element.

Reference: https://github.com/BuilderIO/builder/blob/main/packages/react/src/blocks/Video.tsx

Also, we are handling styles differently for some specific conditions (for example, handling video if aspect ratio is explicitly mentioned etc) that leads to mismatch in video element render

_Screenshot_

Before:
Screenshot 2023-12-12 at 7 18 42 PM

https://github.com/BuilderIO/builder/assets/73601258/7ef938ef-aa11-4595-8f42-6c9d6578ae10

After:

https://github.com/BuilderIO/builder/assets/73601258/73b00e84-754c-4fae-8461-2bbabc8bcfc1

PS - used this [vite plugin](https://www.npmjs.com/package/vite-plugin-realtime-diff) for UI diff


packages/sdks/src/blocks/video/video.lite.tsx
```diff
@@ -46,27 +48,75 @@ export default function Video(props: VideoProps) {
...
+            ? {
+                position: 'absolute',
+              }
+            : null),
+        }}
+        src={props.video || 'no-src'}
+        poster={props.posterImage}
+        class="builder-video"
+      >
+        {!props.lazyLoad && <source type="video/mp4" src={props.video} />}
```
murtaza-builder, line 80: have we handled case when lazyLoad is true?

packages/sdks/src/blocks/video/video.lite.tsx
```diff
@@ -46,27 +48,75 @@ export default function Video(props: VideoProps) {
...
+      </video>
+      <Show when={props.aspectRatio && !(props.fitContent && props.children)}>
+        <div
+          style={{
+            width: '100%',
+            paddingTop: props.aspectRatio! * 100 + '%',
```
murtaza-builder, line 86: I dont think expressions in RHS work in Mitosis. you could define a getter above to return the computed value

packages/sdks/src/blocks/video/video.lite.tsx
```diff
@@ -46,27 +48,75 @@ export default function Video(props: VideoProps) {
...
+          borderRadius: 1,
+          ...(props.aspectRatio
+            ? {
+                position: 'absolute',
+              }
+            : null),
+        }}
+        src={props.video || 'no-src'}
+        poster={props.posterImage}
+        class="builder-video"
```
murtaza-builder, line 78: do we have any css based on this or is it just a placeholder? class names actually come from parent props to this component i think, which might get overriden

sidmohanty11: its actually a placeholder for video same as gen1, the classes that we define are getting added to the parent div of video element

packages/sdks/src/blocks/video/video.lite.tsx
sidmohanty11: thanks, I have updated this!

samijaber (commented):
Nice! Could we add a basic e2e test for the video block to make sure it renders correctly? and maybe is interactive?

See https://github.com/BuilderIO/builder/blob/main/packages/sdks/DEVELOP.md#integration-testing for a guide on creating e2e tests, and let me know if any information is missing.

packages/sdks/src/blocks/video/video.lite.tsx
sidmohanty11: you're right we are not handling lazyload here

packages/sdks/src/blocks/video/video.lite.tsx
sidmohanty11: Adding it to a getter under store is not working for vue2/nuxt2 but this expression is working for all

sidmohanty11:
Hey @murtaza-builder thanks a lot for the review, you're correct (my bad) its `builder-<id>` as classes. The duplication happens with the video element as the parent builder-block already has the styles needed

sidmohanty11:
Hey @samijaber thanks a lot for the review! Yup, just added a basic e2e tests to check for styles please take a look. Also for the mocking of fetching the video URL from the CDN I've added a mocker there that fetches from the local filesystem. Please let me know if I should move forward with this approach or make any changes

packages/sdks-tests/src/tests/blocks.spec.ts
```diff
@@ -92,6 +99,72 @@ test.describe('Blocks', () => {
...
+    const expectedVals = expectedCSS.map((val, i) => ({ val, i }));
+
+    for (const { val, i } of Object.values(expectedVals)) {
+      const video = videoLocator.nth(i);
+      const expected = val;
+      for (const property of Object.keys(expected)) {
+        await expect(video).toHaveCSS(property, expected[property]);
+      }
+    }
+  });
```
samijaber, line 166: That's awesome! Very well done 😄.

One last thing would be to add a test that proves your PR works as intended! From what I can tell, you are providing a child `<p>asfgasgasgasg some test</p`. Could you check that it is visible in the page? 🙏🏽

samijaber (commented):
Just need 1 more test to make sure the children indeed render as expected now, and then we're good to go 👍🏽

packages/sdks-tests/src/tests/blocks.spec.ts
sidmohanty11: Thank you!! I've added the test that shows that the children are visible and rendered, please let me know if any changes are needed to be made

samijaber (commented):
Perfect!

Sorry, I forgot the very last step. Add your changeset so your change appears in the release notes: https://github.com/BuilderIO/builder/blob/main/packages/sdks/PUBLISHING.md#1--add-changeset

sidmohanty11:
Hey @samijaber done! Please have a look : )

samijaber (approved):
Perfect, thanks @sidmohanty11 ! Very well done on your first SDK PR 👍🏽

Issue #2878, opened 2023-12-07
MobX Error Triggered When Selecting Input from Emporix Product Preview Modal List


omspacecode:
**Describe the bug**
When attempting to select an input from the modal list in a content entry that includes the Emporix product preview field, a MobX error is triggered upon clicking a list item.

**To Reproduce**

Navigate to a content entry containing the Emporix product preview field.
Attempt to select an input from the modal list.
Upon clicking on a list item, a MobX error is triggered.

**Expected behavior**
The product should be selected without errors, and the modal should close.

**Screenshots**
<img width="1799" alt="Screenshot 2023-12-07 at 11 59 28 AM" src="https://github.com/BuilderIO/builder/assets/27288383/489575a4-0318-488d-973d-3120e02e8223">

**Additional context**
This issue persists when using it on Page content, data model content, and as a custom component input field.

cc: @lstypka   (Tagging the original creator of the plugin)

omspacecode:
Hey @lstypka  , could you please help with an initial review of the issue and provide your feedback? Thank you!

lstypka:
@omspacecode the issue looks valid. Do you know if the issue is in the plugin code itself or the root cause is related with the core library ?

marwellner71:
@omspacecode I am Product Manager at Emporix and our joined customer is asking for a status. It would be very helpful if you could give us feedback to the question from Lukasz. Thanks a lot in davance!

omspacecode:
@lstypka I've been looking into the reported issue, but I'm unsure if it stems from the plugin code or the core library.
Given your expertise, could you check if it's related to the plugin code? There don't seem to be issues with other plugins.
Thank you so much

omspacecode:
@marwellner71 @lstypka  I look forward to hearing from you at your earliest convenience. Thank you.

lstypka:
Hi @omspacecode 
I've created the pull request that fixes the issue: https://github.com/BuilderIO/builder/pull/2884/files
The issue was related to the fact that the plugin used async function for getRequestObject and builderIO do not accept it. I noticed that before the getRequestObject method other methods are invoked and the token can be cached, therefore the token is taken from a cache for the getRequestObject now. 
All the Emporix API requires authorization token (for example anonymous token), therefore maybe it would be good if BuilderIO would support async function for getRequestObject in the future ? 

Regards
Łukasz

omspacecode:
[...]

PR #2884, opened 2023-12-18
#2878 Fixed issue with Emporix plugin error

lstypka:
## Description

Emporix plugin used async function for getRequestObject method. Changed the implementation to get the token from a cache

_Screenshot_


gustavocodess (approved):
Looks good, thanks for the fix!

PR #2886, opened 2023-12-19
Synced ComponentInfo to the current state of Component interface in gen1 builder.class.ts

murtaza-builder:
## Description
Synced missing fields in gen2's `ComponentInfo` interface with gen1's `Component` interface in `builder.class.ts`

### DIFF
--- a/packages/sdks/src/server-index.ts
+++ b/packages/sdks/src/server-index.ts
@@ -28,3 +28,4 @@ export type { ComponentInfo } from './types/components.js';
 
 export { fetchBuilderProps } from './functions/fetch-builder-props.js';
 export { isBrowser } from './functions/is-browser.js';
+export { isServer } from './functions/is-server.js';
--- a/packages/sdks/src/server-index.ts
+++ b/packages/sdks/src/server-index.ts
@@ -27,3 +27,4 @@ export type { RegisteredComponent } from './context/types.js';
 export type { ComponentInfo } from './types/components.js';
 
 export { fetchBuilderProps } from './functions/fetch-builder-props.js';
+export { isBrowser } from './functions/is-browser.js';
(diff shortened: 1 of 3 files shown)


packages/sdks/src/server-index.ts
```diff
@@ -27,3 +27,5 @@ export type { RegisteredComponent } from './context/types.js';
 export type { ComponentInfo } from './types/components.js';
 
 export { fetchBuilderProps } from './functions/fetch-builder-props.js';
+export { isBrowser } from './functions/is-browser.js';
+export { isServer } from './functions/is-server.js';
```
samijaber, line 31: I don't think we want to export these in the new SDKs. It makes little sense to provide our users with helpers like these: they have nothing to do with SDK functionality.

What is the use case for adding them here?

murtaza-builder: We have exposed these helpers before in Gen1 and they are widely used in client code, example - in our site. Migrating to gen2 will require users with a substitute, thus exporting them

samijaber: I understand, and that's the way to go. I would rather users rely on their own isBrowser/isServer helpers than reuse ours. The Builder SDK shouldn't be exporting generic helpers such as these: they are unrelated to the Builder SDK's purpose.

packages/sdks/src/types/components.ts
```diff
@@ -7,25 +7,70 @@ export interface ComponentInfo {
...
+   */
+  override?: boolean;
+
   /**
    * Input schema for your component for users to fill in the options
    */
   inputs?: Input[];
+  /** @hidden @deprecated */
   class?: any;
+  /** @hidden @deprecated */
```
samijaber, line 51: Thanks for adding these tags. We will want to go ahead and completely remove any deprecated attributes in gen2 SDKs before we bump to v1, as this is our big chance to remove deprecated features.

They can stay here for now though, and these tags will help us know what to remove in the near future 👍🏽

samijaber:
Going over stale PRs: going to close this one. we can reopen if we decide to re-tackle it.

PR #2908, opened 2024-01-11
Emporix plugin access token expired after 4 hours and it wasn't refreshed

lstypka:
## Description

In the initial version of the plugin there was no code responsible for checking if a cached token expired or not. In the pull request I've added a logic that stores information how long the token is valid and in case when the cached token expired then a new token is generated. I've also bumped up the plugin version

_Screenshot_


mrkoreye (approved):

gustavocodess (approved):
Looks good, thanks for working on this!

PR #2920, opened 2024-01-17
[fix] duplicate attributes being applied wrapper + block element

sidmohanty11:
## Description

With the InteractiveElements we are passing all the attributes in the block. In this PR, we are checking if we should pass the props to the block itself or to the wrapper element only. This way we are fixing duplicate attributes being sent to the child component as well.

### DIFF
--- a/.changeset/nasty-lies-beam.md
+++ b/.changeset/nasty-lies-beam.md
@@ -8,4 +8,4 @@
 '@builder.io/sdk-vue': patch
 ---
 
-fix duplicate attributes getting applied to inner element
+Fix: duplicate attributes getting applied to both the block and its wrapper element.
--- a/.changeset/nasty-lies-beam.md
+++ b/.changeset/nasty-lies-beam.md
@@ -1,0 +1,11 @@
+---
+'@builder.io/sdk-react-nextjs': patch
+'@builder.io/sdk-qwik': patch
+'@builder.io/sdk-react': patch
+'@builder.io/sdk-react-native': patch
+'@builder.io/sdk-solid': patch
+'@builder.io/sdk-svelte': patch
+'@builder.io/sdk-vue': patch
+---
+
+fix duplicate attributes getting applied to inner element
--- a/packages/sdks-tests/src/tests/duplicate-attributes.spec.ts
+++ b/packages/sdks-tests/src/tests/duplicate-attributes.spec.ts
@@ -3,6 +3,7 @@ import { test } from './helpers.js';
 
 test.describe('Duplicate Attributes', () => {
   test('wrapped block has no duplicate attributes', async ({ page, packageName }) => {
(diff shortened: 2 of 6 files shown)


packages/sdks-tests/src/tests/duplicate-attributes.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { expect } from '@playwright/test';
+import { test } from './helpers.js';
+
+test.describe('Duplicate Attributes', () => {
+  test('wrapped block has no duplicate attributes', async ({ page, packageName }) => {
+    await page.goto('/duplicate-attributes');
```
samijaber, line 6: you're going to want to `test.skip(packageName === 'react-native')` to ignore the test in react native, because it doesn't have HTML attributes in the same way.

sidmohanty11: Thanks! Got it!

.changeset/nasty-lies-beam.md
```diff
@@ -0,0 +1,11 @@
...
+'@builder.io/sdk-react-nextjs': patch
+'@builder.io/sdk-qwik': patch
+'@builder.io/sdk-react': patch
+'@builder.io/sdk-react-native': patch
+'@builder.io/sdk-solid': patch
+'@builder.io/sdk-svelte': patch
+'@builder.io/sdk-vue': patch
+---
+
+fix duplicate attributes getting applied to inner element
```
samijaber, line 11:
```suggestion
Fix: duplicate attributes getting applied to both the block and its wrapper element.
```

packages/sdks/src/components/block/components/interactive-element.lite.tsx
```diff
@@ -34,16 +38,18 @@ export default function InteractiveElement(
...
-          block: props.block,
-          context: props.context.value,
-        }),
-        ...getBlockActions({
-          block: props.block,
-          rootState: props.context.value.rootState,
-          rootSetState: props.context.value.rootSetState,
-          localState: props.context.value.localState,
-          context: props.context.value.context,
+        ...(props.includeBlockProps && {
```
samijaber, line 41: You could remove one layer of spreading here:

```ts
attributes={props.includeBlockProps
  ? {
      ...getBlockProperties({
        block: props.block,
        context: props.context.value,
      }),
      ...getBlockActions({
        block: props.block,
        rootState: props.context.value.rootState,
        rootSetState: props.context.value.rootSetState,
        localState: props.context.value.localState,
        context: props.context.value.context,
      }),
    }
  : {}}
```

sidmohanty11: Thanks a lot, let me apply this!

samijaber (approved):
Well done! 🎉

PR #2943, opened 2024-01-23
stricter trusted hosts

STRd6:
## Description

No longer allow arbitrary subdomains of builder.io in trusted hosts.

### DIFF
--- a/packages/core/src/builder.class.test.ts
+++ b/packages/core/src/builder.class.test.ts
@@ -12,6 +12,7 @@ describe('Builder', () => {
   test('arbitrary builder.io subdomains', () => {
     expect(Builder.isTrustedHost('cdn.builder.io')).toBe(false);
     expect(Builder.isTrustedHost('foo.builder.io')).toBe(false);
+    expect(Builder.isTrustedHost('evildomainbeta.builder.io')).toBe(false);
   });
 
   test('add trusted host', () => {
--- a/packages/core/src/builder.class.test.ts
+++ b/packages/core/src/builder.class.test.ts
@@ -1,0 +1,10 @@
+import { Builder } from './builder.class';
+
+describe('Builder', () => {
+  test('trustedHosts', () => {
+    expect(Builder.isTrustedHost('example.com')).toBe(false);
+    expect(Builder.isTrustedHost('localhost')).toBe(true);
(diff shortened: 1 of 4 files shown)


packages/core/src/builder.class.test.ts
```diff
@@ -0,0 +1,15 @@
+import { Builder } from './builder.class';
+
+describe('Builder', () => {
+  test('trustedHosts', () => {
+    expect(Builder.isTrustedHost('example.com')).toBe(false);
+    expect(Builder.isTrustedHost('localhost')).toBe(true);
+    expect(Builder.isTrustedHost('builder.io')).toBe(true);
+    expect(Builder.isTrustedHost('123-review-build.beta.builder.io')).toBe(true);
```
samijaber, line 8: can we add a test for the problematic `builder.io` subdomains? `cdn.builder.io`, etc.

mrkoreye: What about white label instances? Such as `umg-mira.builder.io` or  `marketer.mira.umusic.com` etc. I suppose that is why we have the `registerTrustedHost` method? 

Can we add tests to that to make sure it works as expected? My worry is that if we release this as a patch update it might break some weird edge cases.

mrkoreye: One more we probably want to add is `playground.builder.io`

mrkoreye: And `qa.builder.io`, also will this update break support for `www.builder.io`?

samijaber: Should be a minor release, since it potentially involves changes on the user's end to add their hosts

packages/core/src/builder.class.test.ts
```diff
@@ -0,0 +1,16 @@
+import { Builder } from './builder.class';
+
+describe('Builder', () => {
+  test('trustedHosts', () => {
+    expect(Builder.isTrustedHost('localhost')).toBe(true);
+    expect(Builder.isTrustedHost('builder.io')).toBe(true);
+    expect(Builder.isTrustedHost('beta.builder.io')).toBe(true);
```
samijaber, line 7: Another small edge-case to test: `evildomainbeta.builder.io`

samijaber (approved):

PR #2947, opened 2024-01-24
Check `this.useLazyLoading` property to determine loading behavior for Image component

theronpatrick:
## Description

Right now, the built in `Image` component for builder will _always_ set `loading=lazy` for any image that is not the tracking pixel. But there is a prop `lazy` that is meant to configure the loading behavior, which in turn gets checked in `useLazyLoading`. It's counterintuitive that setting `lazy=false` will still result in an image that has the property `loading=lazy` set. 

This change would allow configuring the prop `lazy={false}` on the `Image` component to allow changing the loading behavior from `lazy` to `eager`.

### DIFF
--- a/packages/react/src/blocks/Image.tsx
+++ b/packages/react/src/blocks/Image.tsx
@@ -342,7 +342,7 @@ class ImageComponent extends React.Component<any, { imageLoaded: boolean; load: 
                   },
                 }),
               }}
-              loading={isPixel ? 'eager' : 'lazy'}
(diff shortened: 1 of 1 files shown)


samijaber:
Thank you for the PR! We understand the reasoning for this change, but to support the existing legacy behaviour we will not be changing the lazy-loading logic for the `Image` component in the gen1 React SDK. 

However, we are currently working on our gen2 SDKs, and we plan to modify its `Image`'s lazy-loading behaviour to be configurable as you described.

PR #2952, opened 2024-01-24
WIP: Trusted hosts gen2

STRd6:
## Description

Adding trusted hosts check to gen2 sdks.


packages/sdks/src/functions/is-from-trusted-host.ts
```diff
@@ -0,0 +1,23 @@
+const DEFAULT_TRUSTED_HOSTS = [
+  '*.beta.builder.io',
+  'beta.builder.io',
+  'builder.io',
+  'localhost',
+  'qa.builder.io',
+];
+
+export function isFromTrustedHost(
```
samijaber, line 9: Could we also get the unit tests you added to the gen1 SDK for this function? 🙏🏽

samijaber (commented):
Awesome! 

One last thing to cleanup the changelog: could you

- delete the old changelog entries that say we fixed this in 2.2.0
- add a new changelog for the next release saying we fixed it this time around

.changeset/angry-ears-drop.md
```diff
@@ -0,0 +1,11 @@
...
+'@builder.io/sdk-react-native': patch
+'@builder.io/sdk-react-nextjs': patch
+'@builder.io/sdk-svelte': patch
+'@builder.io/sdk-react': patch
+'@builder.io/sdk-solid': patch
+'@builder.io/sdk-qwik': patch
+'@builder.io/sdk-vue': patch
+---
+
+Stricter checking of trusted hosts.
```
samijaber, line 11: We should mention the new prop! Something like:

```suggestion
Added a `trustedHosts` prop to `Content`. It is used to determine whether the SDK can enable editing/previewing mode within a host.
```

samijaber (approved):

PR #2959, opened 2024-01-27
[Fix][Environments][Algolia Plugin] SUPP-298 Added safety check so that sync-to-algolia does not appear in an env's model

murtaza-builder:
Added safety check so that sync-to-algolia button does not appear, in an env's model so that changes of mixup dont arise.

### DIFF
--- a/plugins/algolia/package.json
+++ b/plugins/algolia/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@builder.io/plugin-algolia",
-  "version": "0.0.5",
+  "version": "0.0.6",
   "description": "",
   "keywords": [],
   "main": "dist/plugin.system.js",
--- a/plugins/algolia/package.json
+++ b/plugins/algolia/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@builder.io/plugin-algolia",
-  "version": "0.0.6",
+  "version": "0.0.5",
   "description": "",
   "keywords": [],
   "main": "dist/plugin.system.js",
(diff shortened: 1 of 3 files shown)


sanyamkamat (changes requested):

plugins/algolia/src/plugin.tsx
murtaza-builder, line 1: @shyam-builder Need your thoughts in this specific change. Is it fine if disable syncing models to algolia if we are in an env? Basically somewht disabling plugins from an env.
wdyt?

shyam-builder: Is there a reason why we would want to disable? I can see customers wanting to finish translation and all in a lower environment and push it up.

shyam-builder: Specifically what breaks from an environment?

plugins/algolia/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@builder.io/plugin-algolia",
-  "version": "0.0.5",
+  "version": "0.0.6",
```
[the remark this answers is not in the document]
murtaza-builder: done

plugins/algolia/src/plugin.tsx
murtaza-builder: If the installation and usage of plugin happens in same env/space its not a problem but if installation is done in main space and the usage (webhook url syncing) is done in env, it messes up the webhook url params. 
Which could lead to confusion and improper working. Happened in SUPP-298.

Thus it would be better if plugins work in main space only

samijaber:
@murtaza-builder Going over stale PRs: can this one be closed for now? It's still in draft mode

PR #2997, opened 2024-02-08
Misc: add gen1 React SDK to automated CI release flow

samijaber:


packages/react/scripts/fix-core-version.sh
```diff
@@ -0,0 +1,9 @@
+# !/bin/bash
+
+# This script is used to update the version of @builderio/mitosis in the CLI package dependencies
```
mrkoreye, line 3: `@builderio/sdk`?

PR #3023, opened 2024-02-26
Fix: React sdk edge bundle

samijaber:
## Description

- add `/index.ts` files for all components
- move all `/edge`, `/browser`. `/node` subpath exports to `/bundle/*`

React SDK:
- lazyload Content component in edge runtime using `React.lazy`
- add a `/edge` subpath export pointing to a new export file that's a modified version of `blocks-exports.ts`. It re-exports everything as normal, but wraps the `Content` export to toggle between edge/browser exports at runtime. This guarantees that the edge Content component, which is lazy loaded, does not make it to the browser.
- thanks to the above, remove need for manual workaround of edge + browser bundles in React SDK when in edge envs.


.changeset/lazy-doors-pull.md
```diff
@@ -0,0 +1,6 @@
+---
+'@builder.io/sdk-react-nextjs': patch
```
samijaber, line 2:
```suggestion
```

packages/sdks/output/svelte/package.json
```diff
@@ -31,9 +31,9 @@
       "electron": "./lib/node/index.js",
       "default": "./lib/browser/index.js"
     },
-    "./edge": "./lib/edge/index.js",
+    "./bundle/edge": "./lib/edge/index.js",
     "./node": "./lib/node/index.js",
```
samijaber, line 35:
```suggestion
    "./bundle/node": "./lib/node/index.js",
```

PR #3038, opened 2024-03-01
Update readme with absolute URLs

omspacecode:
## Description

Update the all Relative URLs  to Absolute URLs. 

_Screenshot_

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -64,7 +64,7 @@ const content = await builder
 
 The builder content is simply json that you pass to a `<BuilderComponent />` to render. [Learn more about it here](https://www.builder.io/c/docs/how-builder-works-technical)
 
-You can view all of the options for `builder.get` for fetching content [in our full reference here](https://github.com/BuilderIO/builder/blob/main/packages/core/docs/interfaces/GetContentOptions.md)
+You can view all of the options for `builder.get` for fetching content [in our full reference here](main/packages/core/docs/interfaces/GetContentOptions.md)
 
 For example, with Next.js, to render Builder as your homepage:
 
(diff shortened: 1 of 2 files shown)


packages/react/README.md
```diff
@@ -1,8 +1,8 @@
 # Builder.io React SDK
 
-NOTE: If you want to register your React Server Components as custom components in Builder, you will need to use our experimental NextJS SDK [here](../sdks/output/nextjs).
+NOTE: If you want to register your React Server Components as custom components in Builder, you will need to use our experimental NextJS SDK [here](https://github.com/BuilderIO/builder/tree/main/packages/sdks/output/nextjs).
```
samijaber, line 3: I think you only need to replace the ../ with packages/

The rest is not needed

omspacecode: That would result in below URL which seems incorrect
NG: https://github.com/BuilderIO/builder/blob/packages/sdks/output/nextjs
OK: https://github.com/BuilderIO/builder/tree/main/packages/sdks/output/nextjs

omspacecode: I am checking if adding main/packages/ , would work for all.
OK: https://github.com/BuilderIO/builder/blob/main/packages/sdks/output/nextjs

omspacecode: Adding main/packages/ would work for some. 
I see some absolute URLs used before.

samijaber: sorry those are broken. You might need to add a missing `/` to the start. Right now the generated URL is:

https://github.com/BuilderIO/builder/blob/om-dev/packages/react/packages/sdks/output/react

samijaber: I'm sorry for confusing you. I think the original absolute URL is the way to go ultimately. My suggestion was off! it won't work properly in npm.

omspacecode: No worries, I was unsure myself of whether or not the links would work on both Github and npm. 
I will revert back to absolute URLs.

omspacecode: Reverted back to absolute URLs.

samijaber (commented):
Thanks @omspacecode :pray

PR #3053, opened 2024-03-18
[fix] default styles to be present only once (not inside nested render)

sidmohanty11:
## Description

When we add a symbol it adds another content div with style tag which contains the `.builder-button` class with `unset:all;` this gets higher priority as it comes below the button styles in the html.

**Jira**
https://builder-io.atlassian.net/browse/ENG-5281

**Loom(Issue)**
https://www.loom.com/share/d7240f0ee97b4632b42348c6fb12f39a

### DIFF
--- a/.changeset/gorgeous-ants-enjoy.md
+++ b/.changeset/gorgeous-ants-enjoy.md
@@ -1,0 +1,11 @@
+---
+'@builder.io/sdk-react-nextjs': patch
+'@builder.io/sdk-qwik': patch
+'@builder.io/sdk-react': patch
+'@builder.io/sdk-react-native': patch
+'@builder.io/sdk-solid': patch
+'@builder.io/sdk-svelte': patch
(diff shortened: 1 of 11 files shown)


.changeset/gorgeous-ants-enjoy.md
```diff
@@ -0,0 +1,11 @@
...
+'@builder.io/sdk-react-nextjs': patch
+'@builder.io/sdk-qwik': patch
+'@builder.io/sdk-react': patch
+'@builder.io/sdk-react-native': patch
+'@builder.io/sdk-solid': patch
+'@builder.io/sdk-svelte': patch
+'@builder.io/sdk-vue': patch
+---
+
+fix: default styles only when content is not a nested render
```
samijaber, line 11:
```suggestion
Fix: Symbol styles overriding subsequent content styles.
```

samijaber (approved):
very nice!

Issue #3060, opened 2024-03-19
Add a way to override fetch function in fetchEntries and fetchOneEntry


jerome-quere:
Hello, I am working with the Next.js framework (with the app router) and I am trying to optimize our cache system. Next.js caches fetch requests automatically and provides tools to manually purge some parts of the cache, for example, using [revalidateTag](https://nextjs.org/docs/app/api-reference/functions/revalidateTag).

It would be super helpful if we could pass an optional custom fetch function when calling `fetchEntries` or `fetchOneEntry`. This would allow me to wrap the global fetch function and add specific Next.js config options before the global fetch is called.

Example:
```
function getMyContent(key: string) {
  return fetchOneEntry({
     ...
     key
     fetch: (url, init) => {
        init.next.tags = [`my-content-${key}`]
        return fetch(url, init)
     }
  })
} 
```

What I am doing today is overriding the global `fetch` method and trying to extract context from the URL to generate correct tags. However, I am having issues because Next.js also overrides the global `fetch` and in some cases overrides my own custom `fetch`.

PR #3061, opened 2024-03-19
feat: add optional fetch override in GetContentOptions

jerome-quere:
## Description

This PR allow to override the fetch function used in fetchEntries

Fixes: #3060


samijaber (approved):
Awesome. thanks for making this contribution @jerome-quere ! I snuck in `fetchOptions` as well while we're at it. 😄

PR #3073, opened 2024-03-22
e2e tests setup for gen2 angular SDK

sidmohanty11:
## Description

Setup e2e tests for gen2 angular sdk.

**Jira**
https://builder-io.atlassian.net/browse/ENG-5474

_Screenshot_


packages/sdks-tests/src/tests/inputs.spec.ts
```diff
@@ -1,7 +1,8 @@
 import { expect } from '@playwright/test';
-import { test } from './helpers.js';
+import { excludeTestFor, test } from './helpers.js';
 
 test.describe('Inputs', () => {
+  test.skip(excludeTestFor({ angular: true }), 'Angular Gen2 SDK not implemented.');
```
samijaber, line 5: All of these should be `test.fail()`. 

`test.fail()` will automatically tell you when the test starts to pass, which is useful when we start to implement more features!

`test.skip()` will not run the test at all, so it wont know if the test starts to pass

sidmohanty11: Sure! Let me update these

samijaber (approved):

PR #3077, opened 2024-03-27
Support for translation of Symbol's content inputs

murtaza-builder:
## Description

Now symbol's content inputs spec'd in `el.component.options.symbol.data` will also be sent for translations to Phrase and also shall be applied to content when requested.

### DIFF
--- a/packages/utils/tsconfig.json
+++ b/packages/utils/tsconfig.json
@@ -6,7 +6,8 @@
     "target": "es6",
     "esModuleInterop": true,
     "strict": true,
-    "declaration": true
+    "declaration": true,
+    "skipLibCheck": true
   },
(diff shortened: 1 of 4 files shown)


packages/utils/tsconfig.json
sanyamkamat, line 1: You can undo this change. The config is already updated via this PR https://github.com/BuilderIO/builder/pull/3076

murtaza-builder: its fine, wont show when I rebase/merge recent main to this PR

rima-builder (approved):
Approving but lets wait for @teleaziz review

gustavocodess (approved):
Looks good, thanks @murtaza-builder

PR #3078, opened 2024-03-27
initial angular gen2 sdk with working build and some blocks

sidmohanty11:
## Description

- Make the build pass
- Text blocks working fine
- Buttons, links works
- 38 e2e tests passing

**Jira**
https://builder-io.atlassian.net/browse/ENG-5520


packages/sdks/overrides/angular/src/constants/target.ts
```diff
@@ -0,0 +1,3 @@
+type Target = import('../types/targets.js').Target;
+/** This file should be overriden for each framework. Ideally this would be implemented in Mitosis.  */
+export const TARGET = 'angular' as Target;
```
samijaber, line 3: one thing missing here is adding `angular` to this `Target` type:

https://github.com/BuilderIO/builder/blob/30f58449160ec2cd30e94e7c36c710e683c43056/packages/sdks/src/types/targets.ts#L1

packages/sdks/src/components/content/content.lite.tsx
```diff
@@ -110,7 +111,7 @@ export default function ContentComponent(props: ContentProps) {
           // eslint-disable-next-line @typescript-eslint/ban-ts-comment
           // @ts-ignore
           reactNative: props.blocksWrapper || ScrollView,
-          default: props.blocksWrapper || 'div',
+          default: props.blocksWrapper || DynamicDiv,
```
samijaber, line 114: Can you explain this workaround? I'm confused by it.

Also, if absolutely necessary, feel free to keep it, but we'll at least want to scope it to just Angular and keeping the default `'div'` for other frameworks

packages/sdks/output/angular/src/index.ts
```diff
@@ -1 +1,15 @@
...
+ * In the React SDK, this file is marked with `use client`.
+ */
+// prettier-ignore
+export * from './index-helpers/blocks-exports';
+
+/**
+ * In the React SDK, this file is not marked with `use client`, to allow
+ * NextJS App Directory to use the SDK helper functions without issues.
+ */
+export * from './server-index'
```
samijaber, line 15: this file shouldn't be tracked either

packages/sdks/overrides/angular/src/functions/evaluate/choose-eval.ts
```diff
@@ -0,0 +1,27 @@
...
+const isBrowser = typeof window !== 'undefined';
+
+/**
+ * Even though we have separate runtimes for browser/node/edge, sometimes frameworks will
+ * end up sending the server runtime code to the browser (most notably in dev mode).
+ */
+export const chooseBrowserOrServerEval = (args: ExecutorArgs) =>
+  isBrowser || shouldForceBrowserRuntimeInNode()
+    ? runInBrowser(args)
+    : evaluator(args);
```
samijaber, line 27: would be good to simplify this override this simplify and make it clear why it's an override:

```suggestion
/** we only have a browser runtime bundle for now */ 
 export const chooseBrowserOrServerEval = runInBrowser;
```

packages/sdks/mitosis.config.cjs
```diff
@@ -198,6 +198,73 @@ const filterActionAttrBindings = (json, item) => {
...
+const ANGULAR_PLUGIN5 = () => ({
+  code: {
+    post: (code) => {
+      if (code.includes('component-ref, ComponentRef')) {
+        code = code.replace('ngOnInit() {\n', `ngOnInit() {\n  this.Wrapper = this.isInteractive ? InteractiveElement : this.componentRef;\n`)
+      }
+      return code;
+    }
+  }
+});
```
samijaber, line 266: Would appreciate a brief comment for each plugin, explaining what the workaround is and why it is needed.

packages/sdks/src/blocks/button/button.lite.tsx
```diff
@@ -39,7 +29,22 @@ export default function Button(props: ButtonProps) {
...
+        reactNative: props.link
+          ? // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
+            props.builderLinkComponent || BaseText
+          : // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
+            BaseText,
+        default: props.link ? props.builderLinkComponent || 'a' : 'button',
+      })}
+      attributes={state.attrs}
```
samijaber, line 47: How come this needed to be moved into its own property? I assume something about Angular not supporting spreads? is there any workaround that avoids having a new piece of state?


 I try to reduce the number of things stored in state because depending on the framework, they might not update/cause re-renders when the original prop changes. So this could cause issues in certain SDKs.

packages/sdks/src/blocks/columns/columns.lite.tsx
```diff
@@ -216,17 +231,7 @@ export default function Columns(props: ColumnProps) {
...
-                reactNative: {
-                  dataSet: { 'builder-block-name': 'builder-column' },
-                },
-                default: {},
-              }),
-              ...(column.link ? { href: column.link } : {}),
-              [getClassPropName()]: 'builder-column',
-              style: mapStyleObjToStrIfNeeded(state.columnCssVars(index)),
-            }}
+            attributes={state.getAttributes(column, index)}
```
samijaber, line 234: same question here

packages/sdks/src/blocks/button/button.lite.tsx
sidmohanty11: Unfortunately angular doesn't support calling functions like this with [fn()] and also object spread with ternary doesn't work inside angular template

https://github.com/sidmohanty11/builder/blob/main/packages/sdks/src/blocks/button/button.lite.tsx#L32-L41

sidmohanty11: cc @shyam-builder is there any other workaround?

packages/sdks/overrides/angular/src/functions/evaluate/choose-eval.ts
sidmohanty11: thanks @samijaber!

packages/sdks/mitosis.config.cjs
```diff
@@ -198,6 +198,337 @@ const filterActionAttrBindings = (json, item) => {
...
+
+const ANGULAR_PLUGIN8 = () => ({
+  code: {
+    post: (code) => {
+      code = code.replace('"div"', 'DynamicDiv');
+      return code;
+    },
+  },
+});
+
```
samijaber, line 531: Why don't we add an `angular` key to the `useTarget`s, so that we don't need any of these workaround plugins?

For example in `Columns`:

```diff
+ import DynamicDiv from '../../dynamic-div.lite.tsx`;


         <DynamicRenderer
            key={index}
            TagName={
              column.link
                ? props.builderLinkComponent ||
                  useTarget({
                    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
                    // @ts-ignore
                    reactNative: BaseText,
+                  angular: DynamicDiv,
                    default: 'a',
                  })
                : useTarget({
                    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
                    // @ts-ignore
                    reactNative: View,
+                  angular: DynamicDiv,
                    default: 'div',
                  })
            }
```

samijaber (approved):
Everything looks good to me, just need to:

- see my last few comments
- fix the build so it passes
- add some documentation for the angular plugins for the future

packages/sdks/mitosis.config.cjs
sidmohanty11: yes! this will be best, let me update it

sidmohanty11: thanks!!

packages/sdks/src/components/content/content.lite.tsx
sidmohanty11: As discussed async, this limitation arises because Angular does not directly support HTML tags (strings) as components within ngComponentOutlet (dynamic components), thereby preventing the use of 'div's. DynamicDiv serves as a wrapper div component that renders its children and assigns all provided attributes to it.

packages/sdks/src/blocks/text/text.types.ts
```diff
@@ -1,3 +1,5 @@
 export type TextProps = {
   text?: string;
+  builderBlock?: any;
+  builderContext?: any;
```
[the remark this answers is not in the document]
sidmohanty11: thanks!

packages/sdks/overrides/angular/src/constants/target.ts
sidmohanty11: done!

packages/sdks/mitosis.config.cjs
sidmohanty11: added

samijaber (approved):
🎉

PR #3085, opened 2024-04-03
fix NextJS SDK docs

samijaber:
## Description

_Screenshot_

### DIFF
--- a/packages/sdks/output/nextjs/README.md
+++ b/packages/sdks/output/nextjs/README.md
@@ -10,7 +10,69 @@
 
 This is the Builder NextJS SDK, `@builder.io/sdk-react-nextjs`. It is intended to be used _only_ with NextJS's app directory, and has hard dependencies on NextJS-specific functionality that only works in the app directory.
 
-For usage information, look at the [examples](#examples).
+## Usage
+
+When registering a custom component, you will need to add the `isRSC: true` option to the component. For example:
+
+```tsx
+// CatFacts.tsx
+async function CatFacts() {
+  const catFacts = await fetch('https://cat-fact.herokuapp.com/facts').then(
+    (x) => x.json()
+  );
+  return (
+    <div>
+      Here are some cat facts from an RSC:
+      <ul>
+        {catFacts.slice(3).map((fact) => (
+          <li key={fact._id}>{fact.text}</li>
+        ))}
+      </ul>
+    </div>
+  );
+}
+
+export const CatFactsInfo = {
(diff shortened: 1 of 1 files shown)


packages/sdks/output/nextjs/README.md
```diff
@@ -1,71 +1,16 @@
...
-## API Reference
-
-To use the SDK, you need to:
-
-- fetch the builder data using `getContent`: you can see how to use it here https://www.builder.io/c/docs/content-api, and how it differs from the React V1 SDK's `builder.get()` function.
-
-- pass that data to the `RenderContent` component, along with the following properties:
+- you should ONLY use this SDK if you are trying to [register](https://www.builder.io/c/docs/custom-components-setup) your RSCs (react server components) in Builder. That is its only advantage over our standard React SDKs.
+- our [Gen1](../../../react/) and [Gen2](../react/) React SDKs work perfectly well with all versions of Next.js. The only feature they do not support is registration of RSCs.
+- to allow registering RSCs, this SDK must make compromises. Most notably, it does not support interactivity within the rendered content. See the [features grid](https://github.com/BuilderIO/builder/tree/main/packages/sdks#features) for more information.
```
steve8708, line 5: this makes it sound like nothing can be interactive, but you can use client component sthat are interactive in hte editor. just doesn't allow interactivity created direclty in builder, which is kind of a rare/power feature anyway 

I think the other thing to call out is the editor will be noticably more laggy, because we have to do round trips to the server for many of the updates

packages/sdks/output/nextjs/README.md
```diff
@@ -1,71 +1,16 @@
...
+- this SDK only works in the NextJS App Directory.
+- this SDK is marked as "Beta" due to the missing features mentioned above. As of this time, there are no solutions to acheiving interactivity in this SDK. It is however actively maintained and developed alongside all other SDKs.
```
steve8708, line 7: similar note on interactivity - you can def have interactive components, just not interactions defined in builder via [state and actions](https://www.builder.io/c/docs/guides/state-and-actions)

packages/sdks/output/nextjs/README.md
```diff
@@ -1,71 +1,16 @@
-# Builder.io React NextJS SDK (EXPERIMENTAL)
+⚠️: Please read the below carefully.
```
steve8708, line 1: current message sounds a bit intense, one suggestion:

```suggestion
## When should I use this SDK (please read carefully)
```

PR #3181, opened 2024-04-08
[fix] hover animations in gen2

### LINKED ISSUE #2889: Animations not supported in Svelte SDK
**Describe the bug**
When setting the hover styles the hover styles do not apply.

**To Reproduce**
See attached video. (I am using the Sveltekit)
https://github.com/BuilderIO/builder/assets/12568665/28686b95-98cc-4a2d-a7ff-71be6fa7ce03

**Expected behavior**
There should be hover styles.

**Screenshots**
See video

sidmohanty11:
## Description

This PR adds hover animation support for Gen2 SDKs

**Jira**
https://builder-io.atlassian.net/browse/ENG-5278

**Loom**
https://www.loom.com/share/1928768351e64d249af641a964062934

Closes #2889

### DIFF
--- a/.changeset/sixty-pugs-hide.md
+++ b/.changeset/sixty-pugs-hide.md
@@ -1,0 +1,11 @@
+---
+'@builder.io/sdk-react-nextjs': patch
+'@builder.io/sdk-qwik': patch
+'@builder.io/sdk-react': patch
+'@builder.io/sdk-react-native': patch
+'@builder.io/sdk-solid': patch
+'@builder.io/sdk-svelte': patch
+'@builder.io/sdk-vue': patch
+---
+
+Fix: hover animations
--- a/.changeset/sixty-pugs-hide.md
+++ b/.changeset/sixty-pugs-hide.md
@@ -8,4 +8,4 @@
 '@builder.io/sdk-vue': patch
 ---
 
-Fix: hover animations
+Feature: add support for hover animations.
--- a/packages/sdks-tests/src/tests/hover-animation.spec.ts
+++ b/packages/sdks-tests/src/tests/hover-animation.spec.ts
@@ -9,18 +9,11 @@ test.describe('Hover animations', () => {
 
     const button = isOldReactSDK ? page.locator('span') : page.locator('button');
 
-    const initialColor = await button.evaluate(el => getComputedStyle(el).backgroundColor, null, {
-      timeout: 10000,
-    });
+    await expect(button).toHaveCSS('background-color', 'rgb(0, 0, 0)');
 
     await button.hover();
     await page.waitForTimeout(1000);
 
-    const hoveredColor = await button.evaluate(el => getComputedStyle(el).backgroundColor, null, {
-      timeout: 10000,
-    });
-
-    expect(initialColor).toContain('rgb(0, 0, 0)');
-    expect(hoveredColor).not.toContain(initialColor);
+    await expect(button).toHaveCSS('background-color', 'rgb(149, 79, 79)');
   });
 });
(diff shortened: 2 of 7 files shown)


packages/sdks-tests/src/tests/hover-animation.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+  test.fail(excludeTestFor({ angular: true }), 'Angular Gen2 SDK not implemented.');
+  test.fail(isRNSDK); // hover animation is not handled in react native SDK
+  test('Button should change color when hovered', async ({ page }) => {
+    await page.goto('/hover-animation');
+
+    const button = isOldReactSDK ? page.locator('span') : page.locator('button');
+
+    const initialColor = await button.evaluate(el => getComputedStyle(el).backgroundColor, null, {
+      timeout: 10000,
+    });
```
samijaber, line 14: I used to get styles this way but it caused issues. Also, not a fan of all the `timeout`s in this test.

Have you tried `await expect(button).toHaveCSS(key, value)`? That's that we use everywhere now and it works OK

packages/sdks/src/components/block/components/block-styles.lite.tsx
```diff
@@ -96,7 +97,34 @@ export default function BlockStyles(props: BlockStylesProps) {
...
+            },
+          }) || '';
+      }
+
+      return [
+        largeStylesClass,
+        mediumStylesClass,
+        smallStylesClass,
+        hoverStylesClass,
+      ].join(' ');
```
samijaber, line 127: Why do we need to apply the hover at this point? Isn't the hover animation meant to be applied in the Animator class here:

https://github.com/BuilderIO/builder/blob/c58bc3be910b47037bed8c99ea463cde07dc7d5e/packages/sdks/src/components/block/animator.ts#L159-L195

https://github.com/BuilderIO/builder/blob/c58bc3be910b47037bed8c99ea463cde07dc7d5e/packages/sdks/src/components/block/block.lite.tsx#L173-L186

sidmohanty11: we are filtering out hover animations here so its not getting applied.

I followed this (gen1 code) for reference, we are setting hover animations in every block level: https://github.com/BuilderIO/builder/blob/c58bc3be910b47037bed8c99ea463cde07dc7d5e/packages/react/src/components/builder-block.component.tsx#L176

packages/sdks-tests/src/tests/hover-animation.spec.ts
sidmohanty11: Thanks Sami! Since this is a hover effect, I thought we might want to use a computed style to get the actual color. Let me use `toHaveCss` and try.

sidmohanty11: Thanks a lot Sami! I've updated the tests to use `toHaveCSS` and removed the timeouts

.changeset/sixty-pugs-hide.md
```diff
@@ -0,0 +1,11 @@
...
+'@builder.io/sdk-react-nextjs': patch
+'@builder.io/sdk-qwik': patch
+'@builder.io/sdk-react': patch
+'@builder.io/sdk-react-native': patch
+'@builder.io/sdk-solid': patch
+'@builder.io/sdk-svelte': patch
+'@builder.io/sdk-vue': patch
+---
+
+Fix: hover animations
```
samijaber, line 11:
```suggestion
Feature: add support for hover animations.
```

samijaber (approved):

PR #3196, opened 2024-04-15
Add a content menu action to allow removing content from smartling translation job

murtaza-builder:
## Description

To remove the content entry **cleanly** from the Smartling's local translation job, added a content menu action for the content that helps in removing job details from content and also removing an entry from the local translation job.

_Screenshot_

snapshot

### DIFF
--- a/plugins/smartling/README.md
+++ b/plugins/smartling/README.md
@@ -2,7 +2,7 @@
 
 ## Installation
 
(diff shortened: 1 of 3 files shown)


shyam-builder (approved):
I'm assuming the API needs to get merged and deployed before this can be released?

PR #3205, opened 2024-04-16
fix: svelte styling to apply during ssr

### LINKED ISSUE #3190: Svelte: critical usability issue: styles do not apply during server-side-rendering due to `use:mitosis_styling` directive
**Describe the bug**

My UX and CLS scores using builder.io/mitosis with svelte are terrible
﻿
All of the following block types make use of `use:mitosis_styling`, which is a client-side directive that relies on the DOM nodes. This means that barely any styles from server-side rendering in my sveltekit app are sent to the client, leading to large layout flashes
- Columns
- Images
- Videos
- Blocks-wrapper
- Section
- img
- BaseText
- Slot

**To Reproduce**
1. Use any of the layouts in SSR, none of the styling is used in SSR mode
2. They are only mounted on the client-side

**Expected behavior**
Currently, mitosis makes use of `use:mitosis_styling` `use` svelte directives, which only run on client-side as they rely on the nodes
```ts
  function mitosis_styling(node, vars) {
    Object.entries(vars || {}).forEach(([p, v]) => {
      if (p.startsWith("--")) {
        node.style.setProperty(p, v);
      } else {
        node.style[p] = v;
      }
    });
  }
```
```html
<img
  use:mitosis_styling={{
      objectPosition: backgroundPosition || "center",
      objectFit: backgroundSize || "cover",
      ...aspectRatioCss(),
    }}
  .../>
```

In svelte, we instead do use [**style directives**](https://svelte.dev/tutorial/style-directive) or style attributes:
```html
<img
  style="color: red"
  style:height="100%"
  style:background-color="red"
  .../>
```

**Screenshots**
> **Server-side HTML response** - notice lack of any styling
> image

> **Client-side mount** - all styles appear
> image

**Additional context**

sidmohanty11:
## Description

This PR updates the mitosis to the latest version where we've added the fix for svelte styling to apply during SSR.

Ref Mitosis PR: https://github.com/BuilderIO/mitosis/pull/1393

_Screenshot_

**Jira**
https://builder-io.atlassian.net/browse/ENG-5640

Closes #3190

### DIFF
--- a/.changeset/nice-laws-notice.md
+++ b/.changeset/nice-laws-notice.md
@@ -2,4 +2,4 @@
 '@builder.io/sdk-svelte': patch
 ---
 
-fix: svelte styling to apply during ssr
+Fix: Blocks styles not applying during SSR.
--- a/.changeset/nice-laws-notice.md
+++ b/.changeset/nice-laws-notice.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/sdk-svelte': patch
+---
+
+fix: svelte styling to apply during ssr
--- a/packages/sdks-tests/src/tests/css-properties.spec.ts
+++ b/packages/sdks-tests/src/tests/css-properties.spec.ts
@@ -1,7 +1,7 @@
 import { expect } from '@playwright/test';
-import { excludeTestFor, test } from './helpers/index.js';
+import { excludeTestFor, isSSRFramework, test } from './helpers/index.js';
 
-test.describe('CSS Properties from Builder Content', () => {
+test.describe('CSS Properties from Builder Content (js enabled)', () => {
   test.fail(excludeTestFor({ angular: true }), 'Angular Gen2 SDK not implemented.');
   test('set image width CSS properties correctly', async ({ page, packageName }) => {
     test.skip(packageName === 'react-native');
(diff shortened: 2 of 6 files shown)


packages/sdks-tests/src/tests/css-properties.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { expect } from '@playwright/test';
+import { excludeTestFor, test } from './helpers/index.js';
+
+test.describe('CSS Properties from Builder Content', () => {
```
samijaber, line 4: could we have this same test rerun once like this, and once with Javascript disabled? that way we can test the styles coming from SSR.

sidmohanty11: yeah that would be awesome! thanks for this suggestion! Should only ssr frameworks run those tests (with js disabled) because client side frameworks will still need js right?

samijaber: correct!

.changeset/nice-laws-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/sdk-svelte': patch
+---
+
+fix: svelte styling to apply during ssr
```
samijaber, line 5:
```suggestion
Fix: Blocks styles not applying during SSR.
```

samijaber (approved):

PR #3215, opened 2024-04-19
[new] gen2 tabs block

sidmohanty11:
## Description

This PR adds tabs block in gen2 SDKs

**Jira**
https://builder-io.atlassian.net/browse/ENG-3856

_Screenshot_

### DIFF
--- a/.changeset/fluffy-berries-sort.md
+++ b/.changeset/fluffy-berries-sort.md
@@ -1,5 +1,4 @@
 ---
-'@builder.io/sdk-react-nextjs': patch
 '@builder.io/sdk-qwik': patch
 '@builder.io/sdk-react': patch
 '@builder.io/sdk-react-native': patch
--- a/.changeset/fluffy-berries-sort.md
+++ b/.changeset/fluffy-berries-sort.md
@@ -8,4 +8,4 @@
 '@builder.io/sdk-vue': patch
 ---
 
-feat: add Builder tabs block
+feature: add the Builder Tabs block (ported from gen1 widgets).
--- a/.changeset/fluffy-berries-sort.md
+++ b/.changeset/fluffy-berries-sort.md
@@ -1,0 +1,11 @@
+---
+'@builder.io/sdk-react-nextjs': patch
+'@builder.io/sdk-qwik': patch
+'@builder.io/sdk-react': patch
+'@builder.io/sdk-react-native': patch
+'@builder.io/sdk-solid': patch
+'@builder.io/sdk-svelte': patch
+'@builder.io/sdk-vue': patch
+---
+
+feat: add Builder tabs block
--- a/packages/sdks/src/blocks/tabs/component-info.ts
+++ b/packages/sdks/src/blocks/tabs/component-info.ts
@@ -1,0 +1,154 @@
+const defaultTab = {
+  '@type': '@builder.io/sdk:Element',
+  responsiveStyles: {
+    large: {
+      paddingLeft: '20px',
+      paddingRight: '20px',
+      paddingTop: '10px',
+      paddingBottom: '10px',
+      minWidth: '100px',
+      textAlign: 'center',
+      // TODO: add to all
+      display: 'flex',
+      flexDirection: 'column',
+      cursor: 'pointer',
+      userSelect: 'none',
+    },
+  },
+  component: {
+    // Builder:text
+    name: 'Text',
+    options: {
+      text: 'New tab',
+    },
+  },
+};
+
+const defaultElement = {
+  '@type': '@builder.io/sdk:Element',
+  responsiveStyles: {
+    large: {
+      height: '200px',
+      display: 'flex',
+      marginTop: '20px',
+      flexDirection: 'column',
+    },
+  },
+  component: {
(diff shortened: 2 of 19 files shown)


packages/sdks/src/blocks/tabs/component-info.ts
```diff
@@ -0,0 +1,152 @@
...
+  },
+  component: {
+    name: 'Text',
+    options: {
+      text: 'New tab content ',
+    },
+  },
+};
+
+export const componentInfo: any = {
```
samijaber, line 43: can we use the `ComponentInfo` type here?

sidmohanty11: yes! thanks!!

sidmohanty11:
https://github.com/BuilderIO/mitosis/pull/1403 will help us unblock the tabs block for React Native

packages/sdks/src/blocks/tabs/tabs.lite.tsx
```diff
@@ -0,0 +1,77 @@
+import { For, Show, useStore, useTarget } from '@builder.io/mitosis';
+import Blocks from '../../components/blocks/blocks.lite.jsx';
+import type { BuilderBlock } from '../../types/builder-block.js';
+import type { TabsProps } from './tabs.types.js';
+
+export default function Tabs(props: TabsProps) {
+  const state = useStore({
+    activeTab: props.defaultActiveTab ? props.defaultActiveTab - 1 : 0,
```
samijaber, line 8: this should be a `getter` so that it reacts to props changes (like changes in the visual editor)

packages/sdks/src/components/block/block.helpers.ts
```diff
@@ -108,3 +108,12 @@ export const getInheritedStyles = ({
   }
   return extractTextStyles(style);
 };
+
+export const shouldPassBuilderData = (blockName: string | undefined) => {
```
samijaber, line 112: good call!

packages/sdks/src/blocks/tabs/tabs.lite.tsx
sidmohanty11: @samijaber sure! if we turn it to a getter should we have another state for whats active, because we are updating it when we click on different tabs and a getter is a read-only property

samijaber: oh right. nevermind then, it's good as-is

packages/sdks/src/constants/extra-components.ts
```diff
@@ -33,6 +35,7 @@ export const getExtraComponents: () => RegisteredComponent[] = () => [
         { component: FormInput, ...formInputComponentInfo },
         { component: FormSubmitButton, ...formSubmitButtonComponentInfo },
         { component: FormSelect, ...formSelectComponentInfo },
+        { component: Tabs, ...tabsComponentInfo },
```
samijaber, line 38: unfortunately `extra-components` is specifically to skip those components from React Native.

So by moving Tabs here it no longer gets included in RN SDK.

I think we should keep it in the original `builder-registered-components.ts` and add a `TARGET === 'rsc'` check there.

packages/sdks/src/constants/builder-registered-components.ts
```diff
@@ -31,5 +33,6 @@ export const getDefaultRegisteredComponents: () => RegisteredComponent[] =
     { component: Slot, ...slotComponentInfo },
     { component: Symbol, ...symbolComponentInfo },
     { component: Text, ...textComponentInfo },
+    { component: Tabs, ...tabsComponentInfo },
```
samijaber, line 36: so we do not need to exclude this when `TARGET === 'rsc'` after all?

samijaber (approved):

packages/sdks/src/constants/builder-registered-components.ts
sidmohanty11: yup! although it still doesn't work because its a client component rendering `<Blocks>`, but no need to exclude as its not breaking or interfering with other tests - what do you think?

samijaber: Let's exclude it. Otherwise customers will see it in their visual editor and try to use it despite it being broken.

sidmohanty11: Makes sense! Will update it

.changeset/fluffy-berries-sort.md
```diff
@@ -0,0 +1,11 @@
...
+'@builder.io/sdk-react-nextjs': patch
+'@builder.io/sdk-qwik': patch
+'@builder.io/sdk-react': patch
+'@builder.io/sdk-react-native': patch
+'@builder.io/sdk-solid': patch
+'@builder.io/sdk-svelte': patch
+'@builder.io/sdk-vue': patch
+---
+
+feat: add Builder tabs block
```
samijaber, line 11:
```suggestion
feature: add the Builder Tabs block (ported from gen1 widgets).
```

samijaber (approved):

PR #3237, opened 2024-04-24
Fix: blocks in angular sdk and e2e setup

sidmohanty11:
## Description

This PR:
- adds `angular-ssr` setup for e2e
- fixes block props issue (image, video, text etc)
- allows visual editing in angular

**Loom**
https://www.loom.com/share/c6996b8f4adc4e248ffb8d68089b250d


packages/sdks/mitosis.config.cjs
```diff
@@ -495,6 +495,72 @@ const ANGULAR_COMPONENT_NAMES_HAVING_HTML_TAG_NAMES = () => ({
...
+              .join(', ')}];
+            properties.forEach(prop => {
+              if (this.wrapperProps && Object.prototype.hasOwnProperty.call(this.wrapperProps, prop)) {
+                this[prop] = this.wrapperProps[prop];
+              }
+            });
+          }
+          ${code.slice(lastEndIndex)}`;
+        }
+      }
```
samijaber, line 558: can we:

- document why this is needed (we're applying properties on component initialization? why not before/elsewhere?)
- remove the duplication, something like:


```suggestion
  const propInitStr = `
              const properties = [${inputs
                .map((input) => `'${input}'`)
                .join(', ')}];
              properties.forEach(prop => {
                if (this.wrapperProps && Object.prototype.hasOwnProperty.call(this.wrapperProps, prop)) {
                  this[prop] = this.wrapperProps[prop];
                }
              });
  `
        if (code.includes('ngOnInit')) {
          code = code.replace(
            'ngOnInit() {',
            `ngOnInit() {${propInitStr}`
          );
        } else {
          const lastEndIndex = code.lastIndexOf('}');
          code = `${code.slice(0, lastEndIndex)}ngOnInit() {
            ${propInitStr}
          ${code.slice(lastEndIndex)}`;
        }
      }
```

samijaber: Thanks for the explanation. Can you add this as a multi-line comment for the plugin for posterity?

sidmohanty11: sure! added!

packages/sdks-tests/src/specs/index.ts
```diff
@@ -143,23 +143,25 @@ export const getProps = async (args: {
   fetchOneEntry?: (opts: any) => Promise<BuilderContent | null>;
   options?: any;
   data?: 'real' | 'mock';
+  apiKey?: string;
```
samijaber, line 146: oh nice, good call on making this more flexible 👍🏽

samijaber (approved):

packages/sdks/src/blocks/tabs/tabs.lite.tsx
```diff
@@ -35,17 +47,10 @@ export default function Tabs(props: TabsProps) {
...
               style={state.getActiveTabStyle(index)}
-              onClick={() => {
-                if (index === state.activeTab && props.collapsible) {
-                  state.activeTab = -1;
-                } else {
-                  state.activeTab = index;
-                }
-              }}
+              // eslint-disable-next-line @typescript-eslint/no-unused-vars
+              onClick={(event: any) => state.onClick(index)}
```
samijaber, line 53: this seems to break in svelte now

sidmohanty11: ahh yes, let me look into this

samijaber (approved):

PR #3250, opened 2024-04-30
fix: `query` option in gen2 sdks to correctly flatten the query object for mongodb

### LINKED ISSUE #3007: `$elemMatch` operator does not work in the new SDK
**Describe the bug**

We are in the process of porting over to the new SDK (which I saw was marked as 1.0.0, nice!)
Unfortunately, the old functionality of the API could not be replicated with the new API with `$elemMatch`.

**To Reproduce**
Not a concrete reproduction but my test script is as follows:
```ts
// @ts-nocheck ignore all issues as this is a script
import { builder } from '@builder.io/sdk';
import { fetchEntries } from '@builder.io/sdk-react';

builder.init(process.env.NX_BUILDER_IO_PUBLIC_KEY);

const callBuilder = async () => {
  const results = await builder.getAll('blog-article', {
    options: {
      sort: {
        createdDate: -1,
      },
      enrich: true,
    },
    query: {
      'data.categories': {
        $elemMatch: {
          'category.id': {
            $in: JSON.stringify(['d491820ec5da422199336a73f388f354']).trim(),
          },
        },
      },
    },
    limit: 5,
    omit: 'data.blocks',
  });
  const resultsNew = await fetchEntries({
    model: 'blog-article',
    apiKey: process.env.NX_BUILDER_IO_PUBLIC_KEY!,
    options: {
      sort: {
        createdDate: -1,
      },
      enrich: true,
    },
    query: {
      'data.categories': {
        $elemMatch: {
          'category.id': {
            $in: JSON.stringify(['d491820ec5da422199336a73f388f354']).trim(),
          },
        },
      },
    },
    limit: 5,
    omit: 'data.blocks',
  });

  const pages = results.map((res) => res.data.title);
  pages.forEach((article) => console.log(article));

  if (results.length && results[0]) {
    console.log(Object.keys(results[0].data));
    console.log(results[0].data.categories);
  }

  console.log('NEW -------------');

  const pagesNew = resultsNew.map((res) => res.data.title);
  pagesNew.forEach((article) => console.log(article));

  if (resultsNew.length && resultsNew[0]) {
    console.log(Object.keys(resultsNew[0].data));
    console.log(resultsNew[0].data.categories);
  }
};

callBuilder();

```

**Expected behavior**
The return in the repo test script should be the same results but the new API returns no results.
It would be helpful to have a debug option attached to the builder API client to output some additional information such as the query keys being passed to the request or something.

**Screenshots**

**Additional context**

sidmohanty11:
## Description

This PR updates the query option to use `flattenMongoQuery` which is basically flatten but when we get an `$key` we just return the object as is as, so that it properly gets parsed in our API (mongo query).

**Jira**
https://builder-io.atlassian.net/browse/ENG-5268

**Loom**
https://www.loom.com/share/10493e4fd2a24b40937590f7f690bff2

Fixes #3007

### DIFF
--- a/packages/sdks/src/helpers/flatten.ts
+++ b/packages/sdks/src/helpers/flatten.ts
@@ -26,6 +26,9 @@ export function flatten<T extends Record<string, any>>(
 /**
  * Flatten a nested MongoDB query object into a flat object with dot-separated keys.
  * $ keys are not flattened and are left as is.
+ *
+ * { foo: { bar: { $gt: 5 }}} -> { 'foo.bar': { '$gt': 5 }}
+ * { foo: {'bar.id': { $elemMatch: { 'baz.id': { $in: ['abc', 'bcd'] }}}}} -> { 'foo.bar.id': { '$elemMatch': { 'baz.id': { '$in': ['abc', 'bcd'] }}}}
  */
 export function flattenMongoQuery(
   obj: any,
--- a/packages/sdks/src/helpers/flatten.ts
+++ b/packages/sdks/src/helpers/flatten.ts
@@ -22,3 +22,29 @@ export function flatten<T extends Record<string, any>>(
       : { ...acc, [newPath]: value };
   }, {} as T);
 }
+
+/**
+ * Flatten a nested MongoDB query object into a flat object with dot-separated keys.
+ * $ keys are not flattened and are left as is.
+ */
+export function flattenMongoQuery(
+  obj: any,
+  _current?: any,
+  _res: any = {}
+): { [key: string]: string } {
+  for (const key in obj) {
+    const value = obj[key];
+    const newKey = _current ? _current + '.' + key : key;
+    if (
+      value &&
+      typeof value === 'object' &&
+      !Array.isArray(value) &&
+      !Object.keys(value).find((item) => item.startsWith('$'))
+    ) {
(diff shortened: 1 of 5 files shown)


packages/sdks/src/helpers/flatten.ts
```diff
@@ -22,3 +22,29 @@ export function flatten<T extends Record<string, any>>(
       : { ...acc, [newPath]: value };
   }, {} as T);
 }
+
+/**
+ * Flatten a nested MongoDB query object into a flat object with dot-separated keys.
+ * $ keys are not flattened and are left as is.
+ */
```
rima-builder, line 29: Could you give some examples in the comment to explain the transformation

samijaber: +1

sidmohanty11: sure! added, please let me know if we should add anything else

Conceptiks:
@samijaber I don't know if this is the right place for it, but after updating to the latest version of the SDK, my query (including `$or` will no longer work: 

```js

$or: [
                {
                    $and: [{ startDate: { $lte: Date.now() } }, { endDate: { $gte: Date.now() } }],
                },
                {
                    $and: [{ startDate: { $exists: false } }, { endDate: { $exists: false } }],
                },
            ],
```

PR #3273, opened 2024-05-09
fix: angular sdk styles

sidmohanty11:
## Description

This PR fixes angular sdk block styles (overall) by marking all the hosts as `display: contents`

**Jira**
https://builder-io.atlassian.net/browse/ENG-5832

_Screenshot_


packages/sdks-tests/src/tests/blocks.spec.ts
```diff
@@ -61,10 +64,31 @@ test.describe('Blocks', () => {
    * We are temporarily skipping this test because it relies on network requests.
    * TO-DO: re-enable it once we have a way to mock network requests.
    */
-  test.skip('Image', async ({ page }) => {
+  test('Image', async ({ page, packageName }) => {
```
samijaber, line 67: Nice!! Can you remove the comment above that says this is temporarily disabled?

packages/sdks-tests/src/tests/helpers/index.ts
```diff
@@ -178,3 +180,6 @@ export const checkIfIsHydrationErrorMessage = (_text: string) => {
 export const getClassSelector = (className: string) => {
   return isRNSDK ? `[data-class*=${className}]` : `.${className}`;
 };
+
+const __filename = fileURLToPath(import.meta.url);
+export const __dirname = path.dirname(__filename);
```
samijaber, line 185: can we name these something other than `__filename` and `__dirname`? Since it's a reserved global variable name in CJS node.js, it can confuse us down the road

packages/sdks-tests/src/tests/basic-styles.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    expect(await leftAlignedText.evaluate(el => getComputedStyle(el).color)).toBe('rgb(255, 0, 0)');
+    expect(await leftAlignedText.evaluate(el => el.getBoundingClientRect().left)).toBeLessThan(10);
+
+    expect(await rightAlignedText.evaluate(el => getComputedStyle(el).color)).toBe(
+      'rgb(0, 24, 255)'
+    );
+    expect(await rightAlignedText.evaluate(el => el.getBoundingClientRect().right)).toBeGreaterThan(
+      1200
+    );
```
samijaber, line 39: Why not `expect(locator).toHaveCSS(key, value)`?

packages/sdks-tests/src/tests/basic-styles.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    page,
+  }) => {
+    const mockImgPath = path.join(__dirname, '..', '..', 'mocks', 'placeholder-img.png');
+    const mockImgBuffer = fs.readFileSync(mockImgPath);
+    await page.goto('/basic-styles');
+
+    const builderBox = await page.waitForSelector(
+      '[builder-id="builder-258a89336ec44f37a38a146e698a858a"]',
+      { timeout: 10000 }
+    );
```
samijaber, line 52: why is this timeout needed? You generally want to avoid arbitrary timeout durations in integration tests, it will cause all sorts of brittleness.

Why do you need to wait for the selector at all? Playwright is meant to take care of waiting until things are ready on its own.

```suggestion
    const builderBox = await page.selector(
      '[builder-id="builder-258a89336ec44f37a38a146e698a858a"]'
    )
```

packages/sdks-tests/src/tests/basic-styles.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+      if (request.url().includes('cdn.builder.io/api/v1/image')) {
+        return route.fulfill({
+          status: 200,
+          contentType: 'image/png',
+          body: mockImgBuffer,
+        });
+      } else {
+        return route.continue();
+      }
+    });
```
samijaber, line 65: `page.route` is meant to be defined before `page.goto`, otherwise you risk not setting up the mock in time.

See https://playwright.dev/docs/api/class-page#page-route, or our other uses of `page.route`

packages/sdks-tests/src/tests/basic-styles.spec.ts
sidmohanty11: the computed value says `0px` margin-left/right so we cannot fully predict if its actually working) -- the alignment

packages/sdks-tests/src/tests/blocks.spec.ts
sidmohanty11: sure!

packages/sdks-tests/src/tests/helpers/index.ts
sidmohanty11: yes! should we name them `current_filename` and `current_dirname` ?

packages/sdks-tests/src/tests/basic-styles.spec.ts
sidmohanty11: @samijaber should we then use `page.locator(...).elementHandle()`?

packages/sdks-tests/src/tests/basic-styles.spec.ts
sidmohanty11: thanks! updated

packages/sdks-tests/src/tests/basic-styles.spec.ts
sidmohanty11: this also doesn't fail correctly and leads to timeout, may need to pass `{ timeout: xxx }` to this

packages/sdks-tests/src/tests/helpers/index.ts
samijaber: sure, `currentFilename` and `currentDirname`

packages/sdks-tests/src/tests/basic-styles.spec.ts
sidmohanty11: wow! thanks for this

packages/sdks-tests/src/tests/basic-styles.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    await page.goto('/basic-styles');
+
+    const builderBoxChildren = page
+      .locator('[builder-id="builder-258a89336ec44f37a38a146e698a858a"]')
+      .locator('div[builder-id]');
```
samijaber, line 68: nice, glad this worked

packages/sdks-tests/src/tests/basic-styles.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+import path from 'path';
+
+test.describe('Basic styles', () => {
+  test('styles should be applied consistently across all SDKs (text and buttons)', async ({
+    page,
+    packageName,
+    sdk,
+  }) => {
+    test.fail(checkIsRN(sdk));
+    test.skip(packageName === 'react-native');
```
samijaber, line 13: you're both skipping and failing react-native 😄. I'd pick just one

samijaber (approved):

packages/sdks-tests/src/tests/basic-styles.spec.ts
sidmohanty11: thanks!!

PR #3308, opened 2024-05-30
fix(sdk): Allow passing readonly arrays

### LINKED ISSUE #3293: TypeScript readonly error
**Describe the bug**
We are trying to create type interfaces for Props to our React components, which requires `as const`, however the `inputs` cannot be `readonly`

**To Reproduce**

``` ts
export const builderColorTheme = {
    name: 'colorTheme',
    type: 'text',
    enum: ['light', 'dark', 'faded'],
    defaultValue: 'light',
} as const;

export const registerPricingTable = {
    name: 'PricingTable',
    noWrap: true,
    inputs: [
        builderColorTheme,
    ],
} as const satisfies Component;
```

**Expected behavior**
Being able to assign `readonly`

**Additional context**
We would like this to be `readonly`
https://github.com/BuilderIO/builder/blob/5c7cd8c57c35ae6209e286e92f3400b2d49d9a0b/packages/sdks/src/types/components.ts#L19

Alexnortung:
## Description

It should be possible to assign readonly types to `Component` and other `@builder.io/sdk` types, since no mutations are(/should) happening.

This is needed to make type inference working in the [`@oak-digital/builder-helpers`](https://github.com/Oak-Digital/builder-helpers) package.

See the following example:

```ts
import Counter from './counter';

const registerCounter = {
    name: 'Counter',
    inputs: [
        {
            name: 'text',
            type: 'string',
        },
    ];
} as const;

Builder.registerComponent(Counter, registerCounter);
```

This example would result in a type error while registering the component to builder. This is not really desired, so this PR makes it possible to ALSO assign readonly types.

This PR fixes #3293

### DIFF
--- a/.changeset/happy-lobsters-cheer.md
+++ b/.changeset/happy-lobsters-cheer.md
@@ -1,0 +1,5 @@
+---
+"@builder.io/sdk": patch
+---
+
+Fix: Mark component types as `readonly`.
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -543,7 +543,7 @@ export interface Input {
   required?: boolean;
   /** @hidden */
   autoFocus?: boolean;
-  subFields?: Input[];
+  subFields?: readonly Input[];
   /**
    * Additional text to render in the UI to give guidance on how to use this
    *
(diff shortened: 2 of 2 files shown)


samijaber (approved):
Thanks a lot for this fix!

samijaber:
I don't have write permissions to this PR, but we need to add a Changesets file so that we can publish your changes properly. Mind merging this PR or providing write permissions so I can make the change? https://github.com/Oak-Digital/builder/pull/1/files#diff-e1ba98944d0ca436baa0fdf8fb51f74d56e420ea215c44ae93a16aac0392cf97

PR #3315, opened 2024-06-04
angular gen2 sdk with new approach

sidmohanty11:
## Description

This PR uses the latest angular generator updates to fix a ton of issues with our angular gen2 SDK like reactivity, symbols, data bindings, style bindings, some custom blocks etc and temporarily overrides video block -> we need to add ref to spread props to a `<video>` tag and spread it onChanges.

**Loom (angular sdk as of now)**
https://www.loom.com/share/10c4b76eae9d48fba36ad0913db95ede


samijaber (approved):
Awesome!

You're missing the changeset @sidmohanty11

PR #3326, opened 2024-06-11
clean up styles on resource picker and refactor out old dependencies

teleaziz:


plugins/salesforce-commerce-api-plugin/package.json
```diff
@@ -86,43 +33,26 @@
...
     "rollup-plugin-replace": "^2.2.0",
     "rollup-plugin-serve": "^1.0.1",
-    "semantic-release": "^21.0.1",
-    "shelljs": "^0.8.3",
-    "travis-deploy-once": "^5.0.9",
-    "ts-jest": "^23.10.2",
-    "ts-node": "^7.0.1",
-    "tslint": "^5.11.0",
-    "tslint-config-prettier": "^1.15.0",
-    "tslint-config-standard": "^8.0.1",
```
samijaber, line 120: That's so useful, thanks for doing it! 

We need to do this for all `plugins/*` that have copy/pasted a huge list of deps that aren't actually being used. they're adding lots of dependabot noise

samijaber (approved):

PR #3328, opened 2024-06-13
[SUPP-451] fix form submissions for qwik sdk

sidmohanty11:
## Description

`event.currentTarget` is always null for qwik, instead we get the content from `event.target` so added a fallback for that. We also need `preventdefault:submit` so that the form content doesn't refresh

**Jira**
https://builder-io.atlassian.net/browse/SUPP-451

_Screenshot_
Screenshot 2024-06-13 at 11 28 44 AM
Screenshot 2024-06-13 at 11 29 05 AM

### DIFF
--- a/.changeset/kind-rabbits-fry.md
+++ b/.changeset/kind-rabbits-fry.md
@@ -1,0 +1,5 @@
+---
+"@builder.io/sdk-qwik": patch
+---
+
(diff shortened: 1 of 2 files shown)


shyam-builder (approved):
LGTM

shyam-builder (approved):

PR #3333, opened 2024-06-14
fix: RN ci e2e tests

sidmohanty11:
## Description

**Jira**
https://builder-io.atlassian.net/browse/ENG-5971

_Screenshot_

### DIFF
--- a/packages/sdks/e2e/angular-ssr/src/app/app.component.ts
+++ b/packages/sdks/e2e/angular-ssr/src/app/app.component.ts
@@ -1,6 +1,7 @@
+import { DOCUMENT } from '@angular/common';
 // fails because type imports cannot be injected
 // eslint-disable-next-line @typescript-eslint/consistent-type-imports
-import { ChangeDetectorRef, Component } from '@angular/core';
+import { ChangeDetectorRef, Component, Inject } from '@angular/core';
 import {
   _processContentResult,
   fetchOneEntry,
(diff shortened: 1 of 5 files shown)


packages/sdks/e2e/react-native/package.json
```diff
@@ -6,9 +6,9 @@
     "android": "expo start --android",
     "ios": "expo start --ios",
     "web": "expo start --web",
-    "build": "echo 'no need to build react-native.'",
+    "build": "npx expo export:web",
     "dev": "yarn run web",
-    "serve": "yarn run web",
+    "serve": "npx serve web-build --single",
```
samijaber, line 11: We should not use `npx` in our commands:
- `npx` will run using `npm` instead of `yarn`, which can lead to surprises
- if a package is not in the dependencies, `npx` will download it on the spot instead of erroring, leading to surprises

sidmohanty11: makes sense! thanks, let me update it

packages/sdks/e2e/angular-ssr/src/app/app.component.ts
```diff
@@ -48,31 +54,34 @@ export class AppComponent {
...
-
-    const builderProps = await getProps({
-      pathname: urlPath,
-      _processContentResult,
-      options: getBuilderSearchParams(
-        new URLSearchParams(window.location.search)
-      ),
-      fetchOneEntry,
-    });
+    if (isPlatformBrowser(this.platformId)) {
```
samijaber, line 63: I see the issue now. This SSR example was never fully working on the server, since it relied on `window`. By adding this browser check here, you're essentially disabling SSR: the `builderProps` are only being set once we have access to the `window`.

We need a way to access this information on the server. Here's what I found from looking around:

- this shows a way to do it in an angular service: https://stackoverflow.com/questions/78417872/how-to-get-query-params-with-angular-17-ssr-on-server-side
- I see this bit about injecting the Request token https://www.npmjs.com/package/@nguniversal/express-engine#using-the-request-and-response

sidmohanty11: you're right!! my bad. I realised it late trying to figure out the `window` issue, let me try these! Thanks!!

samijaber (approved):

PR #3349, opened 2024-06-21
Clean up misleading error message

steve8708:


packages/sdks/src/functions/evaluate/evaluate.ts
```diff
@@ -49,6 +49,7 @@ export function evaluate({
   const args: ExecutorArgs = {
     code: parseCode(code, { isExpression }),
     builder: getBuilderGlobals(),
+    Builder: getBuilderGlobals(),
```
samijaber, line 52: This is already handled here: https://github.com/BuilderIO/builder/blob/d391f3fe36137a579209a464c9cd5a79b8f7f67a/packages/sdks/src/functions/evaluate/helpers.ts#L34-L50

`getFunctionArguments` is called in each `evaluate`:

https://github.com/BuilderIO/builder/blob/d391f3fe36137a579209a464c9cd5a79b8f7f67a/packages/sdks/src/functions/evaluate/edge-runtime/edge-runtime.ts#L42-L47

So this change wouldn't make a difference

steve8708: ah i see, ty will keep digging in

steve8708: updated PR to only update the error message

PR #3350, opened 2024-06-22
[fix][sdk][Embed] ENG-5905 Fix Embed block to support url localizations

murtaza-builder:
## Description
When embed block's url is localized, we need to localize the content field as well which is inserted in the rendered content

### DIFF
--- a/packages/react/src/blocks/Embed.tsx
+++ b/packages/react/src/blocks/Embed.tsx
@@ -80,28 +80,67 @@ export const Embed = withBuilder(EmbedComponent, {
       defaultValue: '',
       helperText: 'e.g. enter a youtube url, google map, etc',
       onChange(options: Map<string, any>) {
-        const url = options.get('url');
+        let url = options.get('url');
+
         if (url) {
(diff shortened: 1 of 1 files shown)


packages/react/src/blocks/Embed.tsx
```diff
@@ -80,28 +80,67 @@ export const Embed = withBuilder(EmbedComponent, {
...
+              Object.entries(localeEntries).map(([locale, url]) => {
+                return fetchHtmlContentForUrl(url).then(
+                  content => (fetchedContent[locale] = content)
+                );
+              })
+            ).then(() => {
               options.set(
                 'content',
-                'There was an error embedding this URL, please try again or another URL'
+                Object.assign(
```
murtaza-builder, line 135: This line supposedly is throwing up with mobx error.
image

@shyam-builder

murtaza-builder: Hi @shyam-builder . Need your thoughts on this one

PR #3360, opened 2024-06-25
Update global.css

mrkoreye:
## Description

Fix typo


teleaziz (approved):

mrkoreye:
@samijaber do you have any idea why the react native tests fail here? This PR only modifies a line of CSS in the examples folder. I have re-ran it a bunch of times, so it does not seem like flakiness.

PR #3374, opened 2024-07-01
fix: check `e.origin` of the message to be a URL first

sidmohanty11:
## Description

URL cannot parse `e.origin` with any values other than the URL protocol so it breaks and gives error in the console. This happened with one of our customers on "Windows" machine -- isn't reproducible in macOS

_Screenshot_
IMG20240701184837
builder io

### DIFF
--- a/.changeset/friendly-grapes-build.md
+++ b/.changeset/friendly-grapes-build.md
@@ -1,0 +1,12 @@
+---
+"@builder.io/sdk-angular": patch
+"@builder.io/sdk-react-nextjs": patch
+"@builder.io/sdk-qwik": patch
+"@builder.io/sdk-react": patch
+"@builder.io/sdk-react-native": patch
+"@builder.io/sdk-solid": patch
+"@builder.io/sdk-svelte": patch
+"@builder.io/sdk-vue": patch
+---
+
+Fix: check `e.origin` of the message to be a URL first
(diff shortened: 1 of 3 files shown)


packages/sdks/src/functions/is-from-trusted-host.ts
```diff
@@ -10,6 +10,9 @@ export function isFromTrustedHost(
   trustedHosts: string[] | undefined,
   e: { origin: string }
 ): boolean {
+  if (!e.origin.includes('http') && !e.origin.includes('https')) {
```
shyam-builder, line 13: why not `startsWith`? and can origin be `null` or `undefined`?

sidmohanty11: thanks! let me update to `startsWith`

Screenshot 2024-07-01 at 8 48 34 PM
must be a string (https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent/origin)

shyam-builder (approved):

.changeset/friendly-grapes-build.md
```diff
@@ -0,0 +1,12 @@
...
+"@builder.io/sdk-react-nextjs": patch
+"@builder.io/sdk-qwik": patch
+"@builder.io/sdk-react": patch
+"@builder.io/sdk-react-native": patch
+"@builder.io/sdk-solid": patch
+"@builder.io/sdk-svelte": patch
+"@builder.io/sdk-vue": patch
+---
+
+Fix: check `e.origin` of the message to be a URL first
```
samijaber, line 12: @sidmohanty11 For the future: this kind of release note is too technical and does not explain what is actually being fixed. It is not useful for SDK consumers.

A better one would be:

```
Fix: crashes during editing caused by invalid host origin.
```

PR #3377, opened 2024-07-03
Added config for disabling Smartling plugin auto-authorization

murtaza-builder:
## Description
Added config (for admins only) - "Authorize Smartling Jobs through Builder" - to disable auto-authorization of Smartling translation jobs. Its enabled by default.

_Screenshot_
image

### DIFF
--- a/plugins/smartling/src/plugin.tsx
+++ b/plugins/smartling/src/plugin.tsx
@@ -30,6 +30,33 @@ import stringify from 'fast-json-stable-stringify';
 // translation status that indicate the content is being queued for translations
 const enabledTranslationStatuses = ['pending', 'local'];
 
+function updatePublishCTA(content: any, translationModel: any) {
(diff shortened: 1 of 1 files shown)


plugins/smartling/src/plugin.tsx
```diff
@@ -108,7 +151,7 @@ registerPlugin(
                 appState.globalState.showGlobalBlockingLoading('Contacting Smartling ....');
                 await api.updateTranslationFile({
                   translationJobId: lastPublishedContent.meta.translationJobId,
-                  translationModel: getTranslationModel().name,
+                  translationModel: translationModel.name,
```
rima-builder, line 154: what if getTranslationmodel returns null ?

murtaza-builder: rectified.

rima-builder (approved):

PR #3383, opened 2024-07-08
[Plugin][Utils][Smartling][Phrase] ENG-5718 Skip 'children' property for symbol inputs which contaings component defs

murtaza-builder:
## Description
Skip `data.children` in symbol, when resolving complex symbol inputs which may contain component defs (that may come to render when 'useChildren' symbol prop is set) in 'data.children'. This is to skip double translations and avoid unpredictable results.

For the translations required in component defs of `data.children`, let the regular `traverse` cycle handle it.

_Screenshot_
image

image

### DIFF
--- a/packages/utils/src/translation-helpers.ts
+++ b/packages/utils/src/translation-helpers.ts
@@ -199,6 +199,10 @@ export function getTranslateableFields(
               symbolInputName: string,
               symbolInputValue: any
             ]) => {
+              if (symbolInputName === 'children') {
+                return;
+              }
+
               recordValue({
                 results,
                 value: symbolInputValue,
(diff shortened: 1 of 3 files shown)


packages/utils/src/translation-helpers.ts
```diff
@@ -199,6 +199,10 @@ export function getTranslateableFields(
               symbolInputName: string,
               symbolInputValue: any
             ]) => {
+              if (symbolInputName === 'children') {
```
murtaza-builder, line 202: this is the key change.

murtaza-builder: @teleaziz @shyam-builder

teleaziz (approved):

PR #3390, opened 2024-07-08
use style tag for solidjs to fix styling issues

steve8708:
this depends on https://github.com/BuilderIO/mitosis/pull/1494 PR to be merged and released, then tests should pass

### DIFF
--- a/.changeset/many-ligers-compete.md
+++ b/.changeset/many-ligers-compete.md
@@ -1,6 +1,5 @@
 ---
-"@builder.io/packages": patch
-"@builder.io/sdks": patch
+"@builder.io/sdk-solid": patch
 ---
 
 Use style tags for solidjs instead of solid-styled-components to fix CLS issues
--- a/.changeset/many-ligers-compete.md
+++ b/.changeset/many-ligers-compete.md
@@ -2,4 +2,4 @@
 "@builder.io/sdk-solid": patch
 ---
 
-Use style tags for solidjs instead of solid-styled-components to fix CLS issues
+Fix: use inlined `<style>` tags instead of `solid-styled-components` to resolve CLS issues.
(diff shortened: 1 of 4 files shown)


steve8708:
this depends on [this](https://github.com/BuilderIO/mitosis/pull/1494) PR to be merged and released, then tests should pass

samijaber (approved):
I don't see how this is dependent on https://github.com/BuilderIO/mitosis/pull/1494, since the mitosis PR is merely changing the update which we are overriding here

steve8708:
@samijaber mitosis PR also removes `jsx` from the `style` tag which is causes checks to fail here as `jsx` doesn't exist on the `style` element in solidjs's TS types

samijaber:
Thanks for clarifying, I missed this small change in the other PR. makes sense!

.changeset/many-ligers-compete.md
```diff
@@ -0,0 +1,5 @@
+---
+"@builder.io/sdk-solid": patch
+---
+
+Use style tags for solidjs instead of solid-styled-components to fix CLS issues
```
samijaber, line 5:
```suggestion
Fix: use inlined `<style>` tags instead of `solid-styled-components` to resolve CLS issues.
```

samijaber (approved):

packages/sdks/package.json
```diff
@@ -63,7 +63,7 @@
     "upgrade-example:all": "yarn loop upgrade-example latest"
   },
   "dependencies": {
-    "@builder.io/mitosis": "^0.3.8",
+    "@builder.io/mitosis": "latest",
```
samijaber, line 66: I prefer a fixed or carat version instead of `latest` here. I know we maintain mitosis, but i worry that `latest` might cause more frequent updates to mitosis than necessary.

steve8708: yeah that was a mistake i just caught as well, `ni @builder.io/mitosis@latest` i always forget does this for somereason (whereas using yarn/npm directly doens't)

samijaber (approved):

PR #3392, opened 2024-07-09
support CSS variables using default values in react native

steve8708:
that way design tokens and figma import will work for react native

relates to https://github.com/BuilderIO/builder-internal/pull/6510


packages/sdks/overrides/react-native/src/functions/transform-style-property.ts
```diff
@@ -166,13 +166,28 @@ function omit<T extends object>(obj: T, ...values: (keyof T)[]): Partial<T> {
...
+
+  // Function to replace var() with its fallback
+  let newValue = value;
+  let match: string[] | null;
+  while ((match = newValue.match(varRegex))) {
+    newValue = newValue.replace(match[0], match[1].trim());
+  }
+
+  return newValue;
+}
```
samijaber, line 181: Can we add some tests for this regex `while` loop? 😅

I know this is in an `overrides`, so it might have to move to packages/sdks/src/functions/extract-var-default-value.ts` so that it can be tested.

.changeset/empty-drinks-nail.md
```diff
@@ -0,0 +1,6 @@
+---
+"@builder.io/packages": patch
```
samijaber, line 2: this is wrong and should list each SDK

.changeset/empty-drinks-nail.md
```diff
@@ -0,0 +1,6 @@
...
+"@builder.io/sdks": patch
```
samijaber, line 3: this is wrong and snould list 
```suggestion
"@builder.io/sdk-react-native": patch
```

.changeset/empty-drinks-nail.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+Support CSS variables that use default values in React Native
```
samijaber, line 6:
```suggestion
Fix: Add support for CSS variables that use default values.
```

packages/sdks/overrides/react-native/src/functions/transform-style-property.ts
steve8708: ah yeah agree on tests, and on the issues with overrides being why i didn't add it. i hate the idea of react native specific code bloating the other SDKs though, but I guess no better way. its copy pasta'd from webapp so we know is working well already, but will add tests to guard from future fixes breaking things

samijaber: yup, I know what you mean about bloat.

But since:
- the file won't be imported by anything else in other SDK
- all SDKs have a bundling step


therefore it won't end up in bundle output of any other SDK. So this is safe to do 👍🏽

steve8708: yup, updated. at first i misread your comment to move that into an existing file vs its own, its in its own with tests now

samijaber (approved):

PR #3396, opened 2024-07-09
fix(react-gen1): fix access to getters and functions on arrays in ser…

teleaziz:
…ver context

### DIFF
--- a/.changeset/forty-mangos-brush.md
+++ b/.changeset/forty-mangos-brush.md
@@ -1,0 +1,5 @@
+---
+"@builder.io/react": patch
+---
+
+Fix binding to array property getters and methods in server context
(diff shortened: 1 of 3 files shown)


samijaber:
@teleaziz you need to add changesets to publish your SDK changes: https://github.com/BuilderIO/builder/blob/main/packages/sdks/PUBLISHING.md#1--add-changeset

mrkoreye (commented):
I don't understand the change here. Can you add a comment on this PR to explain the reason for the change a little bit more?

teleaziz:
@mrkoreye this change is to return the passed objects to its original shape as isolated-vm change them automatically to ivm.refrence which prevents users from accessing getters and functions on the original prototype like binding to `state.arr.length` or `state.arr.join('-')` etc

mrkoreye (approved):

PR #3398, opened 2024-07-10
multi-build support for angular gen2 sdk

sidmohanty11:
## Description

Creates two builds (node and browser) and exports them from `/lib`

Jira
https://builder-io.atlassian.net/browse/ENG-5583

_Screenshot_


packages/sdks/output/angular/scripts/multi-build.mjs
```diff
@@ -0,0 +1,103 @@
...
+      },
+    ],
+  });
+
+  fs.writeFileSync(filePath, result.code, 'utf-8');
+};
+
+// Transform the choose-eval.ts file to replace 'placeholder-runtime'
+// to appropriate runtime value based on the SDK_ENV
+const folderName = `./${getFolderName()}/index`;
```
samijaber, line 89: you should be able to reuse the logic from `@builder.io/sdks/output-generation`. check how we do that in the other SDKs:

CleanShot 2024-07-11 at 09 37 41@2x

this should remove a bunch of logic from this file

packages/sdks/output/angular/scripts/multi-build.mjs
```diff
@@ -0,0 +1,103 @@
...
+  'utf-8'
+);
+
+const foundFiles = glob.sync('src/**/*/choose-eval.*');
+
+if (foundFiles.length !== 1) {
+  throw new Error(
+    'Expected exactly one file named `choose-eval.ts` containing "placeholder-runtime" import.'
+  );
+}
```
samijaber, line 52: you should add a big loud comment at the top of `choose-eval.ts` saying something like:

```
/*
 * THIS FILE IS EXPECTED TO EXIST WITH THIS EXACT NAME BY SUBSEQUENT BUNDLE STEPS THAT TRANSFORM IT.
 * DO NOT RENAME THIS FILE.
 */
```

packages/sdks/output/angular/scripts/multi-build.mjs
```diff
@@ -0,0 +1,103 @@
...
+              const newValue = revert ? 'placeholder-runtime' : replaceValue;
+              if (path.node.source.value === currentValue) {
+                path.node.source.value = newValue;
+              }
+            },
+          },
+        };
+      },
+    ],
+  });
```
samijaber, line 82: I would also add a check that the import was successfully renamed
```suggestion
  let hasReplacedImport = false;
  
  const result = babel.transformFileSync(filePath, {
    plugins: [
      tsPlugin,
      /** @returns {InlinePlugin} */
      () => {
        return {
          visitor: {
            ImportDeclaration(path) {
              const currentValue = revert
                ? replaceValue
                : 'placeholder-runtime';
              const newValue = revert ? 'placeholder-runtime' : replaceValue;
              if (path.node.source.value === currentValue) {
                path.node.source.value = newValue;
                hasReplacedImport = true
              }
            },
          },
        };
      },
    ],
  });
  
  if (!hasReplacedImport) throw new Error(`could not find import '${currentValue}'`);
```

packages/sdks/output/angular/scripts/multi-build.mjs
sidmohanty11: thank you, yes I've now re-used most of the functions from this file

samijaber: You should be able to use `getEvaluatorPathAlias` here

```suggestion
// Transform the choose-eval.ts file to replace 'placeholder-runtime'
// to appropriate runtime value based on the SDK_ENV
const folderName = getEvaluatorPathAlias({ format: 'js' });
```

packages/sdks/output/angular/scripts/multi-build.mjs
sidmohanty11: done

packages/sdks/output/angular/scripts/multi-build.mjs
sidmohanty11: yes a necessary check, added

packages/sdks/output/angular/scripts/multi-build.mjs
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * @typedef {import('@babel/core').PluginObj} babel.PluginObj
+ * @typedef {babel.PluginObj} InlinePlugin
+ */
+
+const sdkEnv = getSdkEnv();
+
+if (!sdkEnv) {
+  throw new Error('SDK_ENV is required to build the SDK.');
+}
```
samijaber, line 23:
```suggestion
```

packages/sdks/output/angular/scripts/multi-build.mjs
```diff
@@ -0,0 +1,109 @@
...
+const ngPackageJsonPath = path.resolve('ng-package.json');
+const ngPackageJson = JSON.parse(fs.readFileSync(ngPackageJsonPath, 'utf-8'));
+
+const outputPath = getSdkOutputPath();
+
+if (!outputPath) {
+  throw new Error(
+    `Unknown SDK_ENV: ${sdkEnv}. Expected one of 'node', 'browser'.`
+  );
+}
```
samijaber, line 34:
```suggestion
```

packages/sdks/output/angular/scripts/multi-build.mjs
sidmohanty11: done! in angular we cannot directly use an extension so I've made a minor update to `getEvaluatorPathAlias` so that if a `format` is mentioned then only use it else don't add the extension. I have also updated `getEvaluatorPathAlias` in places where it was using the default format (ts) -- only one

packages/sdks/output/angular/scripts/multi-build.mjs
sidmohanty11: thanks done!

packages/sdks/output/angular/scripts/multi-build.mjs
sidmohanty11: done

packages/sdks/output-generation/index.js
```diff
@@ -41,11 +41,11 @@ const getFolderName = () => {
  */
 const getEvaluatorPathAlias = (options = {}) => {
   const pointTo = options.pointTo || 'input';
-  const format = options.format || 'ts';
+  const format = options.format;
```
samijaber, line 44: mmm this change will change the default behaviour of this function for all other SDKs. I think many of them are expecting `format` to be set to `ts` by default.

Why do you need this change? You should always be able to use an alias with an explicit path, i.e. `${folder}/index.js` or `${folder}/index.ts`

samijaber: CI seems to pass! I was worried that certain SDK TS configs would complain about extensionless imports...but if they're not complaining I guess there's no issue here

sidmohanty11: i checked it globally it was only used with the default `ts` format in one place which I've updated.

so the issue here is if we set it to any specific format, when angular performs transformation it goes from `ts` to `mjs` which cannot resolve any hardcoded `.ts` extension to create an FESM build. If you remember this is the reason why we added `explicitImportFileExtension` in mitosis

samijaber: Ah, thanks for explaining.

> i checked it globally it was only used with the default ts format in one place which I've updated.

Actually, we use `viteOutputGenerator()` with default arguments in many places, which ends up calling this function without a `format` argument. So there are more places where it's being used.

But now that I think about it, it seems like we never needed this `format` to begin with, and we could get away with not providing it at all.

Either way, this PR is in a good state and can be merged 👍🏽

sidmohanty11: ohh got it

PR #3400, opened 2024-07-10
fix: SDKs: columns width calculation

manucorporat:
Screenshot 2024-07-10 at 15 02 50

## TO-DO:

- [x] add tests


sanyamkamat (approved):

mrkoreye (approved):
math is hard

PR #3357, opened 2024-06-25
fix: extra props passing to components

sidmohanty11:
## Description

Don't pass extra builder props if not necessary

**Jira**
https://builder-io.atlassian.net/browse/ENG-5218
https://builder-io.atlassian.net/browse/ENG-5974

_Screenshot_
If relevant, add a screenshot or two of the changes you made.

PR #3406, opened 2024-07-11
Feature: shouldeReceiveBuilderProps sdks

samijaber:
## Description

- add `shouldeReceiveBuilderProps` config for components used to determine what props to provide.


packages/sdks/src/components/block/block.helpers.ts
```diff
@@ -112,80 +112,76 @@ export const getInheritedStyles = ({
   return extractTextStyles(style);
 };
 
+// once we bump to a major version, toggle this to `false`.
+const DEFAULT_SHOULD_RECEIVE_PROPS = true;
```
sidmohanty11, line 116: this means we will have access to all the builder props by default while developing a custom component right?

do you think for any kind of console errors (for angular and vue) we should communicate in the docs for them to define those props or follow `shouldReceiveBuilderProps` pattern?

samijaber: This change reverts the SDK to how it _used_ to operate. We stopped providing `builderContext` and `builderBlock` in https://github.com/BuilderIO/builder/pull/3357, which is now breaking some customer components that expect these props.

In a followup PR we will do a major version release of the SDK and se the default to `false`

sidmohanty11: got it!

sidmohanty11 (approved):

PR #3414, opened 2024-07-12
remove iframely logic from SDKs

steve8708:
that way we can more easily update this and rotate keys

this is now overriden in our web app and not needed int he SDK source code

### DIFF
--- a/packages/react/src/blocks/Embed.tsx
+++ b/packages/react/src/blocks/Embed.tsx
@@ -3,9 +3,6 @@ import React from 'react';
 import { Builder } from '@builder.io/sdk';
 import { withBuilder } from '../functions/with-builder';
 
-// Provided by the web app as a local variable in onChange functions
-declare const _iframelyApiKey: string;
-
 class EmbedComponent extends React.Component<any> {
   elementRef: HTMLElement | null = null;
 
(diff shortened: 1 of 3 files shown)


packages/react/src/blocks/Embed.tsx
```diff
@@ -109,6 +110,8 @@ export const Embed = withBuilder(EmbedComponent, {
           options.delete('content');
         }
       },
+=======
+>>>>>>> Stashed changes
```
samijaber, line 114: euh oh!

steve8708: i swear i know how to code

packages/sdks/src/blocks/embed/component-info.ts
```diff
@@ -20,35 +20,6 @@ export const componentInfo: ComponentInfo = {
       required: true,
       defaultValue: '',
       helperText: 'e.g. enter a youtube url, google map, etc',
-      onChange: (options: Map<string, any>): void | Promise<void> => {
-        const url = options.get('url');
-        if (url) {
-          options.set('content', 'Loading...');
-
-          const apiKey = _iframelyApiKey;
```
samijaber, line 28: I think you can also delete the `declare const _iframelyApiKey: string;` in these files as a result, no?

steve8708: tough crowd

samijaber (approved):

PR #3426, opened 2024-07-15
Bynder Plugin example code

mix3d:
## Description

Creates a plugin for Bynder, a DAM similar to Cloudinary.

_Screenshots_
Plugin Settings:
image
Unselected State:
image
Bynder Universal Compact View modal:
image
Selected State:
image

## Testing Steps

1. Install the dependencies and run `npm run start`
2. Add the plugin to a space with the following url: `http://localhost:1268/plugin.system.js?pluginId=@builder.io/bynder-plugin`
3. BYO Bynder account. (Builder team: Request a Bynder account from @mix3d)
4. Configure the plugin with Builder's test account info
5. Add a custom input with the BynderSingleSelect, test selecting a file 

## Notes / Discussion items:
* Does not yet support a per-instance file-type selector like the built-in Builder one does, and probably would require changes in our plugin infrastructure to support. 
* Bynder's recommendation to use `"SingleSelectFile"` instead of `"SingleSelect"` for the picker mode may require small changes, but these can also be communicated to any user via README or directly to the interested party.


plugins/bynder-example/README.md
```diff
@@ -0,0 +1,104 @@
+# Bynder Plugin for Builder.io
```
teleaziz, line 1:
```suggestion
# Example Bynder Plugin for Builder.io
```

teleaziz (commented):
looks good , left one comment

teleaziz (approved):

PR #3437, opened 2024-07-17
fix: resolver approach for ssr in angular and `onInit`

sidmohanty11:
## Description

Uses `onInit` in both CSR and SSR and `onMount` & `onUpdate` got shifted to only run on the browser. This uses the resolver approach for fetching content during SSR. This also fixes `blocks-wrapper` not respecting children inside and moving them out.

_Screenshot_


packages/sdks/src/components/content-variants/content-variants.lite.tsx
```diff
@@ -82,15 +82,13 @@ export default function ContentVariants(props: VariantsProviderProps) {
...
-          state.shouldRenderVariants = checkShouldRenderVariants({
-            canTrack: props.canTrack,
-            content: props.content,
-          });
-        }
+        state.shouldRenderVariants = checkShouldRenderVariants({
+          canTrack: getDefaultCanTrack(props.canTrack),
+          content: props.content,
+        });
       },
```
samijaber, line 92: 2 things:

- this is an SDK change, so we'd need a `changesets` entry in this PR
- why do we need this `onInit` at all? The initial value is already set to:

```ts
shouldRenderVariants: checkShouldRenderVariants({
      canTrack: getDefaultCanTrack(props.canTrack),
      content: props.content,
    }),
```

Given the recent improvements in Mitosis, wouldn't this initial value be initialized properly in Angular?

sidmohanty11: added the changeset

right, we should handle the state initialization correctly inside mitosis. should i make a follow-up pr where I will clean up this after we've added it to mitosis?

samijaber (approved):

PR #3446, opened 2024-07-26
fix: models restriction for custom components

sidmohanty11:
## Description

Restrict custom components to the specific `models` (if `models` attribute is explicitly mentioned)

**Jira**
https://builder-io.atlassian.net/browse/ENG-6193

_Screenshot_


packages/sdks/src/components/content/content.lite.tsx
```diff
@@ -53,7 +53,9 @@ export default function ContentComponent(props: ContentProps) {
 
     registeredComponents: [
       ...getDefaultRegisteredComponents(),
-      ...(props.customComponents || []),
+      ...(props.customComponents?.filter(
+        ({ models }) => !models?.length || models.includes(props.model || '')
```
samijaber, line 57: why the `|| ''` in here? If `props.model` does not exist then we always want this to be `true`?

If so, I would prefer making this logic more explicit, as its getting a bit messy to parse:

```suggestion
        ({ models }) => {
          if (!models?.length) return true;
          if (!props.model) return true;
          return models.includes(props.model); 
        }
```

samijaber (approved):

packages/sdks/src/components/content/content.lite.tsx
sidmohanty11: @samijaber `props.model` should be a required field right? I had added it such that it will return `false` if `props.model` isn't present && we have an explicit `models` array. Should we allow to register components if there is no `props.model`?

samijaber: yeah, I see that we have a types mistake: `props.model` should be required but it is currently optional. 

We should fix that in the future and do a major version bump to avoid customers hitting surprise TS issues.

samijaber: for now I think we should go with my suggestion, but not change the way registered components work.

PR #3447, opened 2024-07-26
remove srcset for svg images

JLarky:
## Description

It sounds like we are wasting a lot of time trying to add src set for svg images even though they are not going to use them :) I wonder if this simple change can help out in that case

before|after
-|-
image|image

### DIFF
--- a/.changeset/polite-pets-smell.md
+++ b/.changeset/polite-pets-smell.md
@@ -1,0 +1,6 @@
+---
+"@builder.io/packages": patch
+"@builder.io/sdks": patch
+---
+
+remove srcset for svg images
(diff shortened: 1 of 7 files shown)


.changeset/polite-pets-smell.md
```diff
@@ -0,0 +1,13 @@
...
+'@builder.io/sdk-react-nextjs': patch
+'@builder.io/sdk-qwik': patch
+'@builder.io/sdk-react': patch
+'@builder.io/sdk-react-native': patch
+'@builder.io/sdk-solid': patch
+'@builder.io/sdk-svelte': patch
+'@builder.io/sdk-vue': patch
+---
+
+Fix: remove srcset for svg images
```
samijaber, line 13:
```suggestion
Fix: Image block: remove redundant `srcset` for SVG images
```

samijaber (approved):
Thanks a lot for the contribution @JLarky!

PR #3449, opened 2024-07-29
fix: repeat items angular sdk and some other fixes

sidmohanty11:
## Description

Fixes repeat items in angular SDK and columns state initialization

the generated content in `RepeatBlock` had this computed state which came out `undefined` always. We are now binding the `repeatContext` directly to make it reactive

```
store = this.repeatContext;
```

_Screenshot_

### DIFF
--- a/packages/sdks/mitosis.config.cjs
+++ b/packages/sdks/mitosis.config.cjs
@@ -576,13 +576,17 @@ const ANGULAR_WRAP_SYMBOLS_FETCH_AROUND_CHANGES_DEPS = () => ({
 });
 
 const ANGULAR_FIX_REPEAT_ITEMS_PLUGIN = () => ({
-  code: {
-    post: (code) => {
-      if (code.includes('repeated-block, RepeatedBlock')) {
-        code = code.replace('store = this.repeatContext;', '');
-        code = code.replace('[context]="store"', '[context]="repeatContext"');
+  json: {
+    pre: (json) => {
+      if (json.name === 'RepeatedBlock') {
+        const ctxCode = json.state['store'].code;
+        if (json.children[0].name === 'Block') {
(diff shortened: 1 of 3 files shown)


packages/sdks/src/blocks/columns/columns.lite.tsx
```diff
@@ -206,6 +207,23 @@ export default function Columns(props: ColumnProps) {
...
+            : props.reverseColumnsWhenStacked
+              ? 'column-reverse'
+              : 'column';
+        state.gutterSize =
+          typeof props.space === 'number' ? props.space || 0 : 20;
+        state.cols = props.columns || [];
+        state.stackAt = props.stackColumnsAt || 'tablet';
+      },
+    });
+  });
```
samijaber, line 225: I see you've added many `onInit` hooks that are duplicates of the initial values for certain `state.*` values.

I think we should improve the mitosis generator so that it does this automatically: any initial `state` value should be initialized properly without need for duplication

packages/sdks/mitosis.config.cjs
```diff
@@ -575,6 +575,22 @@ const ANGULAR_WRAP_SYMBOLS_FETCH_AROUND_CHANGES_DEPS = () => ({
...
+        if (json.children[0].name === 'Block') {
+          json.children[0].bindings['context'].code = ctxCode;
+          delete json.state['store'];
+        } else {
+          throw new Error('RepeatedBlock should have a Block child');
+        }
+      }
+    },
+  },
+});
```
samijaber, line 592: can you add a comment for what this is doing?

samijaber (commented):
- `show-hide-if` and `repeat-items` tests are passing but expected to fail
- no changesets

PR #3462, opened 2024-08-01
cleanup `angular-ssr` snippets and comments

sidmohanty11:
## Description

cleanup `angular-ssr` snippets to use resolver approach and eslint disable comments to make snippets easily copy-pastable.

_Screenshot_

### DIFF
--- a/packages/sdks/e2e/angular/src/app/app.component.ts
+++ b/packages/sdks/e2e/angular/src/app/app.component.ts
@@ -1,6 +1,4 @@
-// fails because type imports cannot be injected
-// eslint-disable-next-line @typescript-eslint/consistent-type-imports
-import { ChangeDetectorRef, Component } from '@angular/core';
+import { Component } from '@angular/core';
 import {
   _processContentResult,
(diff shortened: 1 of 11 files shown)


samijaber (approved):

packages/sdks/e2e/angular/src/app/app.component.ts
```diff
@@ -78,7 +74,5 @@ export class AppComponent {
     this.apiKey = builderProps.apiKey;
     this.model = builderProps.model;
     this.apiVersion = builderProps.apiVersion;
-
-    this.cdr.detectChanges();
```
samijaber, line 82: CleanShot 2024-08-01 at 12 33 42@2x

i dont know if this passed by luck!

if it's flakey, might have to use `test.skip()` instead of `test.fail()`

sidmohanty11: yes we should add `test.skip()` to it I think, because A/B tests are not working yet

PR #3464, opened 2024-08-02
Accept 'templateUId' setting from users for phrase plugin

murtaza-builder:
## Description

Accept 'templateUId' setting from users for phrase plugin to help in creating "machine translation" supported Phrase translation projects.

### DIFF
--- a/plugins/phrase-conector/src/plugin.tsx
+++ b/plugins/phrase-conector/src/plugin.tsx
@@ -32,12 +32,14 @@ registerPlugin(
       },
       {
         name: 'templateUId',
-        friendlyName: 'Template Id',
-        helperText: 'ID of template project to be use when creating projects for translation jobs',
+        friendlyName: 'Template ID',
+        helperText:
+          'Template ID is the unique identifier of a Phrase Template used when creating a new Phrase Project',
         type: 'string',
       },
       {
         name: 'isUSDataCenterAccount',
+        friendlyName: "Account's data center is US based",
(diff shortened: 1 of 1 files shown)


plugins/phrase-conector/src/plugin.tsx
```diff
@@ -30,6 +30,12 @@ registerPlugin(
         type: 'password',
         required: true,
       },
+      {
+        name: 'templateUId',
```
shyam-builder, line 34: why is it called templateUId instead of templateId?

rima-builder (approved):

plugins/phrase-conector/src/plugin.tsx
sanyamkamat, line 34:
```suggestion
        name: 'templateId',
```

sanyamkamat:
Same Q as Shyam. Also, I do not see any ticket/slack thread in the description for the change.

plugins/phrase-conector/src/plugin.tsx
murtaza-builder: using same name as used in Phrase api

murtaza-builder: image

sanyamkamat, line 34: Matching the Phrase API
```suggestion
        name: 'templateUid',
```

shyam-builder (approved):

PR #3481, opened 2024-08-07
Update the react native sample to use the latest builder sdk

kylefowler:
## Description

Updates the react native sample app on a few major dependencies
- Updates the builder react-native sample app to the latest version
- Updates expo to the latest
- migrates from react navigation to expo-router
- commits the shell android and ios projects generated by `expo run:ios` and `expo run:android`

TODO:
- After this update the web version of the RN sample will not scroll. Will follow up on that in a future ticket as the mobile sides are working correctly.


examples/react-native/babel.config.js
```diff
@@ -3,4 +3,4 @@ module.exports = function (api) {
   return {
     presets: ['babel-preset-expo'],
   };
-};
+};
```
teleaziz, line 6: there are few linting issues, you probably can fix them all at once using the linter from the root of the repo:
```
yarn lint:fix
```

teleaziz (commented):
are those `android` and `ios` foldeers auto generated? if so, maybe we can omit them and simplify this example a little

examples/react-native/babel.config.js
kylefowler: should the build fail if there are linting issues?

kylefowler: ah the examples directory isnt part of the root project

kylefowler: got the linter to run off the same config as in the root, it says there are no issues in here

teleaziz (approved):

PR #3485, opened 2024-08-08
add support for responsive styles in gen1 react sdk and memoize processBlock

steve8708:
we don't use the post-processed responsiveStyles object that bindings influence, but instead the preprocessed one

gen2 supports this, but not gen1

also, i removed several repeat calls to processBlock() which is very bad for perf as that runs al l theb indings, and it was running many times _per block_ instead of 1x like gen1 SDK


packages/sdks/src/functions/get-processed-block.ts
```diff
@@ -20,11 +45,14 @@ const evaluateBindings = ({
   if (!block.bindings) {
     return block;
   }
-  const copy = fastClone(block);
+  // TODO: don't fast clone this! Any
+  const copy = deepCloneWithConditions(omit(block, 'children', 'meta'));
```
samijaber, line 49: @steve8708 can you expand on the reasoning behind this? does it improve perf? or solve a bug?

steve8708: yeah each block can have infinite child blocks, we don't want to re-clone each child when we visit each child, so this is to improve perf. 

tl;dr we want to clone basically everytihng except blocks. so if we ever hit a block we bail and return the block itself without traversing deeper

we also skip meta bc some people put a lot of stuff in meta and we don't technically claim to support bindings in meta (tehre is no point of that, meta isn't used in rendering)

so the idea is really lets just clone properties, styles, and component options (that aren't blocks)

samijaber:
Closing this due to being superseded by #3519 and #3522

PR #3491, opened 2024-08-13
fix[angular-gen2]: object destructuring in angular

sidmohanty11:
## Description

Uses https://github.com/BuilderIO/mitosis/pull/1522 for destructuring out attributes correctly in Angular SDK

**Jira**
https://builder-io.atlassian.net/browse/ENG-6294

_Screenshot_


packages/sdks/overrides/angular/src/components/dynamic-renderer/dynamic-renderer.ts
```diff
@@ -88,160 +81,2537 @@ export default class DynamicRenderer {
...
+        this.v?.nativeElement,
+        this.attributes,
+        changes.attributes.currentValue
+      );
+    }
+  }
+
+  ngOnDestroy() {
+    this._listenerFns.forEach((fn) => fn());
+  }
```
samijaber, line 2616: haha this is wild 😂. 

would it be possible to write a script that generates these? The same way we have `build-inline-fns` running after Mitosis but before `ng build`. we could have a `generate-dynamic-renderer`.

It could use a template like:

```ts
@Component({
  selector: '${SELECTOR}',
  template: ` <${ELEMENT_NAME} #v><ng-content></ng-content></${ELEMENT_NAME}> `,
  standalone: true,
  styles: [':host { display: contents; }'],
})
export class DynamicIframe {
  @Input() attributes!: any;
  @Input() actionAttributes?: any;

  @ViewChild('v', { read: ElementRef }) v!: ElementRef;

  _listenerFns = new Map<string, () => void>();

  constructor(private renderer: Renderer2) {}

  setAttributes(el: any, value: any, changes?: any) {
    if (!el) return;

    const target = changes ? changes : value;
    Object.keys(target).forEach((key) => {
      if (key.startsWith('on')) {
        if (this._listenerFns.has(key)) {
          this._listenerFns.get(key)!();
        }
        this._listenerFns.set(
          key,
          this.renderer.listen(
            el,
            key.replace('on', '').toLowerCase(),
            target[key]
          )
        );
      } else {
        this.renderer.setAttribute(el, key, target[key] ?? '');
      }
    });
  }

  ngAfterViewInit() {
    this.setAttributes(this.v?.nativeElement, this.attributes);
  }

  ngOnChanges(changes: SimpleChanges) {
    if (changes.attributes) {
      this.setAttributes(
        this.v?.nativeElement,
        this.attributes,
        changes.attributes.currentValue
      );
    }
  }

  ngOnDestroy() {
    this._listenerFns.forEach((fn) => fn());
  }
}
```

and append the code to `dynamic-renderer.ts` looping over an array of HTML tags.

that will make it a lot easier to maintain this override in the future.

packages/sdks/output-generation/angular-dynamic-renderer-generator.js
```diff
@@ -0,0 +1,196 @@
+import fs from 'fs';
```
samijaber, line 1: Oh, you did create a generator, but you're using it to manually create the override and commit the output.

How about, instead of that, you add it to the build pipeline of the angular SDK, and make it write to `output/angular/src/components/dynamic-renderer/dynamic-renderer.ts` instead of `overrides/angular/src/components/dynamic-renderer/dynamic-renderer.ts`.

This way the override can stay simple!

packages/sdks/mitosis.config.cjs
```diff
@@ -538,20 +518,37 @@ const ANGULAR_BIND_THIS_FOR_WINDOW_EVENTS = () => ({
   code: {
     post: (code) => {
       if (code.includes('enable-editor')) {
-        code = code.replace(
-          'window.addEventListener("message", this.processMessage);',
-          'window.addEventListener("message", this.processMessage.bind(this));'
+        // find two event listeners and add bind(this) to the fn passed
+        const eventListeners = code.match(
+          /window\.addEventListener\(\s*['"]([^'"]+)['"]\s*,\s*([^)]+)\)/g
```
samijaber, line 523: yikes 😅

packages/sdks/output-generation/angular-dynamic-renderer-generator.js
sidmohanty11: ahh best! great idea. Let me update it

packages/sdks/mitosis.config.cjs
sidmohanty11: with formatting and whitespaces sometimes the code wasn't matching 😢

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
```diff
@@ -0,0 +1,202 @@
...
+    this.myContent = [this.vcRef.createEmbeddedView(this.tagnameTemplateRef).rootNodes];
+  }
+}
+`;
+
+  htmlElements.forEach((tagName) => {
+    dynamicRendererCode += dynamicComponentTemplate(tagName) + '\n';
+  });
+
+  fs.writeFileSync(PATH_TO_DYNAMIC_RENDERER, dynamicRendererCode);
```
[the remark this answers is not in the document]
sidmohanty11: right thanks for the catch, have added the check

samijaber (approved):

PR #3492, opened 2024-08-13
serialize functions for registering plugins too

steve8708:
so you can have showIf on fields as functions

more context: https://builder-internal.slack.com/archives/CMBPNS6RM/p1723590863859099?thread_ts=1722964078.709639&cid=CMBPNS6RM

### DIFF
--- a/packages/sdks/src/functions/register.ts
+++ b/packages/sdks/src/functions/register.ts
@@ -1,6 +1,7 @@
 import type { BuilderBlock } from '../types/builder-block.js';
 import type { DeepPartial } from '../types/deep-partial.js';
 import { isBrowser } from './is-browser.js';
+import { serializeIncludingFunctions } from './register-component.js';
 
 export interface InsertMenuItem {
   name: string;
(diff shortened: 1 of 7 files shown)


packages/sdks/src/functions/register.ts
```diff
@@ -21,6 +22,9 @@ const registry: { [key: string]: any[] } = {};
 export function register(type: 'insertMenu', info: InsertMenuConfig): void;
 export function register(type: string, info: any): void;
 export function register(type: string, info: any) {
+  if (type === 'plugin') {
+    info = serializeIncludingFunctions(info);
+  }
```
samijaber, line 27: is there a reason not to always call `serializeIncludingFunctions` for every `register` message?

steve8708: we could do that, but i don't want to break anything existing so felt this was safer. also may be an unnecessary perf overhead if what we have currently works fine

my thoughts were if this becomes whack a mole (e.g. some other register message also needs this) we'd just have it run for everything

samijaber (approved):
LGTM, just needs `yarn g:changeset` to add the changesets to all the SDKs

PR #3493, opened 2024-08-15
add support for bindings in text blocks

steve8708:
ticket: https://builder-io.atlassian.net/browse/PLAN-167

### DIFF
--- a/packages/sdks-tests/src/e2e-tests/text-eval.spec.ts
+++ b/packages/sdks-tests/src/e2e-tests/text-eval.spec.ts
@@ -1,0 +1,9 @@
+import { expect } from '@playwright/test';
+import { test } from '../helpers/index.js';
+
+test.describe('Text Eval', () => {
+  test('binds text to state', async ({ page }) => {
+    await page.goto('/text-eval');
+    await expect(page.getByText('Device Size: large')).toBeVisible();
+  });
+});
--- a/.changeset/heavy-goats-sleep.md
+++ b/.changeset/heavy-goats-sleep.md
@@ -1,0 +1,12 @@
+---
+"@builder.io/sdk-angular": patch
+"@builder.io/sdk-react-nextjs": patch
+"@builder.io/sdk-qwik": patch
+"@builder.io/sdk-react": patch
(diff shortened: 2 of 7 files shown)


sidmohanty11 (approved):
LGTM, last thing: we need to add changesets for our gen2 sdks so that we can publish them

for that, we can run `yarn g:changeset` and any packages with `@builder.io/sdk-[framework]` to it is gen2 so we can select all of them

packages/sdks-tests/src/e2e-tests/text-eval.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { expect } from '@playwright/test';
+import { test } from '../helpers/index.js';
+
+test.describe('Text Eval', () => {
+  test('binds text to state', async ({ page }) => {
+    await page.goto('/text-eval');
+    await expect(page.getByText('Device Size: large')).toBeVisible();
```
samijaber, line 7: Something is off about this test.

React Native has a manual override for the Text block component which still doesn't handle bindings:

CleanShot 2024-08-19 at 11 55 56@2x

For some reason playwright still passes the test for it...not sure why.


My main point is that this feature is still unsupported in that SDK. It needs to be added to this file: https://github.com/BuilderIO/builder/blob/main/packages/sdks/overrides/react-native/src/blocks/text/text.jsx#L90-L105

The other issue is understanding why this test is passing despite the `large` text not being in the DOM...

samijaber (changes requested):
Requesting changes because the e2e test has a false positive: it doesn't actually work in RN SDK. We need to mirror the changes in `overrides/react-native/src/blocks/text/text.jsx`, and preferably fix the test so it actually fails when the text is not there.

packages/sdks-tests/src/e2e-tests/text-eval.spec.ts
kylefowler: made these changes. couldnt recreate locally the phantom test passing. modified the test to explicitly fail and it behaved as expected, and after my changes the test also passes correctly as well.

samijaber (approved):

PR #3501, opened 2024-08-20
fix: inner layout "columns" crashing

sidmohanty11:
## Description
`isRSC = true` makes Columns block a non-reactive element (`isInteractive = false`), this then is not able to get `interactiveElementProps` which contains `Wrapper` information. Updated it such that the `isInteractive` check is scoped to RSC SDK only.

Links:

[where `isInteractive` is false](https://github.com/sidmohanty11/builder/blob/fix/inner-layouts-columns-crash/packages/sdks/src/components/block/components/component-ref/component-ref.lite.tsx/#L48)

[`isInteractive = false` so we don't have access to Wrapper](https://github.com/sidmohanty11/builder/blob/fix/inner-layouts-columns-crash/packages/sdks/src/components/block/components/component-ref/component-ref.helpers.ts/#L57-L65)

**Jira**
https://builder-io.atlassian.net/browse/PLAN-76

**Loom**
https://www.loom.com/share/e7cbae3ca62b493e98b5179712754e62

### DIFF
--- a/packages/sdks-tests/src/e2e-tests/editing.spec.ts
+++ b/packages/sdks-tests/src/e2e-tests/editing.spec.ts
@@ -99,10 +99,13 @@ test.describe('Visual Editing', () => {
     }
 
     await sendContentUpdateMessage({ page, newContent: MODIFIED_EDITING_COLUMNS, model: 'page' });
+    // had to hack this so that we can wait for the content update to actually show up (was failing in Qwik)
+    await page.frameLocator('iframe').getByText('third').waitFor();
 
-    await expect(firstText).toBeVisible();
+    const updatedFirstText = page.frameLocator('iframe').getByText('third');
+    await expect(updatedFirstText).toBeVisible();
     await expect(secondText).toBeVisible();
-    const updatedFirstBox = await firstText.boundingBox();
+    const updatedFirstBox = await updatedFirstText.boundingBox();
     const updatedSecondBox = await secondText.boundingBox();
     if (updatedFirstBox && updatedSecondBox) {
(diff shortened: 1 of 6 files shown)


samijaber (approved):
Great find!! 

Would you be able to add an editor test to this? it would be like this one but for Columns:

https://github.com/BuilderIO/builder/blob/ae9ecf13a6f1825dd071060a4708a3d3fdacb044/packages/sdks-tests/src/e2e-tests/editing.spec.ts#L28

packages/sdks-tests/src/e2e-tests/editing.spec.ts
```diff
@@ -67,6 +68,51 @@ test.describe('Visual Editing', () => {
...
     await page.frameLocator('iframe').getByText(NEW_TEXT).waitFor();
   });
+
+  test('correctly updates Box -> Columns when used Inner Layout > Columns option', async ({
+    page,
+    packageName,
+    basePort,
+  }) => {
+    test.skip(
+      packageName === 'react-native' ||
```
samijaber, line 78: Remind me why we're skipping RN SDK? I think this is an old issue before we started building the e2e RN app properly. I wonder if we no longer need to skip it anymore.

If you try and it works, i would remove this. If not, would be good to put it in a separate `test.skip()` and explain briefly why it's failing (if you're able to see why locally)

samijaber (approved):
👏🏽

packages/sdks-tests/src/e2e-tests/editing.spec.ts
sidmohanty11: ahh right, for this specific test, we will have the same layout (stacked) for columns in RN. Have updated the tests to include React native and they all are passing

PR #3510, opened 2024-08-23
fix: drag and drop in Columns and "+Add Block" for RN SDK

sidmohanty11:
## Description

Fixes drag and drop functionality to correctly drop-in components in the desired path and click event for "+Add Block".
**Jira**
https://builder-io.atlassian.net/browse/PLAN-75

_Screenshot_

### DIFF
--- a/packages/sdks/mitosis.config.cjs
+++ b/packages/sdks/mitosis.config.cjs
@@ -786,6 +786,19 @@ module.exports = {
             },
           },
         }),
+        () => ({
+          code: {
+            post: (code) => {
+              if (
+                code.includes('BlocksWrapper') ||
+                code.includes('EnableEditor')
+              ) {
+                code = code.replace('onPress', 'onClick');
+              }
+              return code;
+            },
+          },
+        }),
       ],
     },
     qwik: {
--- a/packages/sdks/mitosis.config.cjs
+++ b/packages/sdks/mitosis.config.cjs
@@ -793,6 +793,10 @@ module.exports = {
                 code.includes('BlocksWrapper') ||
                 code.includes('EnableEditor')
               ) {
+                /**
+                 * Replaces `onPress` event handler with `onClick` for React Native
+                 * such that visual editing "+Add Block" works on web target.
+                 */
                 code = code.replace('onPress', 'onClick');
               }
               return code;
--- a/packages/sdks/overrides/react-native/src/functions/transform-block-properties.ts
+++ b/packages/sdks/overrides/react-native/src/functions/transform-block-properties.ts
@@ -5,6 +5,7 @@ import { isEditing } from './is-editing.js';
 
 export function transformBlockProperties({
   properties,
+  block,
 }: {
   block: BuilderBlock;
   context: BuilderContextInterface;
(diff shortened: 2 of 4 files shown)


packages/sdks/overrides/react-native/src/functions/transform-block-properties.ts
```diff
@@ -26,6 +27,14 @@ export function transformBlockProperties({
       if (el && !(el instanceof Text)) {
         el.setAttribute('builder-id', id);
         el.classList.add(id);
+        // Support inline text editing for React Native SDK
+        if (block.component?.name === 'Text') {
+          const textElement = el.querySelector('[dir="auto"]');
```
samijaber, line 32: the `dir="auto"` selector doesn't completely fix inline editing. it will only grab the first `div` to match it, which is not enough. See https://www.loom.com/share/ad2bfcbd24594d3283ee5a3fb3a0bc67

What you need to do is set `builder-text` on the root `View`/`div` created by `react-native-render-html`. You can do that by grabbing the firstElementChild of `el`

```suggestion
          const textElement = el.firstElementChild
```

samijaber: For now, I would remove this inline editing code fix completely until we figure out how to properly get it working.

packages/sdks/mitosis.config.cjs
```diff
@@ -786,6 +786,19 @@ module.exports = {
...
+        () => ({
+          code: {
+            post: (code) => {
+              if (
+                code.includes('BlocksWrapper') ||
+                code.includes('EnableEditor')
+              ) {
+                code = code.replace('onPress', 'onClick');
+              }
+              return code;
```
samijaber, line 798: a comment on why we need to do this would be good

packages/sdks/src/components/blocks/blocks-wrapper.lite.tsx
```diff
@@ -61,8 +68,23 @@ export default function BlocksWrapper(props: BlocksWrapperProps) {
...
 
+  onMount(() => {
+    useTarget({
+      reactNative: () => {
+        if (isEditing()) {
+          props.path &&
+            blocksWrapperRef.setAttribute('builder-path', props.path);
+          props.parent &&
+            blocksWrapperRef.setAttribute('builder-parent-id', props.parent);
+        }
```
samijaber, line 79: a comment on this would be good (react native strips off custom HTML attributes, so we have to manually set them here)

samijaber (approved):

packages/sdks/mitosis.config.cjs
sidmohanty11: yup! added

packages/sdks/src/components/blocks/blocks-wrapper.lite.tsx
sidmohanty11: thanks! added

samijaber (approved):

packages/sdks/overrides/react-native/src/functions/transform-block-properties.ts
sidmohanty11: thanks @samijaber , have reverted this change for now. Will test this out thoroughly and create a separate PR because if you check my loom (it propagates properly):
https://www.loom.com/share/425267146fb04c24ac7b6c1d636e9739

samijaber: You're right....interesting.

PR #3513, opened 2024-08-23
rn validation fixes

steve8708:
from testing on joann's content

### DIFF
--- a/packages/sdks/overrides/react-native/src/functions/transform-style-property.ts
+++ b/packages/sdks/overrides/react-native/src/functions/transform-style-property.ts
@@ -19,11 +19,11 @@ function extractVarValue(value: string): string {
   return newValue;
 }
 // Common regex patterns
-const numberPattern = /^-?\d+(\.\d+)?$/;
-const lengthPattern = /^-?(\d+)(px|%)?$/;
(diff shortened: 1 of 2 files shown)


packages/sdks/overrides/react-native/src/functions/transform-style-property.ts
```diff
@@ -19,11 +19,11 @@ function extractVarValue(value: string): string {
   return newValue;
 }
 // Common regex patterns
-const numberPattern = /^-?\d+(\.\d+)?$/;
-const lengthPattern = /^-?(\d+)(px|%)?$/;
-const pixelPattern = /^-?(\d+)(px)?$/;
+const numberPattern = /^-?\d*\.?\d+$/;
+const lengthPattern = /^-?\d*\.?\d+(px|%)?$/;
+const pixelPattern = /^-?\d*\.?\d+(px)?$/;
```
samijaber, line 24: is the idea to allow floats? what was wrong with the previous regex?

steve8708: yeah floats

samijaber (approved):

PR #3516, opened 2024-08-26
fix: [Plugins][Smartling] Add appropriate error message when we have switched envs and there are no projects found

murtaza-builder:
## Description

Updating the error message when there are no smartling projects found coz user might have recently switched to a diff env/space.

_Screenshot_
image

### DIFF
--- a/plugins/smartling/src/plugin.tsx
+++ b/plugins/smartling/src/plugin.tsx
@@ -439,13 +439,27 @@ registerPlugin(
           return api.getProject(id).then(res => transformProject(res.project));
         },
         search(q = '') {
-          return api
-            .getAllProjects()
-            .then(res =>
-              res.results
-                .filter(proj => proj.projectName.toLowerCase().includes(q.toLowerCase()))
-                .map(transformProject)
-            );
+          return api.getAllProjects().then(res => {
+            if (!res.results || res.results.length === 0) {
+              return appState.globalState
+                .getPluginPrivateKey(pkg.name)
+                .then((pluginPrivateKey: string) => {
+                  if (api.isPluginPrivateKeySame(pluginPrivateKey)) {
+                    appState.snackBar.show('Oh no! There was an error searching for resources');
+                  } else {
+                    appState.snackBar.show(
+                      'Please refresh your browser to view your Smartling projects.'
+                    );
+                  }
+
+                  return [];
+                });
+            }
+
+            return res.results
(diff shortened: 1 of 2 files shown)


plugins/smartling/src/smartling.ts
```diff
@@ -11,125 +11,132 @@ export type Project = {
...
-  loaded?: Promise<void>;
-  resolveLoaded?: () => void;
-  // TODO: basic cache
-  getBaseUrl(path: string, search = {}) {
-    const params = new URLSearchParams({
-      ...search,
-      pluginId: pkg.name,
-      apiKey: appState.user.apiKey,
-    });
+         private privateKey?: string;
```
shyam-builder, line 14: seems extra indentation snuck in for the whole file?

plugins/smartling/src/plugin.tsx
```diff
@@ -439,13 +439,27 @@ registerPlugin(
...
-                .filter(proj => proj.projectName.toLowerCase().includes(q.toLowerCase()))
-                .map(transformProject)
-            );
+          return api.getAllProjects().then(res => {
+            if (!res.results || res.results.length === 0) {
+              return appState.globalState
+                .getPluginPrivateKey(pkg.name)
+                .then((pluginPrivateKey: string) => {
+                  if (api.isPluginPrivateKeySame(pluginPrivateKey)) {
+                    appState.snackBar.show('Oh no! There was an error searching for resources');
```
shyam-builder, line 448: can you validate with Sajal on the message? Is "oh no" on brand for us? I haven't seen that anywhere else

plugins/smartling/src/smartling.ts
murtaza-builder: yes, but ther is one small addition of `isPluginPrivateKeySame` method. 
I tried pre-pr and linting explicitly, but the extra indendation always applied back.

plugins/smartling/src/plugin.tsx
murtaza-builder: I ahve just reused the message thats already there in ResourcePicker as was the case previously. Only updated the message in this particular scenario
image

murtaza-builder: There are a dozen of places in webapp where "oh no! ..." message are used..
image

shyam-builder (approved):

PR #3525, opened 2024-09-02
fix[qwik-gen2]: export `initializeNodeRuntime` for qwik SDK

sidmohanty11:
## Description

Exports `initializeNodeRuntime` in Qwik SDK so that customers can import this in a server-side only call to correctly import and initialize `isolated-vm` and handle bindings correctly.

**Jira**
https://builder-io.atlassian.net/browse/ENG-6287

_Screenshot_
Build structure now:
Screenshot 2024-09-04 at 12 03 24 PM

### DIFF
--- a/packages/sdks/output/qwik/vite.config.ts
+++ b/packages/sdks/output/qwik/vite.config.ts
@@ -11,19 +11,14 @@ export default defineConfig(() => {
         /**
          * https://github.com/BuilderIO/qwik/issues/4952
          */
-        fileName: (format) => `index.qwik.${format === 'es' ? 'mjs' : 'cjs'}`,
+        fileName: (format, entryName) => `${entryName}.qwik.${format === 'es' ? 'mjs' : 'cjs'}`,
       },
       rollupOptions: {
-        external: ['@builder.io/qwik', 'node:module'],
+        external: ['@builder.io/qwik', 'node:module', 'isolated-vm'],
+        input: ['./src/index.ts', './src/functions/evaluate/node-runtime/init.ts'],
         output: {
-          manualChunks(id, { getModuleIds, getModuleInfo }) {
-            const moduleInfo = getModuleInfo(id);
-
-            // We need to move this node-only code to its own file so that `isServer` can tree-shake it.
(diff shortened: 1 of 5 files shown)


packages/sdks/src/functions/evaluate/node-runtime/node-runtime.ts
```diff
@@ -110,7 +111,7 @@ const getIvm = (): IsolatedVMImport => {
     
     SOLUTION: In a server-only execution path within your application, do one of the following:
   
-    ${SHOULD_MENTION_INITIALIZE_SCRIPT ? '- import and call `initializeNodeRuntime()` from "${SDK_NAME}/node/init".' : ''}
```
samijaber, line 113: oh no haha. did i really make this mistake...

thanks for catching it.

samijaber (approved):

packages/sdks/src/functions/evaluate/node-runtime/node-runtime.ts
sidmohanty11: i just thought my IDE was having trouble formatting it 😆

packages/sdks/output/qwik/vite.config.ts
```diff
@@ -11,19 +11,14 @@ export default defineConfig(() => {
         /**
          * https://github.com/BuilderIO/qwik/issues/4952
          */
-        fileName: (format) => `index.qwik.${format === 'es' ? 'mjs' : 'cjs'}`,
+        fileName: (format, entryName) => `${entryName}.qwik.${format === 'es' ? 'mjs' : 'cjs'}`,
       },
       rollupOptions: {
-        external: ['@builder.io/qwik', 'node:module'],
+        external: ['@builder.io/qwik', 'node:module', 'isolated-vm'],
+        input: ['./src/index.ts', './src/functions/evaluate/node-runtime/init.ts'],
```
samijaber, line 18: @sidmohanty11 just realized this means edge and browse bundles technically have `isolated-vm` imports.

We have to remove those to be safe and make sure no such imports exist in these two bundles

PR #3534, opened 2024-09-04
[snippet/react] - EDU 159 - React Gen2 examples in Custom Components 

uttej-vsk:
## Description

As per Steve's feedback, I have modified the React Gen2 snippets referring to the following pages with test cases:

- [Adding children to custom components](https://www.builder.io/c/docs/custom-components-children#adding-children-to-custom-components)
- [Adding advanced child sub-components](https://www.builder.io/c/docs/custom-components-children#adding-advanced-child-sub-components)

Editable Regions have already been added in the snippets with tested code. 

@samijaber, could you please review this and let me know if I need to modify them?

I will raise different PRs for the different frameworks.


packages/sdks/snippets/react/src/routes/custom-components/advanced-child.tsx
```diff
@@ -0,0 +1,92 @@
...
+                      },
+                    ],
+                    subFields: [{
+                      name: "blocks",
+                      type: 'uiBlocks'
+                    }],
+                  },
+                ],
+              },
+
```
samijaber, line 87: let's move this to a separate object called `customComponents`. We can also import and use the `RegisteredComponent` type from the SDK for it:

```ts
import { RegisteredComponent } from '@builder.io/sdk-react';

const customComponents: RegisteredComponent[] = [{
  //...
}]
```

packages/sdks/snippets/react/src/routes/custom-components/custom-child.tsx
```diff
@@ -0,0 +1,80 @@
...
+                defaultChildren: [
+                  {
+                    "@type": "@builder.io/sdk:Element",
+                    component: {
+                      name: "@builder.io/sdk:Element",
+                      options: { text: "I am child text block!" },
+                    },
+                  },
+                ],
+              },
```
samijaber, line 74: Similarly, let's move this

packages/sdks/snippets/react/src/routes/custom-components/custom-child.tsx
```diff
@@ -0,0 +1,80 @@
...
+  } from '@builder.io/sdk-react';
+  import { useEffect, useState } from 'react';
+import CustomHero from '../../components/CustomHero';
+  
+  const BUILDER_API_KEY = 'ee9f13b4981e489a9a1209887695ef2b';
+  const MODEL_NAME = 'page';
+  
+  // set whether you're using the Visual Editor,
+  // whether there are changes,
+  // and render the content if found
```
samijaber, line 21: I find this comment a bit confusing, especially "set whether you're using the Visual Editor". You're not actually setting that, you're just getting the value from `isPreviewing`. 

I would get rid of it completely: the comments throughout the component offer enough clarity IMO.

packages/sdks-tests/src/snippet-tests/advanced-child.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    test.skip(!['react'].includes(packageName));
+
+    await page.goto('/advanced-child');
+
+    await page.waitForSelector('div[builder-path]', { timeout: 5000 });
+
+    const div = page.locator('div[builder-path]');
+    const count = await div.count();
+
+    expect(count).toBeGreaterThan(1);
```
samijaber, line 15: if it's going to be exactly 2, can we check for that instead of greater than 1?

packages/sdks-tests/src/snippet-tests/advanced-child.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+test.describe('Adding advanced child blocks in custom components', () => {
+  test('render two divs with the attribute builder-path', async ({ page, packageName }) => {
+    test.skip(!['react'].includes(packageName));
+
+    await page.goto('/advanced-child');
+
+    await page.waitForSelector('div[builder-path]', { timeout: 5000 });
+
+    const div = page.locator('div[builder-path]');
```
samijaber, line 12: - we should avoid custom `timeout` values in playwright selectors
- we should avoid `waitForSelector` in general: https://playwright.dev/docs/api/class-elementhandle#element-handle-wait-for-selector
- you should be able to combine these into:

```suggestion
    const div = page.locator('div[builder-path]').waitFor();
```

packages/sdks/snippets/react/src/components/CustomColumns.tsx
```diff
@@ -0,0 +1,35 @@
+import { Blocks, BuilderBlock } from "@builder.io/sdk-react";
+import { Key } from "react";
+
+interface CustomColumnProps 
+{ 
+  inside: { 
+    blocks: BuilderBlock[] | undefined; 
+  }[];
```
samijaber, line 8: i find the `inside` here a bit unusual. Can we remove it and have it be `props.blocks` directly?

packages/sdks/snippets/react/src/routes/custom-components/advanced-child.tsx
uttej-vsk: What if we have 3 custom components (CustomHero, CustomTabs, CustomAdvancedChild)? Should I have to use the same customComponents function, or do I need to create 3 different files?

uttej-vsk: Update: 
I have added all objects into customComponents as suggested, and then I created one utility function to extract the required obj for a corresponding custom component like this: 

```import { RegisteredComponent } from '@builder.io/sdk-react';

export function extractRelevantComponentByName(name: string) {
  return function (arr: RegisteredComponent[]) {
    const filteredComponents = arr.filter(
      (component) => component.name === name
    );
    return filteredComponents.length > 0 ? filteredComponents : [];
  };
}
```

Please let me know if this is not what I'm supposed to do.

packages/sdks/snippets/react/src/components/customTabsInfo.ts
```diff
@@ -0,0 +1,60 @@
+import CustomTabs from './CustomTabs';
+
+export const customTabsInfo = {
+  component: CustomTabs,
+  name: 'TabFields',
+  /** To accept children in your custom component and by default it is false */
+  canHaveChildren: true,
```
samijaber, line 7: `canHaveChildren` should not be set for components that use `<Blocks>`. It should only be set to `true` for components that use `props.children`

samijaber: Otherwise you can drag'n'drop components that end up nowhere. See https://www.loom.com/share/5951d5e1f03649139e06bb6043899446?from_recorder=1&focus_title=1

packages/sdks/snippets/react/src/components/CustomColumns.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+interface CustomColumnsProps {
+  columns: { blocks: BuilderBlock[] | undefined }[];
+  builderBlock: { id: string | undefined };
+}
+
+const CustomColumns = (props: CustomColumnsProps) => {
+  return (
+    <div
+      className="two-columns"
```
samijaber, line 11: all of these `className`s that you have in custom components are unnecessary and should be removed.

packages/sdks/snippets/react/src/components/CustomColumns.tsx
```diff
@@ -0,0 +1,39 @@
+import { Blocks, BuilderBlock } from '@builder.io/sdk-react';
+
+interface CustomColumnsProps {
+  columns: { blocks: BuilderBlock[] | undefined }[];
+  builderBlock: { id: string | undefined };
```
samijaber, line 5: you can use `BuilderBlock` here:

```suggestion
  builderBlock: BuilderBlock;
```

packages/sdks/snippets/react/src/components/CustomColumns.tsx
```diff
@@ -0,0 +1,39 @@
...
+            parent={props.builderBlock.id}
+          />
+
+          <Blocks
+            blocks={props.columns[1]?.blocks}
+            path={`component.options.columns.1.blocks`}
+            parent={props.builderBlock.id}
+          />
+        </>
+      )}
```
samijaber, line 34: I think we can simplify this:

```suggestion
      <Blocks
        blocks={props.columns[0]?.blocks}
        path={`component.options.columns.0.blocks`}
        parent={props.builderBlock.id}
      />

      <Blocks
        blocks={props.columns[1]?.blocks}
        path={`component.options.columns.1.blocks`}
        parent={props.builderBlock.id}
      />
```

packages/sdks/snippets/react/src/components/CustomHero.tsx
```diff
@@ -0,0 +1,11 @@
+import { ReactNode } from 'react';
+
+interface CustomHeroProps {
+  children: ReactNode;
+}
+
+const CustomHero: React.FC<CustomHeroProps> = (props) => {
+  return <div className="Hero">{props.children}</div>;
```
samijaber, line 8: I would add some style or text in this component here so that it's not an empty plain container of Builder content

packages/sdks/snippets/react/src/components/CustomTabs.tsx
```diff
@@ -10,12 +10,12 @@ type Tab = {
   tabName: string;
   children: BuilderBlock[];
 };
-type TabProps = {
+interface TabProps {
   tabList: Tab[];
   builderBlock: {
     id: string | undefined;
   };
```
samijaber, line 17:
```suggestion
  builderBlock: BuilderBlock;
```

packages/sdks-tests/src/helpers/index.ts
```diff
@@ -117,6 +117,15 @@ export const verifyTabContent = async (
...
 };
 
+export const verifyBlocks = async (
+  page: Page,
+  element: Locator,
+  expectedVisibleContent: string
+): Promise<void> => {
+  const visibleContent = page.locator(`[builder-path="${expectedVisibleContent}"]`);
+  await expect(visibleContent).toBeVisible();
+};
```
samijaber, line 127: this is not being used and can be deleted

packages/sdks/snippets/react/src/components/CustomColumns.tsx
samijaber: this className is still here
```suggestion
```

packages/sdks/snippets/react/src/components/customColumnsInfo.ts
```diff
@@ -0,0 +1,26 @@
...
+        {
+          blocks: [],
+        },
+        {
+          blocks: [],
+        },
+      ],
+    },
+  ],
+};
```
samijaber, line 26: Can we merge each one of these component info files into component file itself? so move this to `CustomColumns.tsx`, etc.

uttej-vsk: done

packages/sdks/snippets/react/src/components/CustomTabs.tsx
```diff
@@ -71,4 +73,59 @@ const CustomTabs = (props: TabProps) => {
   );
 };
 
-export default CustomTabs;
+export const customTabsInfo: RegisteredComponent = {
+  component: CustomTabs,
+  name: 'TabFields',
+  /** To receieve Builder props inside your custom component: by default false  */
+  shouldReceiveBuilderProps: {
+    /** To access builder's Blocks relative to your parent */
```
samijaber, line 81: this isn't gramatically correct. Also, I find it weird that the comments start with `to`:

```suggestion
  shouldReceiveBuilderProps: {
    /** enabling this causes the SDK to pass the `builderBlock` prop down to the component */
```

samijaber (approved):

PR #3537, opened 2024-09-05
feat(react-gen1): support block level personalization and scheduling

teleaziz:
This PR introduces the `PersonalizationContainer` component, a new beta feature that allows users to personalize and schedule the release of specific parts of a page at the block level. This feature is currently controlled by a feature flag on the editor side `block-level-personaliztion`.

Key features:
- Scheduling: Ability to set start and end dates for each variant, enabling time-based content releases.
- Custom targeting: Users can define custom rules to show different content to different user segments.
- Preview mode: Editors can preview different variants directly in the Builder interface.
- Server-side rendering support: Ensures good performance and SEO compatibility, by always rendering default variant


samijaber (commented):
can we add integration tests for this to make sure it works as expected?

packages/core/src/builder.class.ts
```diff
@@ -639,6 +639,9 @@ export interface Input {
   model?: string;
 
   meta?: Record<string, any>;
+
+  // internal
+  behavior?: string;
```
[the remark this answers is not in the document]
teleaziz: it looks like historically we've only marked internals with `@hidden` so went with that in last commit

kylefowler (changes requested):
Can we get some integration tests for this change and a more in depth description?

teleaziz:
@kylefowler Tests and pr description added

packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@builder.io/sdk",
-  "version": "3.0.1",
+  "version": "3.0.2-2",
```
kylefowler, line 3: for my knowledge, what does the `-2` signify here?

teleaziz: That's a dev release

kylefowler: is it a common thing for us to have core packages depend on dev versions of the sdk?

kylefowler (approved):

packages/core/package.json
sidmohanty11: @teleaziz do you think we should add this? Because after we merge this PR, changeset bot will automatically bump versions and add a patch version and update that in our React gen1 SDK. I think we can keep it `3.0.1` here and for React gen1 - `5.0.1` as it will auto bump it to `3.0.2` and `5.0.2` respectively and link that to the gen1 React SDK (`5.0.2`). With this change, it might skip `3.0.2` and create `3.0.3` and I think it would've also updated yarn.lock here? We can skip the changesets if we are planning to only dev-release it

teleaziz: @kylefowler It's not common or recommended for core packages to depend on dev versions of the SDK. In this specific PR, nothing is actually depending on a dev version of another library. The change to version "3.0.2-2" is intended to test the core SDK standalone.

Using a dev version (like "3.0.2-2") for the SDK itself is a good practice for testing changes before a final release. However, other packages, especially core ones, typically shouldn't rely on these unstable dev versions.

The purpose here is to validate the SDK changes in isolation, ensuring they work as expected.


@sidmohanty11 The correct sequence would be from "3.0.2-2" to the final "3.0.2" release, not to "3.0.3". In semantic versioning, the "-2" suffix indicates a pre-release version. When finalizing, this suffix is removed to create the stable release. The version would only increment to 3.0.3 if additional changes were made after the 3.0.2 release

sidmohanty11: @teleaziz thanks for explaining the semantic sequencing concept! Makes sense. stupid me, i thought changeset bot will +1 it.

packages/core/src/builder.class.ts
```diff
@@ -118,7 +117,6 @@ function setCookie(name: string, value: string, expires?: Date) {
       (value || '') +
       expiresString +
       '; path=/' +
-      `; domain=${getTopLevelDomain(location.hostname)}` +
```
samijaber, line 121: is this not a significant change? we're no longer setting `domain` on all of our cookies. surprised to see as part of the personalization feature.

PR #3542, opened 2024-09-09
Supp 657

STRd6:
## Description

Scope carousel CSS styles so they won't interfere with other styles on the page.

https://builder-io.atlassian.net/jira/software/c/projects/PLAN/boards/52?selectedIssue=SUPP-657

### DIFF
--- a/packages/widgets/src/components/Carousel.tsx
+++ b/packages/widgets/src/components/Carousel.tsx
@@ -215,7 +215,280 @@ export class CarouselComponent extends React.Component<CarouselProps> {
 }
 
 const slickStyles = `@charset 'UTF-8';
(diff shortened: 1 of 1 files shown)


sidmohanty11 (commented):
LGTM. For the widgets SDK, we don't have a changeset pipeline set up yet, so we'll need to manually deploy it to npm. To do that, we can use `npm run release:patch`.

After that, we'll need to update the dependencies in the webcomponents SDK here: https://github.com/BuilderIO/builder/blob/f2e7923f08b621965e6500499cd09906cfe1ed34/packages/webcomponents/package.json#L149 and publish it.

[_After we publish the webcomponents SDK_] We can ask customers to point to the latest version, for example by using `https://cdn.builder.io/js/webcomponents@1.3.53`, and later update the `STABLE_VERSION` for the Lambda function. You can follow this: https://www.notion.so/builderio/Make-changes-deploy-AWS-Lambda-Edge-function-d2797b9c0300407c952104f38d5bffa0.

samijaber (approved):
@STRd6 as @sidmohanty11 mentioned, you need to follow the release process for this to get it out the door.

PR #3551, opened 2024-09-12
feat[blocks][className]: ENG-6614 add className to Blocks

clyde-builderio:
## Description

Add `className: string` prop support to `Blocks`

_Screenshot_

### DIFF
--- a/packages/sdks-tests/src/e2e-tests/blocks-class-name.spec.ts
+++ b/packages/sdks-tests/src/e2e-tests/blocks-class-name.spec.ts
@@ -1,0 +1,21 @@
+import { expect } from '@playwright/test';
+import { excludeGen1, excludeTestFor, test } from '../helpers/index.js';
+
+test.describe.only('Blocks className', () => {
+  test('call proxy API only once - in page', async ({ page, packageName, sdk }) => {
+    test.skip(excludeGen1(sdk));
+    test.fail(
+      packageName === 'nextjs-sdk-next-app',
+      'editor tests not supported in nextjs-sdk-next-app'
+    );
+    test.fail(
+      excludeTestFor({ qwik: true }, sdk),
+      'error setting httpRequest response or making API call altogether.'
+    );
+    test.skip(excludeTestFor({ vue: true }, sdk), 'Vue flakiness issues');
+
+    await page.goto('/blocks-class-name', { waitUntil: 'networkidle' });
+    const count = await page.locator('.builder-blocks.test-class-name').count();
+    expect(count).toBe(1);
+  });
+});
--- a/packages/sdks/src/components/blocks/blocks-wrapper.lite.tsx
+++ b/packages/sdks/src/components/blocks/blocks-wrapper.lite.tsx
@@ -30,13 +30,20 @@ export type BlocksWrapperProps = {
   BlocksWrapperProps: any;
 
   children?: any;
+
+  className?: string;
 };
 
 export default function BlocksWrapper(props: BlocksWrapperProps) {
+  console.log('props.className', props.className);
   const blocksWrapperRef = useRef<HTMLDivElement>();
   const state = useStore({
     get className() {
-      return 'builder-blocks' + (!props.blocks?.length ? ' no-blocks' : '');
+      return (
+        'builder-blocks' +
+        (!props.blocks?.length ? ' no-blocks' : '') +
+        (props.className ? ' ' + props.className : '')
+      );
     },
     onClick() {
       if (isEditing() && !props.blocks?.length) {
--- a/packages/sdks-tests/src/specs/blocks-class-name.ts
+++ b/packages/sdks-tests/src/specs/blocks-class-name.ts
@@ -1,0 +1,89 @@
+export const BLOCKS_CLASS_NAME = {
+  ownerId: 'b3ef520ea40045149c66fcb6e6a4adec',
+  lastUpdateBy: null,
+  createdDate: 1726138498867,
+  id: 'b3ef520ea40045149c66fcb6e6a4adec_dfbbf7e36abc4937b23f5cdf005eb511',
+  '@version': 4,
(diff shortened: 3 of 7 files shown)


packages/sdks/e2e/angular-ssr/src/app/catch-all.component.ts
```diff
@@ -48,6 +49,46 @@ export class CatchAllComponent {
...
+                  position: 'relative',
+                  flexShrink: '0',
+                  boxSizing: 'border-box',
+                  marginTop: '20px',
+                  lineHeight: 'normal',
+                  height: 'auto',
+                },
+              },
+            },
+          ],
```
sidmohanty11, line 88: could we move this as `DEFAULT_TEXT_ELEMENT` and export it from sdk-tests (see how getProps is used) and use it as the `defaultValue` everywhere?

sidmohanty11 (commented):
Amazing work!! 🚀  Thank you so much for adding custom component tests for each framework, this will help us a lot in the long run. 

Last step would be to add a changeset for this for all the gen2 SDKs except react native, you can do that by running `yarn g:changeset` at the root

packages/sdks/src/components/blocks/blocks-wrapper.lite.tsx
```diff
@@ -30,13 +30,19 @@ export type BlocksWrapperProps = {
...
 export default function BlocksWrapper(props: BlocksWrapperProps) {
   const blocksWrapperRef = useRef<HTMLDivElement>();
   const state = useStore({
     get className() {
-      return 'builder-blocks' + (!props.blocks?.length ? ' no-blocks' : '');
+      return (
+        'builder-blocks' +
+        (!props.blocks?.length ? ' no-blocks' : '') +
+        (props.classNameProp ? ' ' + props.classNameProp : '')
+      );
```
samijaber, line 45: it'll be easier to use an array so you don't have to remember the `' '` spaces each time:

```suggestion
      return [
        'builder-blocks',
        !props.blocks?.length ? 'no-blocks' : '',
        props.classNameProp,
      ]
        .filter(Boolean)
        .join(' ');
```

packages/sdks-tests/src/e2e-tests/blocks-class-name.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+test.describe('Blocks className', () => {
+  test('should be present on element with class builder-blocks', async ({
+    page,
+    sdk,
+    packageName,
+  }) => {
+    test.skip(excludeGen1(sdk));
+    test.skip(packageName === 'react-native');
+
+    await page.goto('/blocks-class-name', { waitUntil: 'networkidle' });
```
samijaber, line 13: why the `{ waitUntil: 'networkidle' }`?

samijaber (commented):
looks good! you're missing the changesets though. run `yarn g:changeset` to create them

packages/sdks/e2e/angular-ssr/src/app/catch-all.component.ts
samijaber: yeah, if you want to reduce all the repetition in each e2e test, you could add a `getBuilderBlockWithClassNameCustomComponentInfo` helper in https://github.com/BuilderIO/builder/blob/main/packages/sdks-tests/src/specs/index.ts

```ts
const getBuilderBlockWithClassNameCustomComponentInfo = (component) => ({
  component,
  name: 'BuilderBlockWithClassName',
  shouldReceiveBuilderProps: {
    builderBlock: true,
  },
 // ...
})
```

not required though, but would cleanup this PR a bunch

samijaber (commented):
please add changests as per https://github.com/BuilderIO/builder/blob/main/packages/sdks/PUBLISHING.md#1--add-changeset

packages/sdks/e2e/sveltekit/src/routes/[...catchall]/+page.svelte
```diff
@@ -1,12 +1,57 @@
...
+  };
 </script>
 
 <main>
   {#if data.props}
     {#key data.props}
-      <Content {...data.props} customComponents={data.customComponents} />
+      <Content
+        {...data.props}
+        customComponents={[builderBlockWithClassNameCustomComponent]}
```
Comment (samijaber) [inline packages/sdks/e2e/sveltekit/src/routes/[...catchall]/+page.svelte:53]: in here you are overriding `customComponents` coming from `data.props`

you need to add the component here https://github.com/BuilderIO/builder/blob/8b372561e2f8e1cd3c5648a71183934d35992f11/packages/sdks/e2e/sveltekit/src/routes/%5B...catchall%5D/+page.ts#L33-L44

packages/sdks/src/components/blocks/blocks-wrapper.lite.tsx
```diff
@@ -30,13 +30,21 @@ export type BlocksWrapperProps = {
   BlocksWrapperProps: any;
 
   children?: any;
+
+  classNameProp?: string;
```
samijaber, line 34: you can name this `className` now that the fix is made

packages/sdks/src/components/blocks/blocks.types.ts
```diff
@@ -6,9 +6,13 @@ import type {
...
-  Omit<BlocksWrapperProps, 'BlocksWrapper' | 'BlocksWrapperProps'>
+  Omit<
+    BlocksWrapperProps,
+    'BlocksWrapper' | 'BlocksWrapperProps' | 'classNameProp'
+  >
 > & {
   context?: Signal<BuilderContextInterface>;
   registeredComponents?: RegisteredComponents;
   linkComponent?: any;
+  className?: string;
```
samijaber, line 17: if you rename `classNameProp` to `className`, you should be able to undo the changes in this file, since the `className` prop in both cases will match, so it will flow through.

samijaber (approved):

PR #3558, opened 2024-09-19
feat(personalization-utils): add trimHtml for dynamic container…

teleaziz:
…s processing on the edge`

## Add trimHtml function for dynamic personalization containers

This PR introduces the `getTrimmedHtml` function, a new utility for processing dynamic personalization containers in HTML content at the edge or in server-side rendering scenarios.
- Enables edge-side and server-side processing of personalized content
- Allows for caching or static generation of personalized pages

### DIFF
--- a/packages/personalization-utils/README.md
+++ b/packages/personalization-utils/README.md
@@ -73,14 +73,14 @@ builder.setUserAttributes({ audience })
 
 Once the cookie is set, all builder content matching from now on will weigh in the current audience segment.
 
-## Using getTrimmedHTML for Dynamic Containers
+## Using trimHtml for Dynamic Containers
 
-The `getTrimmedHTML` function is a utility for handling dynamic personalization containers in your HTML content. It's particularly useful for processing personalized content at the edge or in server-side rendering scenarios.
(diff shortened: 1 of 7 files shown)


steve8708:
Hell yes! Wdyt of just calling it trimHtml()!

packages/personalization-utils/README.md
```diff
@@ -65,12 +65,48 @@ export default function middleware(request) {
...
-  setCookie(`builder.userAttributes.audience`, audience)
+```typescript
+const audience = await myCDP.identifyAudience(userID);
+// this will include the `audience` in all api calls and save it in a cookie `builder.userAttributes`
+builder.setUserAttributes({ audience })
 ```
+
 Once the cookie is set, all builder content matching from now on will weigh in the current audience segment.
 
+## Using trimHtml for Dynamic Containers
```
mrkoreye, line 76: great docs 👍

packages/personalization-utils/src/get-trimmed-html.test.ts
```diff
@@ -0,0 +1,174 @@
+import { trimHtml } from './utils'; // Update this import to match your file structure
+
+describe('trimHtml', () => {
```
mrkoreye, line 3: yes test!

mrkoreye (approved):
this is a great idea

PR #3560, opened 2024-09-19
fix[vue-gen2]: isolated-vm fix workaround for Vue SDK

### LINKED ISSUE #3288: "isolated-vm import error" on Vercel with Nuxt 3
**Describe the bug**
I see the following errors in the Vercel log after visiting a builder.io page (the page seems to render fine though):

```txt
[Builder.io]:  isolated-vm import error. Error: Cannot find module 'isolated-vm'
Require stack:
- /var/task/node_modules/@builder.io/sdk-vue/lib/node/index-551oCh4n.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1140:15)
    at Module._load (node:internal/modules/cjs/loader:981:27)
    at /opt/rust/nodejs.js:1:11508
    at Function.Qt (/opt/rust/nodejs.js:1:11878)
    at Q.e.<computed>.K._load (/opt/rust/nodejs.js:1:11478)
    at Module.require (node:internal/modules/cjs/loader:1231:19)
    at require (node:internal/modules/helpers:177:18)
    at getIvm (file:///var/task/node_modules/@builder.io/sdk-vue/lib/node/index-551oCh4n.js:318:15)
    at runInNode (file:///var/task/node_modules/@builder.io/sdk-vue/lib/node/index-551oCh4n.js:350:13)
    at chooseBrowserOrServerEval (file:///var/task/node_modules/@builder.io/sdk-vue/lib/node/index-551oCh4n.js:391:108) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/var/task/node_modules/@builder.io/sdk-vue/lib/node/index-551oCh4n.js'
  ]
}
```

followed by

```
[Builder.io]:  Failed code evaluation: [Builder.io]: could not import `isolated-vm` module for safe script execution on Node server.
    
    In certain Node environments, the SDK requires additional initialization steps. This can be achieved by 
    importing and calling `initializeNodeRuntime()` from "@builder.io/sdk-react/node/init". This must be done in
    a server-only execution path within your application.

    Please see the documentation for more information: https://builder.io/c/docs/integration-tips#enabling-data-bindings-in-node-environments
     {
  code: 'var __awaiter=function(e,t,n,r){return new(n||(n=Promise))((function(i,o){function a(e){try{c(r.next(e))}catch(e){o(e)}}function u(e){try{c(r.throw(e))}catch(e){o(e)}}function c(e){var t;e.done?i(e.value):(t=e.value,t instanceof n?t:new n((function(e){e(t)}))).then(a,u)}c((r=r.apply(e,t||[])).next())}))},__generator=function(e,t){var n,r,i,o,a={label:0,sent:function(){if(1&i[0])throw i[1];return i[1]},trys:[],ops:[]};return o={next:u(0),throw:u(1),return:u(2)},"function"==typeof Symbol&&(o[Symbol.iterator]=function(){return this}),o;function u(o){return function(u){return function(o){if(n)throw new TypeError("Generator is already executing.");for(;a;)try{if(n=1,r&&(i=2&o[0]?r.return:o[0]?r.throw||((i=r.return)&&i.call(r),0):r.next)&&!(i=i.call(r,o[1])).done)return i;switch(r=0,i&&(o=[2&o[0],i.value]),o[0]){case 0:case 1:i=o;break;case 4:return a.label++,{value:o[1],done:!1};case 5:a.label++,r=o[1],o=[0];continue;case 7:o=a.ops.pop(),a.trys.pop();continue;default:if(!(i=(i=a.trys).length>0&&i[i.length-1])&&(6===o[0]||2===o[0])){a=0;continue}if(3===o[0]&&(!i||o[1]>i[0]&&o[1]<i[3])){a.label=o[1];break}if(6===o[0]&&a.label<i[1]){a.label=i[1],i=o;break}if(i&&a.label<i[2]){a.label=i[2],a.ops.push(o);break}i[2]&&a.ops.pop(),a.trys.pop();continue}o=t.call(e,a)}catch(e){o=[6,e],r=0}finally{n=i=0}if(5&o[0])throw o[1];return{value:o[0]?o[1]:void 0,done:!0}}([o,u])}}};function main(){return __awaiter(this,void 0,void 0,(function(){var e,t,n;return __generator(this,(function(r){return Builder.isServer,Builder.isBrowser&&(e=document.getElementById("toggleSwitch"),t=document.getElementById("price"),n=document.getElementById("membership-link"),e.addEventListener("change",(function(){this.checked?(t.textContent="0000",n.href="https://example.com"):(t.textContent="0000",n.href="https://example.com")}))),[2]}))}))}var _virtual_index=main();return _virtual_index'
}
```

The link provided in the 2nd error message only provides a solution for Next.js (React), not for Nuxt (Vue).

**To Reproduce**
1. Visit our page
2. see the errors in the log

**Expected behavior**
No errors in the logs?

**Screenshots**
image

**Additional context**
Happens on `@builder.io/sdk-vue` v1.0.22 as well as v1.0.27 (latest version) on the current Nuxt3 release.

Our main page template (`[...slug].vue`) loading the content:

```vue filename="[...slug].vue"
<template>
  <div>
    <div v-if="content || isPreviewing()">
      <Content
        model="page"
        :can-track="false"
        :content="content"
        :api-key="builderApiKey"
        :custom-components="REGISTERED_COMPONENTS" />
    </div>
  </div>
</template>

<script setup>
import { Content, fetchOneEntry, isPreviewing } from "@builder.io/sdk-vue";
import FaqItems from "~/components/block/FaqItems.vue";
import Drawer from "~/components/builder.io/Drawer.vue";
import { useCustomHead } from "~/composables/useCustomHead";

// Register your Builder components
const REGISTERED_COMPONENTS = [
  {
    canHaveChildren: false,
    component: FaqItems,
    inputs: [
      {
        name: "faqItems",
        required: true,
        type: "object",
      },
      {
        name: "title",
        type: "text",
      },
    ],
    name: "BlockFaqItems",
  },
  {
    canHaveChildren: true,
    component: Drawer,
    inputs: [
      {
        name: "id",
        required: true,
        type: "text",
      },
    ],
    name: "Drawer",
  },
];

const config = useRuntimeConfig();
const builderApiKey = ref(config.public.builderApiToken);
const isDev = import.meta.dev;
// the app url is also set to example.com on preview environments
const isStaging = config.public.appUrl === "https://example.com";

const route = useRoute();

// fetch builder content data
const { data: content } = await useAsyncData("builderData", () =>
  fetchOneEntry({
    apiKey: builderApiKey.value,
    model: "page",
    options: {
      includeUnpublished: isPreviewing() || isDev || isStaging,
    },
    userAttributes: {
      urlPath: route.path,
    },
  }),
);

if (content.value === null) {
  throw createError({
    fatal: true,
    statusCode: 404,
    statusMessage: "Page Not Found",
  });
}

useCustomHead({
  metaDescription: content.value.data.description,
  metaImage: {
    data: {
      attributes: {
        url: content.value.data.previewImage,
      },
    },
  },
  metaTitle: content.value.data.title,
});
</script>
```

sidmohanty11:
## Description

Adds a workaround to fix `isolated-vm` import issue in gen2 Vue SDK

**Jira**
https://builder-io.atlassian.net/browse/ENG-6177

Fixes #3288

Repo hosted on Vercel with bindings working: https://github.com/sidmohanty11/nuxt-app-isolated-vm
https://nuxt-app-isolated-vm.vercel.app/


packages/sdks/output/vue/vite.config.js
```diff
@@ -19,17 +19,22 @@ export default defineConfig({
...
+        copyFileSync(outPath + '/functions/evaluate/node-runtime/init.d.ts', outPath + '/init.d.cts');
       },
     }),
   ],
   build: {
     lib: {
-      entry: resolve(__dirname, 'src/index.ts'),
+      entry: {
+        index: resolve(__dirname, 'src/index.ts'),
+        init: resolve(__dirname, 'src/functions/evaluate/node-runtime/init.ts')
```
samijaber, line 31: the issue with this approach is that it also creates `lib/edge/init.mjs` and `lib/browser/init.mjs`, both of which will import `isolated-vm`. This can cause issues and confusion.

We want to makes sure the ivm import is nowhere within those two bundles.


'CleanShot 2024-09-30 at 15 02 24@2x

samijaber: you can workaround this by checking the `SDK_ENV` and only adding this entry point for `node`

samijaber (changes requested):
I think we can go one step further:

- define the Nuxt plugin as part of the SDK and export it, so customers don't have to write a plugin by hand
- improve our Nuxt module so it has inline options: (see how they're configured by the user here https://nuxt.com/docs/guide/concepts/modules#add-nuxt-modules)
```ts
type Options = {
  includeCompiledCss?: boolean; // default: true
  initializeNodeRuntime?: boolean; // default: false
}
```
- enable the Nuxt plugin _via_ the Module we already have when `initializeNodeRuntime` is `true`

packages/sdks/output/vue/nuxt-isolated-vm-plugin.js
```diff
@@ -0,0 +1,12 @@
+export default defineNuxtPlugin((nuxtApp) => {
```
samijaber, line 1: - this works without needing to import `defineNuxtPlugin`?
- one last thing is to add an `exports` field for this plugin, so that customers can import it directly if they'd like. could be under `@builder.io/sdk-vue/nuxt-initialize-node-runtime-plugin`

sidmohanty11: i haven't wrapped up this yet, will ping you once I test this out. got caught up with the Angular thing

sidmohanty11:
Thanks @samijaber for the amazing idea, now it's down to just this for Nuxt:
```typescript
  modules: [
    [
      "@builder.io/sdk-vue/nuxt",
      {
        includeCompiledCss: true,
        initializeNodeRuntime: true,
      },
    ],
  ],
```

Repo (code and hosted): https://github.com/sidmohanty11/nuxt-app-isolated-vm

packages/sdks/output/vue/vite.config.js
sidmohanty11: thanks for the catch, done! we also need to do this for React SDK and Qwik SDK next

.changeset/calm-deers-serve.md
```diff
@@ -0,0 +1,23 @@
+---
+'@builder.io/sdk-vue': patch
+---
+
+Feature: add `@builder.io/sdk-vue/node/init` entry point with `initializeNodeRuntime` export that sets the IVM instance.
```
samijaber, line 5: - I would describe it in more detail, show how it can be used manually, like https://github.com/BuilderIO/builder/blob/634a0f31f87aca232b719d54111db2578ae3200f/packages/sdks/output/react/CHANGELOG.md?plain=1#L231-L274

and mention the new plugin:

> Exports a Nuxt Plugin `nuxt-initialize-node-runtime-plugin` that initializes `isolated-vm` in node runtimes. Here's how you can use it:

(and show an example)

and then explicitly mention module updates:

> Updated `'@builder.io/sdk-vue/nuxt'` Nuxt module:
> - "added `includeCompiledCss` flag that adds the compiled Builder.io CSS in Nuxt (defaults to `true`)
> - "added `initializeNodeRuntime` flag that executes `nuxt-initialize-node-runtime-plugin` plugin(defaults to `false`)

samijaber (changes requested):
Sorry, one thing I forgot to ask for: update this for the error message in Vue SDK:

https://github.com/BuilderIO/builder/blob/634a0f31f87aca232b719d54111db2578ae3200f/packages/sdks/src/functions/evaluate/node-runtime/node-runtime.ts#L95-L99

.changeset/calm-deers-serve.md
sidmohanty11: updated, please take a look

samijaber (approved):
awesome 👍🏽

PR #3564, opened 2024-09-23
fix[resourcePicker][SFCC]: Initialization bug for Picker Modal when Targeting Multi Select input types

anaghav2023:
## Description

The multi-select pickers for SF Commerce Products and SF Commerce Categories do not update properly when selecting an item for the first time (coming from an empty targeting menu). This does not happen when using this input type as a data model input. 

Once the array has data in it, reopening the picker works as expected, and any additions / removals are instantly updated in the UI.

**Loom**
https://www.loom.com/share/9278b07dac9549699213c4653ee8bbd3

### DIFF
--- a/packages/plugin-tools/src/editors/ResourcesList.tsx
+++ b/packages/plugin-tools/src/editors/ResourcesList.tsx
@@ -1,7 +1,7 @@
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
 import React, { useEffect } from 'react';
-import { action } from 'mobx';
+import { action, toJS } from 'mobx';
 import { useObserver, useLocalStore } from 'mobx-react';
 import {
   ResourcePreviewCell,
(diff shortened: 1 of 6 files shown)


packages/plugin-tools/src/editors/ResourcesList.tsx
```diff
@@ -135,7 +136,10 @@ export function PickResourceList(props: PickResourceListProps) {
                 omitIds={store.value}
                 onChange={action(resource => {
                   if (resource) {
-                    props.onChange([...(store.value || []), String(resource.id)]);
+                    if(store.value){
+                      store.value.push(String(resource.id));
+                    }
+                    props.onChange(store.value);
```
sanyamkamat, line 142: The previous code returned `String(resource.id)` with or without there being value from prop. Now we are adding it conditionally. Also, it feels like store.value will mostly _always_ be `true`.
If we know we always get an Array in the `value` property then the check is unnecessary. If it can be anything other than Array, then I would update the code to 
```
if (!Array.isArray(store.value)) {
  store.value = [];
}

store.value.push(String(resource.id));
props.onChange(store.value);

```

sanyamkamat (approved):

PR #3565, opened 2024-09-23
fix[widgets]: SUPP-653 extra inline-block styles getting applied

sidmohanty11:
## Description
React Slick applies extra styles - inline-block and width: 100% to each item. By wrapping them in a fragment, no extra styles are applied.

Ref: https://github.com/akiran/react-slick/issues/1378#issuecomment-636649924

Jira
https://builder-io.atlassian.net/browse/SUPP-653

Loom
https://www.loom.com/share/a486a37834d246eaaa84b40ff39468ae


mrkoreye (approved):

samijaber (commented):
hmm this will technically need to be a major change, since it alters the layout of the children

sidmohanty11:
@samijaber I think we can release this: https://github.com/BuilderIO/builder/pull/3542 first as a patch bump and then release this fix as the next major release, what do you think?

PR #3572, opened 2024-09-24
fix[gen1-react]: SUPP-685 embedded iframe reloads when page orientation changes

sidmohanty11:
## Description

Component was re-rendering triggering a reload to the iframe when page orientation changes incase of Next.js

Jira
https://builder-io.atlassian.net/browse/ENG-6882

Loom
https://www.loom.com/share/0d40545eb22d4fb9b796f08c3cb9aeeb

### DIFF
--- a/.changeset/smooth-chefs-destroy.md
+++ b/.changeset/smooth-chefs-destroy.md
@@ -2,4 +2,4 @@
 "@builder.io/react": patch
 ---
 
-Fix: embed or custom code video reloading when page orientation changes
+Fix: prevent Embed and Custom Code blocks from re-rendering when page orientation changes
--- a/.changeset/smooth-chefs-destroy.md
+++ b/.changeset/smooth-chefs-destroy.md
@@ -1,0 +1,5 @@
+---
+"@builder.io/react": patch
+---
+
+Fix: embed or custom code video reloading when page orientation changes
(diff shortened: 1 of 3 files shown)


packages/react/src/blocks/CustomCode.tsx
```diff
@@ -90,6 +90,10 @@ class CustomCodeComponent extends React.Component<Props> {
     }
   }
 
+  shouldComponentUpdate(nextProps: Readonly<Props>): boolean {
+    return nextProps.code !== this.props.code;
+  }
```
samijaber, line 95: we should be able to test this with a content entry that has a custom code block that increments a `state.foo`. when the bug is present, it should update it twice, and when the fix is there it should update once.

sidmohanty11: i was only able to reproduce this issue when I was using gen1 React SDK in a next.js app directory project. If we use pages directory or even vanilla react page orientation doesn't trigger a re-render

.changeset/smooth-chefs-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@builder.io/react": patch
+---
+
+Fix: embed or custom code video reloading when page orientation changes
```
samijaber, line 5:
```suggestion
Fix: prevent Embed and Custom Code blocks from re-rendering when page orientation changes
```

samijaber (approved):

PR #3575, opened 2024-09-25
feat(personaliztion-utils): Add A/B Test Trimming Support to trimHtml Function

teleaziz:
This PR extends the `trimHtml` function to support trimming A/B test variants in addition to personalization containers.

### Changes:

- Updated `trimHtml` function to process A/B test variants before personalization containers
- Added `abTests` parameter to `trimHtml` options
- Implemented logic to select winning A/B test variants based on provided `abTests` object
- Updated README with information on A/B test trimming and usage examples

### DIFF
--- a/packages/personalization-utils/README.md
+++ b/packages/personalization-utils/README.md
@@ -91,7 +91,7 @@ const abTests = {
   'content-id-2': 'variant-b'
 };
 
-const trimmedHTML = trimHtml(fullHTML, { userAttributes, abTests });
+const { html } = trimHtml(fullHTML, { userAttributes, abTests });
 ```
 
 To get the `userAttributes`, you should parse the `builder.userAttributes` cookie. Here's an example of how you might do this:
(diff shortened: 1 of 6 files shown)


steve8708:
awesome! I wonder if we should also return an object so if we want to return more stuff in the future its easier. e.g. 

```js
const { html } = trimHtml({ userAttributes: { ... } })
```

e.g. if we want to setup something that auto dice rolls for a/b tests that you aren't assigned to yet, we ideally want to make sure you set that as a cookie, so we could always return like

```js
const { html, newCookies } = ...
```

if we wanted. maybe this is already accounted for, but just an idea

samijaber (commented):
when I read `trimHtml` I think of a utility that trims excess whitespace.

Could we name it `personalizeHTML` or something like that, which more explicitly describes that it's going to process/handle personalization?

steve8708:
Open to names, but this isn't specific to personalization. It handles scheduling, ab testing, and potentially more in future

mrkoreye (approved):

PR #3593, opened 2024-10-03
fix[angular-gen2]: slot placement in dynamic components in Angular

sidmohanty11:
## Description

`ngAfterContentInit` makes sure that we have access to children and project them in a slot

Ref: https://stackoverflow.com/questions/36545478/angular-2-access-ng-content-within-component/53527323#53527323

_Screenshot_

### DIFF
--- a/packages/sdks-tests/src/e2e-tests/custom-components.spec.ts
+++ b/packages/sdks-tests/src/e2e-tests/custom-components.spec.ts
@@ -33,6 +33,14 @@ test.describe('Custom components', () => {
     await customComponentMsgPromise;
   });
 
+  test('children placement is correct', async ({ page, sdk }) => {
+    test.skip(!['angular'].includes(sdk));
+    await page.goto('/children-slot-placement');
+    const h1 = page.locator('h1').first();
+    await expect(h1).toBeVisible();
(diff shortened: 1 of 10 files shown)


packages/sdks-tests/src/e2e-tests/custom-components.spec.ts
```diff
@@ -33,6 +33,14 @@ test.describe('Custom components', () => {
     await customComponentMsgPromise;
   });
 
+  test('children placement is correct', async ({ page, sdk }) => {
+    test.skip(!['angular'].includes(sdk));
```
samijaber, line 37: shouldn't this test work on all SDKs?

samijaber: oh nevermind. its testing a custom component

samijaber (approved):

packages/sdks-tests/src/e2e-tests/custom-components.spec.ts
sidmohanty11: the `hello` custom component is there only in Angular SDK as of now, but yes, the test should pass in all SDKs

PR #3602, opened 2024-10-04
Blog Article snippets with Angular Gen2 SSR

uttej-vsk:
## Description

Blog Article snippets in the Blueprints now include Angular Gen 2 SSR capabilities with this PR.


packages/sdks/snippets/angular-ssr/src/app/blog-article/blog-article.component.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+    });
+    fixture = TestBed.createComponent(BlogArticleComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
```
samijaber, line 21: I wouldn't add specs like these: all they do is check that the component rendered. it's unnecessary: we already have a comprehensive integration test checking the behaviour of the rendered page, which is more useful and applies to every snippet. This one only exists in one snippet (it's also not running in CI)

packages/sdks/snippets/angular-ssr/src/app/blog-article/blog-article.component.html
```diff
@@ -0,0 +1 @@
+<p>blog-article works!</p>
```
samijaber, line 1: what is this? do we need this? can we remove it?

I want to remove any redundant file from our snippets to keep them as easy to read as possible

uttej-vsk: Ah! the boiler plate file. I just removed it.

packages/sdks/snippets/angular-ssr/src/app/blog-article/blog-article.component.spec.ts
uttej-vsk: @samijaber My bad, I forgot to remove these boilerplate files generated by CLI. I didn't add any test cases there. I will remove it and push it again.

packages/sdks/snippets/angular-ssr/src/app/blog-article/blog-article.component.ts
```diff
@@ -0,0 +1,23 @@
+import { CommonModule } from '@angular/common';
+import { Component } from '@angular/core';
+import { ActivatedRoute } from '@angular/router';
+import { Content, type BuilderContent } from '@builder.io/sdk-angular';
+
+@Component({
+  selector: 'app-blog-article',
+  standalone: true,
+  imports: [Content, CommonModule],
+  templateUrl: './blog-article.component.html',
```
samijaber, line 10: i just realized you're using a template `HTML` file instead of inlining it here.

Can we inline the `template` here instead? The reason is that we want to provide the entire blueprint in a single code block. By having a separate HTML file, this will require 2 code snippets which will be more work for us to sync when we have the script.

See other components in this codebase to see how we do this

uttej-vsk: @samijaber Sure, Sami, this is very doable. I have done this to make the code clean. But as you suggested, I have reverted it to inline and pushed the code. Kindly review :)

packages/sdks/snippets/angular-ssr/src/app/blog-article/blog-article.component.ts
```diff
@@ -0,0 +1,32 @@
...
+import { Content, type BuilderContent } from '@builder.io/sdk-angular';
+
+@Component({
+  selector: 'app-blog-article',
+  standalone: true,
+  imports: [Content, CommonModule],
+  template: `<div *ngIf="article" class="content">
+    <h1>{{ article.data?.title }}</h1>
+    <p>{{ article.data?.['blurb'] }}</p>
+    <img [src]="article.data?.['image']" alt="" />
```
sidmohanty11, line 13: can we just use
```
article.data?.blurb
article.data?.image
```
and give img `alt=article.data?.title` maybe?

sidmohanty11 (approved):

PR #3610, opened 2024-10-07
feat[svelte-gen2]: dynamic loading of components

sidmohanty11:
## Description

Able to dynamically load custom components in Svelte SDK by,

```html
<script lang="ts">
  import LoadingSpinner from '../../components/LoadingSpinner.svelte'; // Fallback loader component
  import type { RegisteredComponent } from '@builder.io/sdk-svelte';
  const customComponents: RegisteredComponent[] = [
    {
      name: 'LazyComponent',
      component: {
        // dynamically loads the custom component only when it needs to be initialized
        load: () => import('../../components/LazyComponent.svelte'),
        fallback: LoadingSpinner,
      },
    },
  ];
</script>

<Content
    content={data.content}
    apiKey={'YOUR_KEY'}
    model="page"
    customComponents={customComponents}
/>  
```

_Screenshot_

### DIFF
--- a/packages/sdks/src/context/types.ts
+++ b/packages/sdks/src/context/types.ts
@@ -6,7 +6,14 @@ import type { Dictionary, Nullable } from '../types/typescript.js';
 import type { ExtraContextTypes } from './extra-context-types.js';
 
 export type RegisteredComponent = ComponentInfo & {
-  component: any;
+  component: /**
+   * Dynamically load components (currently only available for Svelte SDK)
+   */
+  | {
+        load: () => Promise<any>;
+        fallback?: any;
+      }
+    | any;
 };
 
 export type RegisteredComponents = Dictionary<RegisteredComponent>;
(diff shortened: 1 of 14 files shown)


packages/sdks/src/context/types.ts
```diff
@@ -6,7 +6,14 @@ import type { Dictionary, Nullable } from '../types/typescript.js';
 import type { ExtraContextTypes } from './extra-context-types.js';
 
 export type RegisteredComponent = ComponentInfo & {
-  component: any;
+  component: /**
+   * Dynamically load components (currently only available for Svelte SDK)
+   */
+  | {
+        load: () => Promise<any>;
+        fallback?: any;
```
samijaber, line 14: can we add JSDoc comments to each one? 

```suggestion
        load: () => Promise<any>;
        fallback?: any;
```

samijaber, line 14: We can improve this:

- create a `ComponentReference` type, move it to its own file (`src/context/component-reference-types.ts`)
- add a svelte override for it


this will be great in the future to use each framework's component types: `React.Component` etc., instead of the current `any`

samijaber (approved):
Brilliant 🎉

packages/sdks/src/context/types.ts
sidmohanty11: added!

sidmohanty11: awesome idea! i have made the updates

PR #3615, opened 2024-10-08
Update package.json with zone.js version adhering to semver

jakehockey10:
## Description

The intent of this change is to "assume that only changes in the host package's major version will break your plugin."  This allows apps using the latest version of Angular with 0.14.10 version of zone.js to still "build."

I am not sure how to create a changeset, but I do believe this will require a new release.  Is that correct?  Currently, I cannot use @builder.io/angular in Angular projects using the latest version.


samijaber (approved):
Thank you for the contribution!

PR #3616, opened 2024-10-08
fix[vue-gen2]: isolated-vm changelog - remove plugin approach and add example for manual approach

sidmohanty11:
## Description

Removes the plugin from being exported from the SDK and the plugin approach mentioned in the CHANGELOG

_Screenshot_

### DIFF
--- a/.changeset/five-cows-bow.md
+++ b/.changeset/five-cows-bow.md
@@ -1,0 +1,5 @@
+---
+"@builder.io/sdk-vue": patch
+---
+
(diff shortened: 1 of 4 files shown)


samijaber (commented):
you'll wanna update the following line:

```
- added `initializeNodeRuntime` flag that executes `nuxt-initialize-node-runtime-plugin` plugin (defaults to `false`)
```

since there's no longer a plugin with that name

samijaber (approved):

PR #3622, opened 2024-10-08
Fix: SDKs onInit

samijaber:


packages/sdks/src/functions/evaluate/edge-runtime/edge-runtime.ts
```diff
@@ -1,32 +1,87 @@
 import { logger } from '../../../helpers/logger.js';
-import { set } from '../../set.js';
 import type { ExecutorArgs } from '../helpers.js';
-import { getFunctionArguments } from '../helpers.js';
+import { flattenState, getFunctionArguments } from '../helpers.js';
 import Interpreter from './acorn-interpreter.js';
 
+/**
+ * https://stackoverflow.com/a/46503625
+ */
```
samijaber, line 8: in node and edge runtimes, we use interpreter/VM to execute the arbitrary code in.

the main challenge with that is when customers are trying to set values in `state.*`:

```ts
state.foo = state.bar + 10;
state.baz = state.foo + 2
```

we need to:
- take those assignments and apply them on the SDK's global `state`
- apply the changes within the arbitrary code, because they might be needed right after (as in the example above)

for edge runtime, the old broken approach I had was to add a `global.setRootState` property and then add a call to that at the end of the `code`. However this only worked for very simple statements. I then found https://stackoverflow.com/questions/46496739/js-interpreter-changing-this-context/46503625#46503625 which offered a more elegant solution, which will work for long code blocks, by creating a "connected object" which will feed any of its changes back to the original object. So I changed it to that instead!

packages/sdks/src/functions/evaluate/edge-runtime/edge-runtime.ts
```diff
@@ -1,32 +1,87 @@
...
+      .replace(
+        /^.*?function main\(\)/,
+        `
+var __awaiter = function (e, t, n, r) {return r()},
+__generator = function (e, t) { return t() };
+function main()`
+      )
+      // replace ?. with .
+      .replace(/\?\./g, '.')
+  );
```
samijaber, line 82: Our `jsCode` blocks add a huge polyfill for `async/await` which reuses Promises. We need to remove that completely and replace it with identify functions because the js-interpreter does not support Promises

PR #3629, opened 2024-10-11
Fix: SDKs Eval

samijaber:


packages/sdks/src/functions/evaluate/evaluate.ts
```diff
@@ -54,12 +16,22 @@ export function evaluate({
   rootSetState,
   event,
   isExpression = true,
-  enableCache,
```
samijaber, line 57: not mention in changelogs that we removed all eval caching code, because we already disabled it in a previous PR a while back

packages/sdks/src/functions/evaluate/evaluate.ts
```diff
@@ -1,50 +1,12 @@
...
-  }
-
-  static setCachedValue(key: string, value: EvalValue) {
-    if (EvalCache.cache.size > 20) {
-      EvalCache.cache.delete(EvalCache.cache.keys().next().value);
-    }
-    EvalCache.cache.set(key, { value });
-  }
-}
+const STATE_GETTER_REGEX = /^(return )(\s*)?state(?<getPath>(\.\w+)+)(\s*);?$/;
```
sidmohanty11, line 9: can we add a small comment here? Does it do a multi-level search or just one level?

sidmohanty11 (approved):
🚀

steve8708:
does this also handle the fact that the bindings rarely actually look like `state.foo` and instead look like `var _virtual_index=state.foo;return _virtual_index`

looking at your PR it seems to assume the binding value is always in the unprocessed format (but could be wrong)

tl;dr in the app we compile TS to JS, and the way our rollup setup works `state.whatever` in JS becomes `var _virtual_index=state.whatever;return _virtual_index` so we'd ideally be able to match that and parse from there as well

packages/sdks/src/functions/evaluate/evaluate.ts
```diff
@@ -1,50 +1,17 @@
...
-    if (EvalCache.cache.size > 20) {
-      EvalCache.cache.delete(EvalCache.cache.keys().next().value);
-    }
-    EvalCache.cache.set(key, { value });
-  }
-}
+/** 
+ * handles multi-level gets on state: `state.x.y.z`
+ * does not handle bracket notation
+ * see regexr.com/87a9j
```
samijaber, line 12:
```suggestion
 * see https://regexr.com/87a9j
```

PR #3640, opened 2024-10-16
Fix: Svelte SDK: Awaiter

samijaber:
## Description

- add test for SSR'ing lazy-loadable components
- tweak Sveltekit e2e test to do the above


sidmohanty11 (approved):
🚀

PR #3644, opened 2024-10-17
warn users if they are on node 20 + Ubuntu

sidmohanty11:
## Description

Checks if the OS is Ubuntu running Node.js v20, then warns users that it might cause issues with the `isolated-vm` library.

To test `whichOS` function where OS is Ubuntu: https://onecompiler.com/nodejs/42vd2jj7s

**Jira**
https://builder-io.atlassian.net/browse/PLAN-303

_Screenshot_

### DIFF
--- a/packages/sdks/output/vue/nuxt.js
+++ b/packages/sdks/output/vue/nuxt.js
@@ -1,4 +1,4 @@
-import { addPlugin, defineNuxtModule } from '@nuxt/kit';
+import { addPlugin, defineNuxtModule, createResolver } from '@nuxt/kit';
 
 export default defineNuxtModule({
   setup(options, nuxt) {
@@ -30,8 +30,9 @@ export default defineNuxtModule({
         };
       }
 
+      const resolver = createResolver(import.meta.url);
       addPlugin({
-        src: './nuxt-isolated-vm-plugin.js',
+        src: resolver.resolve('nuxt-isolated-vm-plugin.js'),
         mode: 'server',
       });
     }
--- a/packages/sdks/src/functions/evaluate/node-runtime/init.ts
+++ b/packages/sdks/src/functions/evaluate/node-runtime/init.ts
@@ -1,3 +1,37 @@
+import fs from 'fs';
+import os from 'os';
+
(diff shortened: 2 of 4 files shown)


packages/sdks/src/functions/evaluate/node-runtime/init.ts
```diff
@@ -1,3 +1,37 @@
...
+      return 'Windows';
+    default:
+      return platform;
+  }
+};
+
+const osName = whichOS();
+if (osName.includes('Ubuntu') && isNode20) {
+  console.warn(
+    'WARNING: You are running Node 20 on Ubuntu. This combination is known to cause crashes with `isolated-vm`. See https://github.com/BuilderIO/builder/issues/3605 for more information.'
```
samijaber, line 31: instead of linking to github, lets add a new section for Node 20 + Ubuntu in https://www.builder.io/c/docs/integration-tips#enabling-data-bindings-in-node-environments and link to that

samijaber (commented):
@sidmohanty11 were you able to test this on an actual ubuntu machine? I want to make sure that the log does show up in the console and doesn't get swallowed (I've seen that happen under certain conditions, even though the problematic import is after the `console.warn`)

samijaber (commented):
also needs changesets

packages/sdks/src/functions/evaluate/node-runtime/init.ts
sidmohanty11: got it

packages/sdks/output/vue/nuxt.js
```diff
@@ -30,8 +30,9 @@ export default defineNuxtModule({
         };
       }
 
+      const resolver = createResolver(import.meta.url);
       addPlugin({
-        src: './nuxt-isolated-vm-plugin.js',
+        src: resolver.resolve('nuxt-isolated-vm-plugin.js'),
```
sidmohanty11, line 35: while testing a dev release, found out that it was failing to resolve. Now works

sidmohanty11:
Closing this as of now, will explore if we can dynamically import and catch errors to better determine the cause

PR #3650, opened 2024-10-18
fix tests

samijaber:

### DIFF
--- a/packages/sdks-tests/src/specs/index.ts
+++ b/packages/sdks-tests/src/specs/index.ts
@@ -156,7 +156,12 @@ const apiVersionPathToProp = {
 
 export type Path = keyof typeof PAGES;
 
-const GEN1_ONLY_PATHNAMES: Path[] = ['/api-version-v1', '/personalization-container'];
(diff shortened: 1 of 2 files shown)


packages/sdks-tests/src/specs/index.ts
```diff
@@ -156,7 +156,12 @@ const apiVersionPathToProp = {
 
 export type Path = keyof typeof PAGES;
 
-const GEN1_ONLY_PATHNAMES: Path[] = ['/api-version-v1', '/personalization-container'];
+const GEN1_ONLY_PATHNAMES: Path[] = [
+  '/api-version-v1',
+  '/personalization-container',
+  '/get-query',
+  '/get-content',
+];
```
samijaber, line 164: @clyde-builderio for future reference: paths only tested on gen1 need to be added to this array, otherwise they end up getting pre-rendered by certain e2e tests (React gen1 SDK + NextJS Pages router). And that can cause crashes if those content JSONs are testing something that doesn't work in gen2

PR #3656, opened 2024-10-22
fix[plugins][commercetools]: ENG-7186 Component that uses Commercetools plugin doesn't update Builder UI when components are selected successively

midhunadarvin:
## Description

- Update dependency from `@builder.io/commerce-plugin-tools` to `@builder.io/plugin-tools` as the former is deprecated.
- Add fix for `resourceId` not updating on props change
- Tested on CMS with local plugin URL.

**Link to JIRA ticket:**
https://builder-io.atlassian.net/browse/ENG-7186

**Screenshot**
Screenshot 2024-10-22 at 4 50 35 PM

**Loom**
https://www.loom.com/share/a9abb23801194f3c8ff967ea76af7b18


packages/plugin-tools/src/editors/ResourcesPicker.tsx
```diff
@@ -189,6 +189,13 @@ export const ResourcesPickerButton: React.FC<ResourcesPickerButtonProps> = props
       : typeof props.value === 'string'
       ? props.value
       : props.value?.options?.get(props.resourceName),
+    updateResourceId(newProps: ResourcesPickerButtonProps) {
```
midhunadarvin, line 192: When we are using `useLocalStore`, the values are set for the initial props, but doesn't update for any successive prop changes. Calling this function updates the `resourceId`.

packages/plugin-tools/src/editors/ResourcesPicker.tsx
```diff
@@ -189,6 +189,13 @@ export const ResourcesPickerButton: React.FC<ResourcesPickerButtonProps> = props
...
+      this.resourceId = newProps.handleOnly
+      ? undefined
+      : typeof newProps.value === 'string'
```
clyde-builderio, line 195: should it be `this.resourceId === newProps.handleOnly` ?

midhunadarvin: No, we are updating the `this.resourceId`, given here : https://github.com/BuilderIO/builder/pull/3656/files#diff-34006edc80e20c930eeb7596aa28b3830fc551637be592c16edb6a7e76dab2e7L187

sanyamkamat (approved):

PR #3659, opened 2024-10-22
Feature: SDK headers

samijaber:
## Description

https://builder-io.atlassian.net/browse/ENG-3160

- pending API changes to allow the new headers
- pending #3668


packages/core/src/builder.class.ts
```diff
@@ -2369,8 +2376,36 @@ export class Builder {
...
+        'X-Builder-SDK-GEN': '1';
+        'X-Builder-SDK-Version': string;
+      }
+    | {} {
+    if (!Builder.sdkInfo) {
+      return {};
+    }
+
+    return {
+      'X-Builder-SDK': Builder.sdkInfo.name,
```
shyam-builder, line 2391: Any ideas / suggestions on how we can track webcomponents and Angular here as well?

samijaber: good callout. we'll need to set `Builder.sdkInfo` there before the API calls, "on init" like we do in the gen1 react SDK: https://github.com/BuilderIO/builder/pull/3659/files#diff-809fb0cfc53140e43dabe33aad3b09a2cfe973d83c4f3614fd58d5fe9c9ff8afR8-R11

samijaber: But we'll also need to improve those SDKs to start storing the SDK version in a constant which requires an extra build step

shyam-builder (approved):

PR #3660, opened 2024-10-22
chore:Changing email from help to support

nando-builder:
## Description

Changing the email from `help@builder.io` to `support@builder.io` as the email help@builder.io is not under monitoring.
This email is used in some error messages, for example:

```
No express request set! Builder cannot target appropriately without this, please contact [help@builder.io](mailto:help@builder.io) to learn how to set this as required
```

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,4 +2,4 @@
 
 ## Reporting a Vulnerability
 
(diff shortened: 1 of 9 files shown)


samijaber (approved):
Most of these will require us to publish the new versions of the plugins for the update to be reflected.

we can wait until the next time there's another code update being made to include this change

PR #3662, opened 2024-10-23
EDU-248: Angular Gen 2 None -  Custom Child component

uttej-vsk:
## Description

This PR introduces the Angular Gen 2 version of the Custom Child snippet within the custom components, which closely resembles the previously merged [custom child in the React framework](https://github.com/BuilderIO/builder/pull/3534/files#diff-ad9391cc4a5c924c5b6eb21f4dcdb11adeae273688919fe47e5bfbf71a1f7c71).

@samijaber Please review :)

### DIFF
--- a/packages/sdks/snippets/angular/src/app/custom-child/custom-child.component.ts
+++ b/packages/sdks/snippets/angular/src/app/custom-child/custom-child.component.ts
@@ -1,0 +1,50 @@
+import { CommonModule } from '@angular/common';
+import { Component, type OnInit } from '@angular/core';
+import {
+  Content,
+  fetchOneEntry,
+  isPreviewing,
+  type BuilderContent,
+} from '@builder.io/sdk-angular';
+import { customHeroInfo } from './custom-hero/custom-hero.component';
+
+@Component({
+  selector: 'app-custom-child',
+  standalone: true,
+  imports: [Content, CommonModule],
+  template: `
+    <div *ngIf="content">
+      <builder-content
+        [content]="content"
+        [model]="model"
+        [apiKey]="apiKey"
+        [customComponents]="[customHeroInfo]"
+      ></builder-content>
+    </div>
+
+    <div *ngIf="notFound">404 Not Found</div>
+  `,
+})
+export class CustomChildComponent implements OnInit {
+  notFound = false;
+  content: BuilderContent | null = null;
+  customHeroInfo = customHeroInfo;
+  model = 'page';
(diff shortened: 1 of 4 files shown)


packages/sdks/snippets/angular/src/app/custom-child/custom-child.component.ts
```diff
@@ -0,0 +1,50 @@
...
+  `,
+})
+export class CustomChildComponent implements OnInit {
+  notFound = false;
+  content: BuilderContent | null = null;
+  customHeroInfo = customHeroInfo;
+  model = 'page';
+  apiKey = 'ee9f13b4981e489a9a1209887695ef2b';
+
+  isPreviewing = isPreviewing;
```
samijaber, line 35: what is this for? doesn't seem to be used anywhere?

packages/sdks/snippets/angular/src/app/custom-child/custom-child.component.ts
```diff
@@ -0,0 +1,50 @@
...
+      model: this.model,
+      apiKey: this.apiKey,
+      userAttributes: {
+        urlPath: window.location.pathname,
+      },
+    });
+
+    if (!this.content) {
+      this.notFound = true;
+    }
```
samijaber, line 48: this seems overly complex. instead of defining and setting `this.notFound`, can't we use the `ngIfElse` syntax, or:

```html
<div *ngIf="!content">404 Not Found</div>
```

uttej-vsk: My initial approach was to keep the 404 in the _else_ block. But I felt it was giving a bad developer experience when it displayed the "404 Not Found" block momentarily when I refreshed. So I figured it would have to wait for _fetchOneEntry_ to have something in the _content_

So, to remove that, I added another guard. But please let me know if we can remove this.

uttej-vsk:
```
 if (!this.content) {
      this.notFound = true;
    }
```

This code ensures this won't happen like that. This loom shows this behaviour without that check. 
https://www.loom.com/share/290e4998fbcf48d3bf2fff1eab01bc2b

samijaber: good point, lets keep it as is

samijaber (approved):

PR #3664, opened 2024-10-23
[ENG-6918] Update Magento dependencies to the latest plugin-tools

shyam-builder:
## Description

Update magento2 plugin tools to the latest `plugin-tools` dependency from the older one which is deprecated.

In addition, cleanup dependencies to use the same set from one of the newer plugins (`salesforce-commerce-api-plugin`) and simplify the `package.json`


plugins/magento2/package.json
```diff
@@ -85,42 +33,25 @@
...
     "rollup-plugin-serve": "^1.0.1",
-    "semantic-release": "^21.0.1",
-    "shelljs": "^0.8.3",
-    "travis-deploy-once": "^5.0.9",
-    "ts-jest": "^23.10.2",
-    "ts-node": "^7.0.1",
-    "tslint": "^5.11.0",
-    "tslint-config-prettier": "^1.15.0",
-    "tslint-config-standard": "^8.0.1",
-    "typedoc": "^0.12.0",
```
samijaber, line 120: amazing. removing all of these for all plugins will drastically reduce dependabot alerts eventually 🙏🏽

PR #3669, opened 2024-10-24
feat[gen2][apiBaseUrl] ENG-7147 allow user to pass custom apiHost for Gen2 SDKs

clyde-builderio:
## Description

Allow user to set baseUrl by making it a writable field of builder.class.ts

**Link to JIRA ticket:**
https://builder-io.atlassian.net/browse/ENG-7147


packages/sdks/src/functions/track/index.ts
```diff
@@ -65,6 +65,13 @@ const getTrackingEventData = async ({
   };
 };
 
+export type CustomApiBaseUrl = {
+  /**
+   * Optional override of the `baseUrl` of the Builder API. (Defaults to global `https://cdn.builder.io`)
+   */
+  apiBaseUrl?: string;
```
samijaber, line 72: I would rename this to `apiHost` and the description like this:

```suggestion
  /**
   * Sets the host of Builder API calls. (Defaults to global `https://cdn.builder.io`)
   */
  apiHost?: string;
```

packages/sdks/src/components/content-variants/content-variants.lite.tsx
```diff
@@ -37,7 +38,7 @@ type VariantsProviderProps = ContentVariantsPrps & {
    * For internal use only. Do not provide this prop.
    */
   isNestedRender?: boolean;
-};
+} & CustomApiBaseUrl;
```
samijaber, line 41: to keep the types of `Content` consistent, you should add `apiBaseUrl` here: 

https://github.com/BuilderIO/builder/blob/main/packages/sdks/src/components/content-variants/content-variants.types.ts#L35


then you shouldn't need to add `& CustomApiBaseUrl` everywhere

packages/sdks/src/functions/track/index.ts
```diff
@@ -118,7 +123,9 @@ export async function _track(eventProps: EventProps) {
     return;
   }
 
-  return fetch(`https://cdn.builder.io/api/v1/track`, {
+  const baseUrl = eventProps.apiHost || 'https://cdn.builder.io';
```
samijaber, line 126: You have to make sure that `apiHost` isn't included in the call to `createEvent(eventProps)`, or we will be sending the `apiHost` property with every tracking event.

packages/sdks/src/components/content-variants/content-variants.lite.tsx
```diff
@@ -37,6 +37,11 @@ type VariantsProviderProps = ContentVariantsPrps & {
    * For internal use only. Do not provide this prop.
    */
   isNestedRender?: boolean;
+
+  /**
+   * Sets the host of Builder API calls. (Defaults to global `https://cdn.builder.io`)
+   */
+  apiHost?: string;
```
samijaber, line 44: isn't this a duplicate of the type in `packages/sdks/src/components/content-variants/content-variants.types.ts`? why do we need this one?

packages/sdks-tests/src/e2e-tests/hit-overridden-base-url.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { expect } from '@playwright/test';
+import { excludeGen1, test } from '../helpers/index.js';
+
+test.describe('Hit Overridden Base Url', () => {
+  test('call custom API only once - in page', async ({ page, sdk }) => {
```
samijaber, line 5: 👍🏽

samijaber (commented):
don't forget the changset

packages/sdks/src/components/content-variants/content-variants.lite.tsx
clyde-builderio: Yes, good catch! It is not required. Deleting this.

packages/sdks/src/functions/track/index.ts
```diff
@@ -118,7 +123,10 @@ export async function _track(eventProps: EventProps) {
...
+
+  delete eventProps.apiHost;
```
samijaber, line 128: this approach is dangerous. it might delete `apiHost` from the parent function, which means subsequent uses of `apiHost` won't have access to the variable.

Here is an example:
CleanShot 2024-10-28 at 11 49 30@2x


so `parentFunction` now loses its `props.foo`.

Instead, what you should do is change:

```ts
export async function _track(eventProps: EventProps) {
```

to:

```ts
export async function _track(({ apiHost, ...eventProps }): EventProps) {
```

this does not involve deleting anything

samijaber: you still didn't get rid of the `delete` here. it's redundant and can be removed

clyde-builderio: My bad, pushed the change

samijaber (approved):

PR #3678, opened 2024-10-24
plugin: trieve conversational discovery

drew-harris:
## Description

Adds a new plugin that allows models to sync their content with [Trieve](https://trieve.ai/). 
Trieve is an alternative to Algolia which offers conversational discovery and semantic recommendations.

Trieve has webhook support for [builder.io](http://builder.io/) data syncing, which you can find the original PR for [here](https://github.com/devflowinc/trieve/pull/2627)

Open to feedback especially in regards to further publishing steps.

### DIFF
--- a/plugins/trieve/.babelrc
+++ b/plugins/trieve/.babelrc
@@ -1,0 +1,17 @@
+{
+  "presets": [
+    "@babel/preset-react"
+  ],
+  "env": {
+    "development": {
+      "presets": [
+        [
+          "@babel/preset-react",
+          {
+            "development": true
+          }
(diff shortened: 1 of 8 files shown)


samijaber:
Thanks for making this plugin contribution to Builder!

We are revamping our submission process for community plugins to maintain our ecosystem better. As part of that change, we now ask contributors to create their own repos and self-publish their plugins, and then make a submission using the following form: https://forms.gle/hq9ybaBDAdB4dvzy6

I am going to go ahead and close this PR as it is no longer necessary.

PR #3681, opened 2024-10-25
feat[gen1][core] ENG-7238 remove query API and apiEndpoint field

clyde-builderio:
## Description
Remove query API for Gen1 SDK

Objectives:
1. Remove query API from Gen1 SDK.
2. Remove `apiEndpoint` and the code that depends on this field as an input.
3. Remove unwanted tests and modify existing tests.

**Link to JIRA ticket:**
https://builder-io.atlassian.net/browse/ENG-7238


packages/core/src/builder.class.ts
```diff
@@ -2549,7 +2541,7 @@ export class Builder {
       for (const key of properties) {
         const value = options[key];
         if (value !== undefined) {
-          if (apiEndpoint === 'query') {
+          if (format === 'solid' || format === 'react') {
```
samijaber, line 2544: if we move `isApiCallForCodegen` above this line we can reuse it here
```suggestion
          if (isApiCallForCodegen) {
```

.changeset/pink-fans-serve.md
```diff
@@ -0,0 +1,6 @@
+---
+"@builder.io/sdk": major
+"@builder.io/react": major
+---
+
+Permanently removes `apiEndpoint` prop from `builder.get()` and `builder.getAll()` which would have options `'content'` and `'query'`. Content API will be the default API endpoint is used for the content fetching.
```
samijaber, line 6:
```suggestion
Permanently removes the `apiEndpoint` prop from `builder.get()` and `builder.getAll()` which had options `'content'` and `'query'`. Content API is now the only possible API endpoint for content fetching.
```

samijaber (approved):

PR #3687, opened 2024-10-26
EDU-292: Angular Gen 2 None - Editable Regions

uttej-vsk:
## Description

This PR provides our documentation with Angular Gen 2 None code snippet for editable regions in custom components.

@samijaber Please review

### DIFF
--- a/packages/sdks/snippets/angular/src/app/app.module.ts
+++ b/packages/sdks/snippets/angular/src/app/app.module.ts
@@ -12,10 +12,11 @@ import { BlogArticleComponent } from './blog-article/blog-article.component';
 import { CatchAllComponent } from './catch-all/catch-all.component';
 import { CustomChildComponent } from './custom-child/custom-child.component';
 import { CustomHeroComponent } from './custom-child/custom-hero/custom-hero.component';
+import { CustomColumnsComponent } from './editable-regions/custom-columns/custom-columns.component';
+import EditableRegionComponent from './editable-regions/editable-regions.component';
 import { NavBarComponent } from './nav-bar/nav-bar.component';
 import { ProductDetailsComponent } from './product-details/product-details.component';
 import { ProductEditorialComponent } from './product-editorial/product-editorial.component';
-
 @NgModule({
   declarations: [AppComponent],
   imports: [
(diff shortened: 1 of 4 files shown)


packages/sdks/snippets/angular/src/app/editable-regions/editable-regions.component.ts
```diff
@@ -0,0 +1,46 @@
...
+        [content]="content"
+        [model]="model"
+        [apiKey]="apiKey"
+        [customComponents]="customComponents"
+      />
+    </div>
+    <div *ngIf="notFound">404</div>
+  `,
+})
+export default class EditableRegionComponent implements OnInit {
```
sidmohanty11, line 29: we can remove `implements OnInit`

packages/sdks/snippets/angular/src/app/app.module.ts
```diff
@@ -12,10 +12,11 @@ import { BlogArticleComponent } from './blog-article/blog-article.component';
 import { CatchAllComponent } from './catch-all/catch-all.component';
 import { CustomChildComponent } from './custom-child/custom-child.component';
 import { CustomHeroComponent } from './custom-child/custom-hero/custom-hero.component';
+import { CustomColumnsComponent } from './editable-regions/custom-columns/custom-columns.component';
+import EditableRegionComponent from './editable-regions/editable-regions.component';
```
sidmohanty11, line 16: can we use named export for `EditableRegionComponent` to maintain consistency?

packages/sdks/snippets/angular/src/app/editable-regions/editable-regions.component.ts
uttej-vsk: Ah! Though it works fine, after reading the Angular docs, one issue in [github](https://github.com/angular/angular/issues/5814) and other posts in Stackoverflow, it is more of the best practices. TS will help us with typos during compile time. 

If you also see the below images, if I implement **_onInit_**, when I mistype **_ngOnInit_**(), it throws an error. If we don't implement it, Angular compiles fine and gives no error, and then we will have difficulty figuring it out. 

Screenshot 2024-10-26 at 9 55 48 AM
Screenshot 2024-10-26 at 9 56 02 AM

That is the reason why it is there in the code. Let me know if it makes sense :)

packages/sdks/snippets/angular/src/app/app.module.ts
uttej-vsk: This is an overlook. Made the change and pushed the code.

packages/sdks/snippets/angular/src/app/editable-regions/editable-regions.component.ts
sidmohanty11: sure, do you think we should also add this to other components to maintain consistency?

sidmohanty11 (approved):

PR #3688, opened 2024-10-26
feat: support indexName

manucorporat:


steve8708:
fyi this only adds support for gen2 SDKs which are comparatively low usage today compared to gen1

packages/sdks/src/components/block/block.helpers.ts
```diff
@@ -77,7 +77,7 @@ export const getRepeatItemData = ({
         $index: index,
         $item: item,
         [itemNameToUse]: item,
-        [`$${itemNameToUse}Index`]: index,
```
steve8708, line 80: this looks like a breaking change, no?
should we not support both so we don't break existing content?

PR #3690, opened 2024-10-28
fix[gen2-sdks]: nested ContentVariants with same model name should not update content

sidmohanty11:
## Description

https://builder-io.atlassian.net/browse/ENG-7210

_Screenshot_


packages/sdks/src/components/content/components/enable-editor.lite.tsx
```diff
@@ -326,7 +325,7 @@ export default function EnableEditor(props: BuilderEditorProps) {
 
   onMount(() => {
     if (isBrowser()) {
-      if (isEditing()) {
+      if (isEditing() && !props.isNestedRender) {
```
samijaber, line 328: good find!

samijaber (approved):
🐛 💥

PR #3692, opened 2024-10-28
Export text evaluate function

phelma:
## Description

We're overwriting the default Text component in order to add some functionality.

However by doing this we've lost the ability to use [Custom data and Context](https://www.builder.io/c/docs/custom-actions) in our text component ( inserting values into text fields like this: {{state.value}}.

By exporting the evaluate function we will be able to add this functionality into our custom text component.

### DIFF
--- a/.changeset/new-lemons-visit.md
+++ b/.changeset/new-lemons-visit.md
@@ -1,0 +1,14 @@
+---
+'@builder.io/sdk': patch
+'@builder.io/react': patch
+'@builder.io/sdk-angular': patch
+'@builder.io/sdk-react-nextjs': patch
+'@builder.io/sdk-qwik': patch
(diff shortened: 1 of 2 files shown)


samijaber:
@phelma thanks for the PR!

While this would get the job done, I am hesitant to expose our `evaluate` logic as part of the SDK's API due to its inherent complexity and edge cases.

What we can do instead to unblock you is move the `evaluate()` logic that's inside of `text.lite.tsx` to run as part of the core SDK logic, so that anyone overriding Text will receive a pre-evaluated `text` property.

phelma:
@samijaber yes that would be great.
we've unblocked ourselves by copy/pasting the code into our project, but you're proposal would be a better solution

PR #3702, opened 2024-10-31
docs(nuxt): use proper key to useAsyncData

kerwanp:
## Description

After struggling to understand why navigation was not working properly on prerendered routes I finally understood what was the culprit.

Nuxt uses the key of `useAsyncData` to cache that value on the server and the client. Prerendered routes will have their async data stored inside a payload file that is shared to the client when navigating.

In the current documentation it uses a static key for all pages content:
```ts
const { data: content } = await useAsyncData('builderData', () =>
  fetchOneEntry({
    model,
    apiKey,
    userAttributes: {
      urlPath: route.path,
    },
  })
))
```

When routes are prerender, navigation will fetch the payload of the page and replace async data on the client based on the key. If the key already exist, it is not replaced. So in this case, navigating between pages will always result with the content of the first visited page.

To avoid this issue, each page should have its own key:

```ts
const { data: content } = await useAsyncData(`builderData-page-${route.path}`, () =>
  fetchOneEntry({
    model,
    apiKey,
    userAttributes: {
      urlPath: route.path,
    },
  })
))
```

This PR update the documentation accordingly

### DIFF
--- a/examples/vue/nuxt-3-catchall/pages/[...app].vue
+++ b/examples/vue/nuxt-3-catchall/pages/[...app].vue
@@ -45,7 +45,7 @@ const BUILDER_PUBLIC_API_KEY = 'f1a790f8c3204b3b8c5c1795aeac4660'; // ggignore
 const route = useRoute();
 
 // fetch builder content data
-const { data: content } = await useAsyncData('builderData', () =>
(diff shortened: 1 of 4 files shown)


samijaber (approved):
Awesome, thanks a lot for this contribution!

PR #3716, opened 2024-11-04
Misc/refactor SDK tests

samijaber:
## Description

cleanup the logic around our SDK tests. we had tons of arrays modifying test behaviour, now its all consolidated in the same object with the content JSON

### DIFF
--- a/packages/sdks-tests/src/specs/index.ts
+++ b/packages/sdks-tests/src/specs/index.ts
@@ -181,13 +181,7 @@ export type Path = keyof typeof PAGES;
 
 export const getAllPathnames = (target: 'gen1' | 'gen2'): string[] => {
   return Object.entries(PAGES)
-    .filter(([_, page]) => {
-      if (target === 'gen1') {
-        return page.target === 'gen1' || page.target === 'all';
-      } else {
-        return page.target === 'gen2' || page.target === 'all';
-      }
-    })
+    .filter(([_, page]) => page.target === target || page.target === 'all')
     .map(([pathname]) => pathname);
 };
 
--- a/packages/sdks-tests/src/specs/index.ts
+++ b/packages/sdks-tests/src/specs/index.ts
@@ -90,12 +90,12 @@ type Page = {
    * network request instead.
    * NOTE: This is why I can only test visual editing for the old SDK when using a non-SSR app
    */
-  isVisualEditingTest?: boolean;
+  isGen1VisualEditingTest?: boolean;
 };
 
 export const PAGES: Record<string, Page> = {
   '/': { content: HOMEPAGE },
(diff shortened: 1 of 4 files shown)


packages/sdks-tests/src/specs/index.ts
```diff
@@ -74,110 +74,119 @@ function isBrowser(): boolean {
...
+  '/custom-components-nowrap': { content: CUSTOM_COMPONENTS_NOWRAP },
+  '/override-base-url': { content: HTTP_REQUESTS },
 } as const;
 
 export type Path = keyof typeof PAGES;
 
-const GEN1_ONLY_PATHNAMES: Path[] = ['/api-version-v1', '/personalization-container'];
-const GEN2_ONLY_PATHNAMES: Path[] = [];
-
 export const getAllPathnames = (target: 'gen1' | 'gen2'): string[] => {
```
sidmohanty11, line 182: target can be of type `Page['target']` to include `'all'` what do you think?

packages/sdks-tests/src/specs/index.ts
```diff
@@ -74,110 +74,119 @@ function isBrowser(): boolean {
...
+   * Defaults to `all`.
+   */
+  target?: 'gen1' | 'gen2' | 'all';
+  /**
+   * To test visual editing in Gen 1 SDK, we cannot provide a hardcoded JSON.
+   * Instead, we have to let the SDK fetch the data from the API and mock the
+   * network request instead.
+   * NOTE: This is why I can only test visual editing for the old SDK when using a non-SSR app
+   */
+  isVisualEditingTest?: boolean;
```
sidmohanty11, line 93: @samijaber can we name this specifically targeted to gen1, something like `isGen1VisualEditingTest` or anything similar? As there are other visual editor tests for columns or symbols where we only have them for gen2

samijaber: mm sure. we can rename it to `isVisualEditingTest` once all the visual editing tests have been updated to run on gen1

packages/sdks-tests/src/specs/index.ts
samijaber: `getAllPathnames` is meant to be called within a specific e2e test app, to retrieve the pathnames that it needs to pre-render. so it doesn't make sense to allows `'all'` as an option

sidmohanty11: ohh right, makes sense

packages/sdks-tests/src/specs/index.ts
sidmohanty11: yes! thank you

sidmohanty11 (approved):

PR #3717, opened 2024-11-05
fix[gen2-sdks]: accordion visual editing and order

sidmohanty11:
## Description

for visual editing, we needed to pass the correct path. In gen1, if `component.options` is not already prefixed we add them to the path: https://github.com/sidmohanty11/builder/blob/fix/accordion-gen2/packages/react/src/components/builder-blocks.component.tsx#L48 so passing just `item.$index.detail` didn't work

for misordering of item details, we should only use `order` when `props.grid === true`: https://github.com/sidmohanty11/builder/blob/fix/accordion-gen2/packages/widgets/src/components/Accordion.tsx/#L75-L80

Jira
https://builder-io.atlassian.net/browse/ENG-7330

_Screenshot_


samijaber (approved):

packages/sdks/src/blocks/accordion/accordion.lite.tsx
```diff
@@ -177,23 +177,25 @@ export default function Accordion(props: AccordionProps) {
...
+                  }) as any),
+                }),
               }}
               data-index={index}
               onClick={() => state.onClick(index)}
             >
               <Blocks
                 blocks={item.title}
-                path={`items.${index}.title`}
+                path={`component.options.items.${index}.title`}
```
samijaber, line 198: actually can we improve `Blocks` to also add the prefix if its not there? then we can simplify code for end users building with `Blocks`

sidmohanty11: yup sure!

oxcabe:
Hi team, thanks for addressing my issue so quickly.

I can now use the Accordion as intended (yay!), but am still getting errors related to it. From my Chrome DevTools console:

```
Warning: Each child in a list should have a unique "key" prop.
```

I haven't used mitosis, so I'm unsure if the `<For>` element manages the `key` attribute somehow. If this is not the case, then I'd say this error would be solved by adding a key to the fragment on line 175:
```diff
diff --git a/packages/sdks/src/blocks/accordion/accordion.lite.tsx b/packages/sdks/src/blocks/accordion/accordion.lite.tsx
index abcdefg..hijklmn 100644
--- a/packages/sdks/src/blocks/accordion/accordion.lite.tsx
+++ b/packages/sdks/src/blocks/accordion/accordion.lite.tsx
@@ -1,6 +1,7 @@
 import {
   For,
   Show,
+  Fragment,
   useMetadata,
   useStore,
   useTarget,
@@ -173,11 +174,11 @@ export default function Accordion(props: AccordionProps) {
   return (
     <div class="builder-accordion" style={state.accordionStyles}>
       <For each={props.items}>
-        {(item, index) => (
-          <>
+        {(item, index) => (
+          <Fragment key={index}>
             <div
               class={state.getAccordionTitleClassName(index)}
               style={{
@@ -193,11 +194,11 @@ export default function Accordion(props: AccordionProps) {
                 />
               </div>
               <Show when={state.open.includes(index)}>
                 <div
                   class={state.getAccordionDetailClassName(index)}
                   style={state.accordionDetailStyles}
                 >
                   <Blocks
                     blocks={item.detail}
                     path={`items.${index}.detail`}
                     parent={props.builderBlock.id}
@@ -205,7 +206,7 @@ export default function Accordion(props: AccordionProps) {
                   />
                 </div>
               </Show>
-          </>
+          </Fragment>
         )}
       </For>
     </div>
   );
 }
 ```

I'd be more than happy to open a PR tomorrow if you're okay with that. Otherwise, feel free to just patch the diff. Again, thanks for the quick resolution.

sidmohanty11:
@oxcabe thanks for the prompt feedback! I must have missed this. The approach you've suggested should fix this issue, feel free to create the PR

PR #3724, opened 2024-11-07
fix[gen2-nextjs]: import component info objects from client modules

sidmohanty11:
## Description

Jira
https://builder-io.atlassian.net/browse/ENG-7324

_Screenshot_


packages/sdks/src/components/content/content.lite.tsx
```diff
@@ -56,11 +56,7 @@ export default function ContentComponent(props: ContentProps) {
 
     registeredComponents: [
       ...getDefaultRegisteredComponents(),
-      ...(props.customComponents?.filter(({ models }) => {
-        if (!models?.length) return true;
-        if (!props.model) return true;
-        return models.includes(props.model);
-      }) || []),
```
samijaber, line 63: I see that you permanently got rid of this filtering for `registeredComponents`.

For the sake of consistency, let's keep this logic and move it to:
 https://github.com/BuilderIO/builder/blob/25f00aca07a395df8a5b663ffc963d0e8c808570/packages/sdks/src/components/block/block.lite.tsx#L118-L123

To make sure that the registered components and the rendered components perfectly match.

samijaber (commented):
Needs a changeset!

packages/sdks/src/components/content/content.lite.tsx
sidmohanty11: sure @samijaber makes sense, let me do that

sidmohanty11: @samijaber added the restriction part, do you think we should also filter out registered components that we expose to the users when they do `shouldReceiveBuilderProps: { builderComponents: true }` here?
https://github.com/BuilderIO/builder/blob/8a209a99793802ae09aeec0664455caaadb7fded/packages/sdks/src/components/block/block.lite.tsx#L191-L194

samijaber: Ah yes, good catch. Let's do that for consistency as well

sidmohanty11: done!

packages/sdks/src/components/block/block.helpers.ts
```diff
@@ -25,7 +27,13 @@ export const getComponent = ({
 
   const ref = registeredComponents[componentName];
 
-  if (!ref) {
+  const isRestricted =
+    model &&
+    ref?.models &&
+    ref.models?.length > 0 &&
+    !ref.models.includes(model);
```
samijaber, line 34: can we move this to a `checkIsRestrictedModel` that is reused across this PR?

samijaber (approved):
👍🏽 

but the new tests I added are failing on angular/nuxt. Is it related to this PR, or unrelated flakiness?

packages/sdks/src/components/block/block.helpers.ts
sidmohanty11: nice idea, done

sidmohanty11:
looks like due to unrelated flakiness @samijaber

PR #3725, opened 2024-11-07
fix[gen2-sdks]: add keys for Fragment inside Accordion JSX loop

oxcabe:
## Description

This PR adds the `key` attribute inside the Accordion component. This attribute is required by React's use of loops in JSX code, and will throw an error when the top-level children in a loop are missing this attribute.

From the [TypeScript handbook on `JSX.IntrinsicAttributes`](https://www.typescriptlang.org/docs/handbook/jsx.html#attribute-type-checking:~:text=Additionally%2C%20the%20JSX,on%20every%20tag.):

> Additionally, the JSX.IntrinsicAttributes interface can be used to specify extra properties used by the JSX framework which are not generally used by the components’ props or arguments - for instance key in React. Specializing further, the generic JSX.IntrinsicClassAttributes<T> type may also be used to specify the same kind of extra attributes just for class components (and not Function Components). In this type, the generic parameter corresponds to the class instance type. In React, this is used to allow the ref attribute of type Ref<T>. Generally speaking, all of the properties on these interfaces should be optional, unless you intend that users of your JSX framework need to provide some attribute on every tag.

## Additional concerns

Although this is more a situation exclusively about Mitosis, I wanted to, at least, mention it here.

I'd argue that the `key` attribute, being intrinsic to React's use of JSX, should be managed accordingly inside Mitosis heuristics in order to avoid polluting the Mitosis JSX markup with references to specific transpiler back-ends. Moreover, how could we do the same we're doing in this commit if we were to use Sveltosis instead?

For this reason, I want to propose considering this a temporary fix and solve the problem in the right context, which would be Mitosis handling of key attributes in loops.


.changeset/three-shrimps-eat.md
```diff
@@ -0,0 +1,11 @@
...
+"@builder.io/sdk-angular": patch
+"@builder.io/sdk-qwik": patch
+"@builder.io/sdk-react": patch
+"@builder.io/sdk-react-native": patch
+"@builder.io/sdk-solid": patch
+"@builder.io/sdk-svelte": patch
+"@builder.io/sdk-vue": patch
+---
+
+Fix: add keys for Fragment inside Accordion JSX loop
```
samijaber, line 11:
```suggestion
Fix: add `key` prop for loop inside Accordion Block,
```

samijaber (approved):
Thanks a lot for this helpful contribution 🙏🏽

samijaber:
@oxcabe Github is acting up and not running the CI tests on your PR. I've re-raised the PR over at https://github.com/BuilderIO/builder/pull/3813 to work around this problem.

PR #3727, opened 2024-11-08
Gen1 SDKs: Fix: XSS exploit

samijaber:
## Description

- disable editing unless within an iframe


packages/core/src/builder.class.ts
```diff
@@ -1663,9 +1663,9 @@ export class Builder {
       this.authToken = authToken;
     }
     if (isBrowser) {
-      this.bindMessageListeners();
-
       if (Builder.isEditing) {
+        this.bindMessageListeners();
```
samijaber, line 1667: moving this call inside of `Builder.isEditing` is the fundamental fix in this PR: before that, we were always listening to `builder.*` messages like `builder.evaluate`.

packages/react/src/components/builder-component.component.tsx
```diff
@@ -510,6 +510,9 @@ export class BuilderComponent extends React.Component<
   }
 
   messageListener = (event: MessageEvent) => {
+    const isTrusted = Builder.isTrustedHostForEvent(event);
+    if (!isTrusted) return;
```
samijaber, line 514: added missing `isTrusted` check here

PR #3729, opened 2024-11-12
fix(sdks): serialize single arg functions with no parens correctly

teleaziz:
a single argument arrow function without parens would cause this invalid js syntax
```js
return (function e=>!0===e.get(\"isABTest\")).apply(this, arguments)
```

### DIFF
--- a/packages/sdks/src/functions/__snapshots__/register-component.test.ts.snap
+++ b/packages/sdks/src/functions/__snapshots__/register-component.test.ts.snap
@@ -42,6 +42,7 @@ exports[`Component Registration and Serialization > serializeFn handles differen
     "func3": "return (function func3(x) {
           return x - 1;
         }).apply(this, arguments)",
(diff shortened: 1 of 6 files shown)


packages/sdks/src/functions/__snapshots__/register-component.test.ts.snap
```diff
@@ -42,6 +42,7 @@ exports[`Component Registration and Serialization > serializeFn handles differen
     "func3": "return (function func3(x) {
           return x - 1;
         }).apply(this, arguments)",
+    "func4": "return (e => !0 === e.get(\\"isABTest\\")).apply(this, arguments)",
```
teleaziz, line 45: it was appending `function` here, causing an invalid js syntax
Screenshot 2024-11-12 at 12 31 23 PM

samijaber (approved):

PR #3734, opened 2024-11-14
fix[gen2][fetchEntries] ENG-7327 ensure fetchEntries function throws an error when API fails

clyde-builderio:
## Description

`fetchEntries` does not throw error

**Jira Ticket**
https://builder-io.atlassian.net/browse/ENG-7327

### DIFF
--- a/packages/sdks/src/functions/get-content/index.ts
+++ b/packages/sdks/src/functions/get-content/index.ts
@@ -104,18 +104,13 @@ export const _processContentResult = async (
  * Returns a paginated array of entries that match the given options.
  */
 export async function fetchEntries(options: GetContentOptions) {
-  try {
-    const url = generateContentUrl(options);
-    const content = await _fetchContent(options);
+  const url = generateContentUrl(options);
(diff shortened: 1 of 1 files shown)


.changeset/odd-pandas-look.md
```diff
@@ -0,0 +1,12 @@
...
+'@builder.io/sdk-react-nextjs': major
+'@builder.io/sdk-qwik': major
+'@builder.io/sdk-react': major
+'@builder.io/sdk-react-native': major
+'@builder.io/sdk-solid': major
+'@builder.io/sdk-svelte': major
+'@builder.io/sdk-vue': major
+---
+
+Throw errors when API fails for `fetchEntries`
```
samijaber, line 12: Breaking Change 🧨: `fetchEntries` and `fetchOneEntry` calls will now throw any errors thrown by `fetch`, or any non-success response returned from the Builder API.

Previously, both functions would swallow all errors and return `null`.

samijaber (approved):

PR #3736, opened 2024-11-15
feat(react): allow editing symbols in the context of a parent page pr…

teleaziz:
…eview url

Allow editing symbols when the preview URL is the parent page.

Makes it easier to tweak symbols by letting you edit them directly where they're being used, rather than having to find and open the symbol file separately. Way more convenient for users

### DIFF
--- a/.changeset/late-pugs-do.md
+++ b/.changeset/late-pugs-do.md
@@ -1,0 +1,5 @@
+---
+"@builder.io/react": patch
+---
+
+feat: allow symbols to be edited in the context of a parent entry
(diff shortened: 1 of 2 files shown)


samijaber (commented):
@teleaziz I don't fully follow what this enables. can you expand on what you mean by "allows editing of symbols in the context of a parent page"? a loom showing how this works would be great

teleaziz:
@samijaber Thanks, but I want to keep this PR focused on fixing the React SDK user-blocking issue. Let's track gen2 SDK improvements separately - I'm not sure Gen2 has the same issue here since the implementations has a different approach.

samijaber (approved):

PR #3737, opened 2024-11-15
SDKs: Add test for Symbol JsCode state updates

samijaber:
## Description

Add test to confirm that Symbols `jsCode` correctly updates global state. Currently fails for Qwik, React and NextJS gen2 SDKs.

### DIFF
--- a/packages/sdks-tests/src/specs/index.ts
+++ b/packages/sdks-tests/src/specs/index.ts
@@ -73,7 +73,7 @@ import { BLOCKS_CLASS_NAME } from './blocks-class-name.js';
 import { DUPLICATED_CONTENT_USING_NESTED_SYMBOLS } from './duplicated-content-using-nested-symbols.js';
 import { CUSTOM_COMPONENTS_NOWRAP } from './custom-components-nowrap.js';
 import { XSS_EXPLOIT } from './xss-exploit.js';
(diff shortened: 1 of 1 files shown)


packages/sdks-tests/src/e2e-tests/countdown.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { expect } from '@playwright/test';
+import { test } from '../helpers/index.js';
+
+test.describe('Symbol with JS Code', () => {
+  test('correctly updates countdown date', async ({ page, sdk }) => {
+    test.fail(
+      sdk === 'qwik' || sdk === 'react' || sdk === 'rsc',
```
sidmohanty11, line 7: curious on why this doesn't work in the react SDK, `props.symbol.data` should update in this case causing a re-render right?

sidmohanty11 (approved):

packages/sdks-tests/src/e2e-tests/countdown.spec.ts
samijaber: yeah, that's the bug we need to solve. I couldn't fix it quickly enough but wanted to commit the test for future. 

Most surprising is that this works in react native SDK. I can't see how it handles bindings any differently than react SDK.

PR #3752, opened 2024-11-20
fix[gen1][core] ENG-7515 add query.id when hitting content API for symbols

clyde-builderio:
## Description

In Gen1 SDK, when content API is hit for Symbols, it does not render symbols properly

**Jira ticket**
https://builder-io.atlassian.net/browse/ENG-7515

**Loom**
https://www.loom.com/share/90d3b6e317bb493e98ad418ac2f0b7a0


packages/core/src/builder.class.ts
```diff
@@ -2633,6 +2633,10 @@ export class Builder {
       }
     }
 
+    if (apiEndpoint === 'content' && queue[0].model === 'symbol') {
+      queryParams['query.id'] = queue[0].entry;
+    }
+
```
samijaber, line 2639: We should set this in `Symbol.tsx` instead. I think it can go in here: https://github.com/BuilderIO/builder/blob/bff68585eca182176952265a4a0c1f52caabb3cc/packages/react/src/blocks/Symbol.tsx#L159-L165

The reason for that is because the model name might be something other than `symbol`

clyde-builderio: The change is working as expected. Here is a loom with a demo of the changes https://www.loom.com/share/083d6711f94048ce8b88c514c71012e4

packages/react/src/blocks/Symbol.tsx
```diff
@@ -160,6 +160,9 @@ class SymbolComponent extends React.Component<PropsWithChildren<SymbolProps>> {
                     ...(!this.isEditingThisSymbol && {
                       key: builderComponentKey,
                       noEditorUpdates: true,
+                      query: {
```
samijaber, line 163: this query param should not be condition in editingThisSymbol. it should always be provided

samijaber (changes requested):
the query param is in a condition it shouldn't be in

packages/react/src/blocks/Symbol.tsx
samijaber: also, this query param should only be added when the entry value exists AND when the apiEndpoint value is "content"

packages/core/src/builder.class.ts
```diff
@@ -2634,7 +2634,6 @@ export class Builder {
     const isApiCallForCodegenOrQuery = isApiCallForCodegen || this.apiEndpoint === 'query';
 
     if (this.apiEndpoint === 'content') {
-      queryParams.enrich = true;
```
samijaber, line 2637: Why are we removing this arg?

packages/sdks-tests/src/snippet-tests/api-endpoint.spec.ts
```diff
@@ -35,5 +35,24 @@ test.describe('API Endpoint', () => {
...
+    test('records symbol API calls after page load', async ({ page, packageName }) => {
+      test.skip(packageName !== 'gen1-next');
+      // Start monitoring network requests
+      const urlMatchForContentAPI = /https:\/\/cdn\.builder\.io\/api\/v3\/content\/symbol/;
+
+      const responsePromise = page.waitForResponse(urlMatchForContentAPI);
+
+      await page.goto(
+        '/contentwithsymbol?builder.space=ee9f13b4981e489a9a1209887695ef2b&builder.user.permissions=read%2Ccreate%2Cpublish%2CeditDesigns%2CeditLayouts%2CeditLayers%2CeditContentPriority&builder.user.role.name=Designer&builder.user.role.id=creator&builder.cachebust=true&builder.preview=page&builder.noCache=true&builder.allowTextEdit=true&__builder_editing__=true&builder.overrides.page=6a2f7e1a9ab2454ba6766522d6d99f0d&builder.overrides.6a2f7e1a9ab2454ba6766522d6d99f0d=6a2f7e1a9ab2454ba6766522d6d99f0d&builder.overrides.page:/contentwithsymbol=6a2f7e1a9ab2454ba6766522d6d99f0d&builder.frameEditing=page&builder.options.locale=Default',
+        { waitUntil: 'networkidle' }
```
samijaber, line 48: We should avoid `waitUntil` unless absolutely necessary. I don't think we need it in this case.

packages/sdks-tests/src/snippet-tests/api-endpoint.spec.ts
```diff
@@ -35,5 +35,24 @@ test.describe('API Endpoint', () => {
...
+
+    test('records symbol API calls after page load', async ({ page, packageName }) => {
+      test.skip(packageName !== 'gen1-next');
+      // Start monitoring network requests
+      const urlMatchForContentAPI = /https:\/\/cdn\.builder\.io\/api\/v3\/content\/symbol/;
+
+      const responsePromise = page.waitForResponse(urlMatchForContentAPI);
+
+      await page.goto(
+        '/contentwithsymbol?builder.space=ee9f13b4981e489a9a1209887695ef2b&builder.user.permissions=read%2Ccreate%2Cpublish%2CeditDesigns%2CeditLayouts%2CeditLayers%2CeditContentPriority&builder.user.role.name=Designer&builder.user.role.id=creator&builder.cachebust=true&builder.preview=page&builder.noCache=true&builder.allowTextEdit=true&__builder_editing__=true&builder.overrides.page=6a2f7e1a9ab2454ba6766522d6d99f0d&builder.overrides.6a2f7e1a9ab2454ba6766522d6d99f0d=6a2f7e1a9ab2454ba6766522d6d99f0d&builder.overrides.page:/contentwithsymbol=6a2f7e1a9ab2454ba6766522d6d99f0d&builder.frameEditing=page&builder.options.locale=Default',
```
samijaber, line 47: do we need all of the query params in this test?

packages/core/src/builder.class.ts
clyde-builderio: Here is a loom explaining why i removed that line - https://www.loom.com/share/89fbde11974a43839537868e59708e1d

packages/sdks-tests/src/e2e-tests/hit-content-api.spec.ts
```diff
@@ -40,4 +41,41 @@ test.describe('Get Content', () => {
...
+        },
+      });
+    });
+
+    await page.goto('/get-content-with-symbol', { waitUntil: 'networkidle' });
+
+    await expect(x).toBeGreaterThanOrEqual(2);
+
+    urls.forEach(url => {
+      expect(url).toContain('query.id=');
```
samijaber, line 73: can we check that the value exists for sure, just to be safe?

```suggestion
      expect(url).toContain('query.id=foobar');
```

whatever `foobar` actually is

samijaber (approved):
Awesome! 🏆

PR #3755, opened 2024-11-25
[phrasePlugin][Stale Private Key]: 

anaghav2023:
## Description
The Private key in phrase plugin gets updated only in constructor i.e. when the plugin loads. This happens on initial set-up or when we refresh builder web-app. This stale private key causes authorization failure when we switch b/w environments.
The solution is a small fix which gets the latest private key everytime before making api-call.

**Loom showing before after**

https://www.loom.com/share/63c07935720749b7800aaf422b771ae4

### DIFF
--- a/plugins/phrase-conector/src/phrase.ts
+++ b/plugins/phrase-conector/src/phrase.ts
@@ -1,3 +1,4 @@
+import { action } from 'mobx';
 import pkg from '../package.json';
 import appState from '@builder.io/app-context';
 
@@ -24,6 +25,11 @@ export class Phrase {
   constructor(private apiHost?: string) {
     this.loaded = new Promise(resolve => (this.resolveLoaded = resolve));
     this.init();
+    appState.globalState.orgChanged?.subscribe(
+      action(async () => {
+        await this.init();
+      })
+    );
   }
 
   async init() {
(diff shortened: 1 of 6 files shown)


midhunadarvin (approved):
LGTM

plugins/phrase-conector/src/phrase.ts
```diff
@@ -34,11 +34,12 @@ export class Phrase {
   }
 
   async request(path: string, config?: RequestInit, search = {}) {
+    let privateKey = await appState.globalState.getPluginPrivateKey(pkg.name);
```
sanyamkamat, line 37: Nitpick
```suggestion
    const privateKey = await appState.globalState.getPluginPrivateKey(pkg.name);
```

sanyamkamat (commented):
Do add connected Jira ticket in the PR title and description

plugins/phrase-conector/src/phrase.ts
shyam-builder, line 37: This would be expensive, since it would hit firestore everytime instead of once per load. I think that's the reason why it is cached.

Could we look at how many times this URL is called daily to understand impact?

anaghav2023:
> Do add connected Jira ticket in the PR title and description

@sanyamkamat since Jira is our internal project and this is open source, I thought we were not adding jira ids here

murtaza-builder (approved):

PR #3758, opened 2024-11-27
Include custom components subFields inputs for translation

murtaza-builder:
## Description
We now support subFields in inputs of custom component to be sent for translation and update the content when translations are received.

### DIFF
--- a/packages/utils/src/translation-helpers.ts
+++ b/packages/utils/src/translation-helpers.ts
@@ -3,7 +3,7 @@ import traverse from 'traverse';
 import omit from 'lodash/omit';
 import unescape from 'lodash/unescape';
 import set from 'lodash/set';
-import { get } from 'lodash';
+import get from 'lodash/get';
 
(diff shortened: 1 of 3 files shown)


packages/utils/src/translation-helpers.ts
```diff
@@ -165,11 +166,10 @@ export function getTranslateableFields(
         const localizedTextInputs = el.meta.localizedTextInputs as string[];
         if (localizedTextInputs && Array.isArray(localizedTextInputs)) {
           localizedTextInputs
-            .filter(input => el.component?.options?.[input]?.['@type'] === localizedType)
+            .filter(input => get(el.component?.options || {}, `${input}.@type`) === localizedType)
```
sanyamkamat, line 169: How is this different from the previous code?

murtaza-builder: lodash `get` accepts path to fetch an item in an object - could be "title.text". 
Not possible via direct `[]`

sanyamkamat (approved):

PR #3762, opened 2024-11-29
fix[sdks]: `locale` prop to resolve localized values correctly

sidmohanty11:
## Description

A user can use two ways to resolve a localized value: either by passing it through fetchOneEntry/get/getAll and resolving it on the API level, or by providing a `locale` prop which should do it for them. Turns out the `locale` prop was not resolving to the correct values.

We receive localized values in this format:
Screenshot 2024-12-03 at 3 53 02 PM

which errors out as text is now an object.

### Approach:
If a user has not provided locale on the API fetch level then we receive options having these localized objects so if a user has provided a `locale` prop we resolve them to the correct value else we warn the users that they may encounter issues.

Jira
https://builder-io.atlassian.net/browse/ENG-7798

_Screenshot_


packages/sdks/src/functions/extract-localized-values.ts
```diff
@@ -0,0 +1,58 @@
...
+      '[Builder.io] In order to use localized fields in Builder, you must pass a locale prop to the BuilderComponent or to options object while fetching the content to resolve localized fields. Learn more: https://www.builder.io/c/docs/localization-inline#targeting-and-inline-localization'
+    );
+  }
+
+  if (
+    block.component?.options &&
+    containsLocalizedValues(block.component?.options) &&
+    locale
+  ) {
+    block.component.options = extractLocalizedValues(
```
teleaziz, line 51: I think you'd want to use 'traverse' here, to account for sub fields

sidmohanty11: thanks for the catch, will update the logic

samijaber (commented):
This looks great!

Could we get some integration tests as well? With and without subfields to account for Aziz's note too

packages/react/src/functions/extract-localized-values.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+export const extractLocalizedValues = (data: Record<string, any>, locale: string) => {
+  if (!data || !Object.getOwnPropertyNames(data).length) {
+    return {};
+  }
+
+  traverse(data, (value, update) => {
+    if (isLocalizedField(value)) {
+      update(value[locale] ?? undefined);
```
teleaziz, line 28: should this also default to `value.Default` ?

sidmohanty11: yes, I had a discussion with @shyam-builder about this and he was worried that it might cause some confusion showing them content that they wouldn't expect and instead we should be pointing them to use the `locale` prop mandatorily if they are opting in for localization (for this reason we are adding the warning in the editor: https://github.com/BuilderIO/builder-internal/pull/7536) but he'll be able to share more details on this.

teleaziz: I see the point, but should match how the API resolve localized fields when it get a localized param, but maybe this should be:
```ts
const defaultValue = typeof locale === undefined ? undefined : value.Default

update(value[locale] ?? defaultValue);

```

sidmohanty11: yeah, that makes sense actually, and we'd also display the warning in our web app for any previously released versions where locale is not passed so I think we're fine. Will update this

sidmohanty11: Thanks @teleaziz i've updated the logic to fallback to "Default", can you please take a look? Also, we'll always have a "Default" key for a localized field right?

PR #3764, opened 2024-12-03
chore[sdks]: extend allowed file types of Image and Video block

sidmohanty11:
## Description

Extends `allowedFileTypes` of Image and Video blocks so that we can restrict our AssetPicker to use these file types moving forward

Jira
https://builder-io.atlassian.net/browse/ENG-6403

_Screenshot_


clyde-builderio:
For the purpose of consistency would it be okay to make the same change to https://github.com/sidmohanty11/builder/blob/chore/allowed-file-types-img-video/packages/react/src/blocks/raw/Img.tsx#L39

sidmohanty11:
@clyde-builderio you're right, let me make the change

clyde-builderio:
LGTM!

samijaber (approved):

PR #3767, opened 2024-12-04
[sfccPlugin] Validation on sfcc config

anaghav2023:
## Description

In the SFCC plugin, previously, settings were not validated after being saved. This could lead to confusion for users, as any incorrect configuration would only surface as errors when the plugin was used later.

To improve user experience and reduce failure points, we've implemented a "fail-fast" approach. Now, as soon as users enter the configuration details, an authentication call is triggered to validate the settings immediately. This ensures that any issues are identified and communicated to the user upfront.

**Loom**
Before loom:
https://www.loom.com/share/8d4af13e5b184465a5a53369e083de34

After loom:
https://www.loom.com/share/35f74aa2d6414668bf192dcf694083be

### DIFF
--- a/plugins/salesforce-commerce-api-plugin/src/api.ts
+++ b/plugins/salesforce-commerce-api-plugin/src/api.ts
@@ -90,6 +90,19 @@ export class Api {
     }
   }
 
+  async validateConfig(){
+    try{
+      let res = await this.request(`validate-config`);
+      console.log('log4', res);
+      if(res.errors){
+        throw res.errors;
+      }
+      return res;
+    }catch(e){
+      throw e;
+    }
+  }
+
   getProduct(id: string): Promise<any> {
     if (basicCache.get(id)) {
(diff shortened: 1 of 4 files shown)


plugins/salesforce-commerce-api-plugin/src/api.ts
```diff
@@ -90,6 +90,17 @@ export class Api {
...
   }
 
+  async validateConfig(){
+    try{
+       const response = await this.request(`validate-config`);
+       if(response.errors){
+        throw response.errors[0]?.title;
+       }
+    }catch(e){
+      throw e;
```
midhunadarvin, line 100: if we are just throwing error, we can avoid the try catch. Check for `response.errors[0]?.title` at the location where we are catching the error

anaghav2023: right! removed the try catch

midhunadarvin (approved):

PR #3772, opened 2024-12-09
feat: nextjs 15 e2e integration test setup

sidmohanty11:
## Description

https://builder-io.atlassian.net/browse/ENG-7752

_Screenshot_


packages/sdks-tests/src/e2e-tests/ab-test.spec.ts
```diff
@@ -45,7 +45,7 @@ const initializeAbTest = async (
   if (!baseURL) throw new Error('Missing baseURL');
 
   // SSR A/B tests do not seem to work on old NextJS. Likely a config issue.
-  test.skip(packageName === 'gen1-next');
+  test.skip(packageName === 'gen1-next-14' || packageName === 'gen1-next-15');
```
sidmohanty11, line 48: from what i found, if we don't set `builder.canTrack` as false ssr a/b tests work

packages/sdks-tests/src/e2e-tests/tracking.spec.ts
```diff
@@ -30,7 +30,7 @@ test.describe('Tracking', () => {
     }) => {
       // TO-DO: figure out why React gen1 fails this test,
       // track is not called in a published content if builder.canTrack = false
-      test.fail(packageName === 'gen1-react');
+      test.fail(packageName === 'gen1-react' || packageName === 'gen1-next-15');
```
sidmohanty11, line 33: not sure why this is failing, could be the same reason as gen1-react but need to check

sidmohanty11:
yup @samijaber , makes sense. I have renamed the projects. Had a question for you: will we be maintaining next v15 pages? As far as I can see the team is more involved in the recommended app router improvements and most of the docs utilize the same but wanted to know your thoughts

samijaber:
> yup @samijaber , makes sense. I have renamed the projects. Had a question for you: will we be maintaining next v15 pages? As far as I can see the team is more involved in the recommended app router improvements and most of the docs utilize the same but wanted to know your thoughts

Yeah, we should maintain both, many customers are still on `pages/`. As long as Vercel supports it officially, we'll keep testing on it.

packages/sdks-tests/src/e2e-tests/ab-test.spec.ts
samijaber: Ahh, good find! so this:

https://github.com/BuilderIO/builder/blob/e77aae4d8ad29fb03c65c8cb51a0add7f3c17c0b/packages/react-tests/nextjs/pages/[[...index]].tsx#L36-L38

is breaking SSR tests. We can probably fix it then by including SSR tests in this clause: 

https://github.com/BuilderIO/builder/blob/e77aae4d8ad29fb03c65c8cb51a0add7f3c17c0b/packages/react-tests/nextjs/pages/[[...index]].tsx#L46-L51

sidmohanty11: ahh makes sense now on why the call was still happening, as we use it inside useEffect, the first track fires off before reaching that check. Have updated the builder.canTrack calls to happen on global level before component initialization which fixes our issue

packages/sdks-tests/src/e2e-tests/tracking.spec.ts
sidmohanty11: @samijaber I think we might need to re-check this for gen1, I can see pages with symbols not respecting canTrack. Could this be because of some recent updates?

clyde-builderio:
LGTM

packages/sdks-tests/src/e2e-tests/ab-test.spec.ts
```diff
@@ -44,9 +44,6 @@ const initializeAbTest = async (
 ) => {
   if (!baseURL) throw new Error('Missing baseURL');
 
-  // SSR A/B tests do not seem to work on old NextJS. Likely a config issue.
-  test.skip(packageName === 'gen1-next');
```
samijaber, line 48: 🎉

packages/react-tests/react-vite/src/App.tsx
```diff
@@ -9,6 +9,14 @@ builder.init(getAPIKey());
 // default to not tracking, and re-enable when appropriate
 builder.canTrack = false;
 
+if (
+  typeof window !== 'undefined' &&
+  !window.location.pathname.includes('can-track-false') &&
+  !window.location.pathname.includes('symbol-tracking')
+) {
+  builder.canTrack = true;
```
samijaber, line 17: only issue is im not sure if this will work on tests that include client-side navigations? This code should only run a single time when the JS is first loaded into the browser.

Just something to keep in mind

samijaber (approved):

packages/react-tests/react-vite/src/App.tsx
sidmohanty11: got it, so as canTrack=false tests are only on singular pages we don't face this

PR #3773, opened 2024-12-09
[sfccPlugin]: Improved Error message

anaghav2023:
## Description
Getting rid of the generic error message when we know that error has come from authentication failure

_Screenshot_

Screenshot 2024-12-09 at 5 55 55 PM


plugins/salesforce-commerce-api-plugin/src/api.ts
```diff
@@ -93,7 +93,7 @@ export class Api {
   async validateConfig(){
        const response = await this.request(`validate-config`);
        if(response.errors){
-        throw response.errors[0]?.title;
+        throw 'We failed to authenticate your access to salesforce commerce api. Please review all plugin fields and make sure they are correct.';
```
samijaber, line 96: sweet. Can you make sure that the spelling is exactly correct? From looking it up here: https://github.com/SalesforceCommerceCloud/commerce-sdk 

it looks like the exact name is "Salesforce Commerce Cloud B2C Commerce API", but you probably know best.

anaghav2023: Done

samijaber (approved):

PR #3782, opened 2024-12-13
fix[gen1][gen2] ENG-7760 onChange does not fire when an `async` function is passed to it

clyde-builderio:
## Description

onChange does not fire when an `async` function is passed to it

**Jira ticket**
https://builder-io.atlassian.net/browse/ENG-7760

**Loom**
Gen1: https://www.loom.com/share/b6012b2b37f54944ab94074c264f3bb6
Gen2: https://www.loom.com/share/ca3d895ddf5d4b69b13eab341003dde1

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -1104,9 +1104,15 @@ export class Builder {
       // 2. `name(args) => {code}`
       // 3. `(args) => {}`
       // 4. `args => {}`
+      // 5. `async name(args) => {code}`
+      // 6. `async (args) => {}`
+      // 7. `async args => {}`
       const isArrowWithoutParens = /^[a-zA-Z0-9_]+\s*=>/i.test(fnStr);
       const appendFunction =
-        !fnStr.startsWith('function') && !fnStr.startsWith('(') && !isArrowWithoutParens;
+        !fnStr.startsWith('function') &&
+        !fnStr.startsWith('async') &&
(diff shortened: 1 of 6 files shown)


packages/core/src/builder.class.ts
```diff
@@ -1104,9 +1104,15 @@ export class Builder {
...
       // 3. `(args) => {}`
       // 4. `args => {}`
+      // 5. `async name(args) => {code}`
+      // 6. `async (args) => {}`
+      // 7. `async args => {}`
       const isArrowWithoutParens = /^[a-zA-Z0-9_]+\s*=>/i.test(fnStr);
       const appendFunction =
-        !fnStr.startsWith('function') && !fnStr.startsWith('(') && !isArrowWithoutParens;
+        !fnStr.startsWith('function') &&
+        !fnStr.startsWith('async') &&
```
sidmohanty11, line 1113: great find! can we add an e2e test for this?

samijaber: might be hard to write an e2e test for this given that it's serializing functions sent to the editor, but at the very least we should write some unit tests for `serializeFn` with all the different inputs we expect:
- starts with function
- starts with async
- with parens
- without parens

sidmohanty11 (commented):
can we add some tests for:
- starts with just `function`
- with parens (no async) ?

overall looks good to me, needs a changeset

sidmohanty11 (approved):
🥇

PR #3786, opened 2024-12-17
Sample Chrome extension for modifying `set-cookie` responses

shyam-builder:
## Description

Creates a sample chrome extension that still relies on Manifest v2 to modify `set-cookies` headers on the fly to make it more permissive


samijaber (commented):
should this be in `examples/chrome-extension`? I assume it's not intended for customers to build and compile, no?

PR #3787, opened 2024-12-17
fix[gen1][core] ENG-7841 change to queryParams

clyde-builderio:
## Description

Gen1 SDK: use same Content API defaults as Gen2 SDK


packages/core/src/builder.class.ts
```diff
@@ -2640,7 +2640,9 @@ export class Builder {
     const isApiCallForCodegenOrQuery = isApiCallForCodegen || this.apiEndpoint === 'query';
 
     if (this.apiEndpoint === 'content') {
-      queryParams.enrich = true;
+      queryParams.noTraverse = queue[0].limit !== 1;
+      queryParams.includeRefs = true;
+
```
samijaber, line 2645: We should move this logic before the following code blocks:

-  https://github.com/BuilderIO/builder/blob/1908d27cb9e27aac831a451ae3a66a842252ffb1/packages/core/src/builder.class.ts#L2604-L2616
- https://github.com/BuilderIO/builder/blob/1908d27cb9e27aac831a451ae3a66a842252ffb1/packages/core/src/builder.class.ts#L2627-L2630

These are two ways customers can override properties, so they need to exist after you set the `noTraverse` and `includeRefs` defaults.


I also just noticed these two `noTraverse` settings: https://github.com/BuilderIO/builder/blob/1908d27cb9e27aac831a451ae3a66a842252ffb1/packages/core/src/builder.class.ts#L2495-L2497
https://github.com/BuilderIO/builder/blob/1908d27cb9e27aac831a451ae3a66a842252ffb1/packages/core/src/builder.class.ts#L2863-L2867

This means that `noTraverse` is already handled (set to `true` in `getAll()` which has `limit > 1`. So you don't need to set `noTraverse`, you only need to add the `includeRefs = true`.

TLDR:
- remove `noTraverse` default
- move the `includeRefs` default higher up so it can still be overridden by customer
- add some tests that show:
  - default value of `includeRefs` is `true`
  - if customer provides `includeRefs: false` to `builder.get()` or `builder.getAll()`, then they are able to update the URL

.changeset/famous-pants-agree.md
```diff
@@ -0,0 +1,6 @@
+---
+'@builder.io/sdk': patch
+'@builder.io/react': patch
+---
+
+Fix: Content API invocations will have `includeRefs` set to `false` by default and will not have `enrich` as part of its query-params
```
sidmohanty11, line 6: `includeRefs` default should be set to `true` right?

clyde-builderio: Yes, good catch, miss from my side

samijaber (approved):
Awesome, thanks for the quick turnaround on this

.changeset/famous-pants-agree.md
```diff
@@ -0,0 +1,6 @@
+---
+'@builder.io/sdk': patch
+'@builder.io/react': patch
+---
+
+Fix: Content API invocations will have `includeRefs` set to `true` by default and will not have `enrich` as part of its query-params
```
samijaber, line 6: I would prefer not to mention `enrich` at all in here as it might confuse customers:

```suggestion
Fix: Content API invocations will have `includeRefs` set to `true` by default.
```

samijaber (changes requested):
I just realized, `noTraverse: false` is not provided by default in `builder.get()`. I was wrong about that part, and it's something we need to add in this PR

packages/core/src/builder.class.test.ts
```diff
@@ -915,7 +915,145 @@ describe('flushGetContentQueue', () => {
...
+    builder.apiEndpoint = 'content';
+    builder.getLocation = jest.fn(() => ({
+      search: `?builder.params=includeRefs%3D${includeRefs}`,
+    }));
+
+    await builder.get(expectedModel, {});
+
+    expect(builder['makeFetchApiCall']).toBeCalledTimes(1);
+    expect(builder['makeFetchApiCall']).toBeCalledWith(
+      `https://cdn.builder.io/api/v3/content/${expectedModel}?omit=meta.componentsUsed&apiKey=${API_KEY}&userAttributes=%7B%22device%22%3A%22desktop%22%7D&includeRefs=${includeRefs}&model=%22${expectedModel}%22`,
```
samijaber, line 959: looking at this default value for `builder.get()`, i realize that `noTraverse=false` is missing and should be here.

packages/core/src/builder.class.ts
```diff
@@ -2594,6 +2594,13 @@ export class Builder {
         queryParams.staleCacheSeconds = options.staleCacheSeconds;
       }
 
+      if (this.apiEndpoint === 'content') {
+        queryParams.includeRefs = true;
+        if (!('noTraverse' in options)) {
+          queryParams.noTraverse = false;
+        }
```
sidmohanty11, line 2601: can we move this to `get()`? similarly how we do it in case of `getAll`?

packages/core/src/builder.class.ts
```diff
@@ -2227,6 +2227,11 @@ export class Builder {
     let instance: Builder = this;
     let finalLocale =
       options.locale || options.userAttributes?.locale || this.getUserAttributes().locale;
+
+    if (!('noTraverse' in options)) {
```
sidmohanty11, line 2231: should it be?
```suggestion
    if (!('noTraverse' in options) && this.apiEndpoint === 'content') {
```

samijaber (approved):

packages/core/src/builder.class.ts
sanyamkamat: When will there be any `apiEndpoint` which will not be `content`?

sanyamkamat: @samijaber Replied to your question in slack thread.

sidmohanty11: makes sense

PR #3789, opened 2024-12-18
fix[gen1-react]: hydration errors in Next 15

sidmohanty11:
## Description

https://builder-io.atlassian.net/browse/ENG-7608

Loom
https://www.loom.com/share/b03efe8502fd4c67a71279993fda6752

### DIFF
--- a/.changeset/healthy-ads-end.md
+++ b/.changeset/healthy-ads-end.md
@@ -1,0 +1,5 @@
+---
+"@builder.io/react": patch
+---
+
+Fix: hydration errors in Next v15 while a user is editing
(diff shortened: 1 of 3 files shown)


teleaziz (approved):

samijaber (approved):
nice. I'm surprised this is an issue in next15 only. Maybe the underlying React version in next 15 is doing something new in how it sets `builder-path={''}` on SSR?

sidmohanty11:
@samijaber I checked on a next 14 app for the difference and looks like we were getting these hydration errors as warnings on the console but not as pop-up errors. Next 15 bumped up the strictness i think

samijaber:
@sidmohanty11 thanks for investigating and getting to the bottom of it. 👍🏽

PR #3726, opened 2024-11-07
fix[gen2-sdks]: able to edit and preview when we have no content

sidmohanty11:
## Description

Jira
https://builder-io.atlassian.net/browse/PLAN-338

Fixes #3653

_Screenshot_
If relevant, add a screenshot or two of the changes you made.

PR #3795, opened 2024-12-20
fix: hydration errors in next pages

sidmohanty11:
## Description

#3726 causes hydration errors in next pages combination, to avoid that converted logic such that the div only gets attached to the SDKs that require it for visual editing

_Screenshot_

### DIFF
--- a/packages/sdks/src/components/content/components/enable-editor.helpers.ts
+++ b/packages/sdks/src/components/content/components/enable-editor.helpers.ts
@@ -13,6 +13,10 @@ export const SDKS_USING_ELEMENT_REF_APPROACH = [
   'vue',
 ] as Target[];
 
+/**
+ * We need to attach this div only when content exists or isPreviewing/isEditing even when content is null,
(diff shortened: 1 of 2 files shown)


packages/sdks/src/components/content/components/enable-editor.helpers.ts
```diff
@@ -0,0 +1,22 @@
...
+  'vue',
+] as Target[];
+
+/**
+ * We need to attach this div only when content exists or isPreviewing/isEditing even when content is null,
+ * as we need to set the elementRef and allow previewing and visual editing
+ */
+export const needsElementRefDivForEditing =
+  SDKS_USING_ELEMENT_REF_APPROACH.includes(TARGET) &&
+  (isEditing() || isPreviewing());
```
samijaber, line 22: I wonder if `needsElementRefDivForEditing` should be a function, so that if query params change, we're always computing the latest `isEditing()` and `isPreviewing()`... 🤔

samijaber (approved):

PR #3805, opened 2024-12-30
fix[sdkCore][Symbols]: ENG-7640 Set preview to true only if its not a symbol call

anaghav2023:
## Description
If we are using Symbol in any page. It reflects the unpublished changes. This is happening only on the content editor and not the actual website.

**Before/After loom**
https://www.loom.com/share/4ac7c396c70a4887a19b9f0eec295c63

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -2630,15 +2630,18 @@ export class Builder {
         }
       }
     }
-    if (this.preview) {
-      queryParams.preview = 'true';
-    }
-    const hasParams = Object.keys(queryParams).length > 0;
 
     // TODO: option to force dev or qa api here
     const host = this.host;
 
     const keyNames = queue.map(item => encodeURIComponent(item.key!)).join(',');
+
+    const isSymbol = keyNames.includes('symbol');
+
+    if (this.preview && !isSymbol) {
+      queryParams.preview = 'true';
+    }
+    const hasParams = Object.keys(queryParams).length > 0;
 
     if (this.overrideParams) {
       const params = omit(QueryString.parse(this.overrideParams), 'apiKey');
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -2630,18 +2630,15 @@ export class Builder {
         }
       }
     }
+    if (this.preview && this.previewingModel === queue?.[0]?.model) {
+      queryParams.preview = 'true';
+    }
+    const hasParams = Object.keys(queryParams).length > 0;
 
     // TODO: option to force dev or qa api here
     const host = this.host;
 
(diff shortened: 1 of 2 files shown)


builder-adi (approved):
LGTM

packages/core/src/builder.class.ts
```diff
@@ -2630,16 +2630,19 @@ export class Builder {
...
-    const hasParams = Object.keys(queryParams).length > 0;
 
     // TODO: option to force dev or qa api here
     const host = this.host;
 
     const keyNames = queue.map(item => encodeURIComponent(item.key!)).join(',');
 
+    const isSymbol = keyNames.includes('symbol');
+
+    if (this.preview && !isSymbol) {
```
sidmohanty11, line 2641: rather than checking for 'symbol' should we check if the `previewingModel === modelPassed`?

```suggestion
    if (this.preview && this.previewingModel === parseLogic(keyNames)) { // parse to take out anything before %
```

or

```suggestion
    if (this.preview && this.previewingModel === queue[0].model)) {
```

anaghav2023: @sidmohanty11 the issue here is that this.previewingModel is always coming as 'page' in both the page call as well as the symbol call. While the model in the queue is coming as 'page' for page call and 'symbol' for symbol call

sidmohanty11: is that not what we want? `previewingModel` for page will be `page` === `queue[0].model` where we want preview=true and later for symbol call `previewingModel` is still `page` which is not equal to `symbol` model where we shouldn't add preview right or am I missing something?

anaghav2023: Right ! makes sense. I got confused. I have pushed the changes

anaghav2023:
> LGTM! @anaghav2023
> 
> Last thing: it needs changeset. For that you can run `yarn g:changeset` and add changesets (patch update) for both gen1 React (`@builder.io/react`) and gen1 core (`@builder.io/sdk`) SDKs

@sidmohanty11 I have made change only in the core still I need to add changesets for react as well? Then do I need to add it for other frameworks as well?

sidmohanty11:
@anaghav2023 as `@builder.io/sdk` (core) is a dependency of `@builder.io/react` (gen1 react), the bot will trigger a version bump of the gen1 react package as well, using the latest version. So it's better if we add the changeset for it so that it reflects what came in - in the latest release (which is just the version bump). Otherwise, it will bump the version, but no changelog will be added. We don't need to add for other SDKs (`@builder.io/sdk-*`) as they are Gen2 SDKs, and they don't have the gen1 core SDK as their dependency; they are solely present in the `packages/sdks/*` directory.

sidmohanty11 (approved):
💯

PR #3806, opened 2024-12-30
fix[angular-gen2]: add angular 19 ssr e2e and fix hydration errors

sidmohanty11:
## Description

Fixes the hydration error thrown from angular for using ngComponentOutlet, adds angular-19-ssr to e2e and an example of how users can override fetch to use HTTP client from angular that should help in caching HTTP requests

Jira
https://builder-io.atlassian.net/browse/ENG-7946
https://builder-io.atlassian.net/browse/ENG-7945
https://builder-io.atlassian.net/browse/ENG-7665

_Screenshot_


packages/sdks-tests/src/e2e-tests/custom-components.spec.ts
```diff
@@ -24,7 +24,7 @@ test.describe('Custom components', () => {
     packageName,
     sdk,
   }) => {
-    test.skip(!['angular', 'react'].includes(sdk));
+    test.skip(!['angular-16', 'react'].includes(sdk));
```
samijaber, line 27: This looks like a wrong check: you're comparing a `packageName` (`angular-16`, etc.) to `sdk` (should always be `angular` for all 3 angular e2e servers)

samijaber: This is the case for many other checks changed in `.spec.ts` files in this PR.

packages/sdks/mitosis.config.cjs
```diff
@@ -718,6 +718,25 @@ const ANGULAR_COMPONENT_REF_UPDATE_TEMPLATE_SSR = () => ({
   },
 });
 
+/**
+ * Angular doesn't support hydration for components created dynamically.
```
samijaber, line 722: can you link to any docs or code that expands on this?

packages/sdks/mitosis.config.cjs
```diff
@@ -718,6 +718,25 @@ const ANGULAR_COMPONENT_REF_UPDATE_TEMPLATE_SSR = () => ({
...
+        const componentDecorator = code.match(/@Component\s*\({/)[0];
+        const componentDecoratorWithHost = componentDecorator.replace(
+          /@Component\s*\({/,
+          `@Component({\n\thost: { ngSkipHydration: "true" },`
+        );
+        code = code.replace(componentDecorator, componentDecoratorWithHost);
+      }
+      return code;
+    },
+  },
```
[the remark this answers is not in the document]
sidmohanty11: Content will be ssred but the ssred DOM structure won't be reused in the client and that will perform hydration again. This shouldn't break any SDK features but might affect site perf a bit

packages/sdks-tests/src/e2e-tests/custom-components.spec.ts
sidmohanty11: ahh my bad, I might have used replace all. Thanks for the catch, have updated it

packages/sdks/mitosis.config.cjs
sidmohanty11: added

samijaber: Thanks!

packages/sdks/mitosis.config.cjs
```diff
@@ -763,6 +763,27 @@ const ANGULAR_MARK_SAFE_INNER_HTML = () => ({
   },
 });
 
+/**
+ * Angular doesn't support hydration for components created dynamically.
+ * Refer: https://angular.dev/errors/NG0503
+ * GitHub issue: https://github.com/angular/angular/issues/51798
```
samijaber, line 769: Does this issue resolve content duplication as well like https://builder-io.atlassian.net/browse/ENG-8042 ?

samijaber (approved):
🙏🏽

PR #3809, opened 2025-01-02
fix[angular-gen2]: embed and custom code block and text block angular sanitizing html warning

### LINKED ISSUE #3793: Angular Gen2 - HTML sanitization warning
The content component is emitting this warning: `WARNING: sanitizing HTML stripped some content, see https://g.co/ng/security#xss`

I'm assuming somewhere in the library innerHTML is being set and should be using the DomSanitizer service to mark the content as safe.

sidmohanty11:
## Description

Uses https://angular.dev/api/platform-browser/DomSanitizer to mark text as safe html to use inside innerHTML. As text is already sanitized from the HTML editor we should be able to bypass it safely. This only shows up in the dev server and we don't receive these warnings in a prod build.

Fixes Embed and Custom Code block wrapping them inside `sanitizer.bypassSecurityTrustHtml` so that they allow embedding of iframe

Fixes #3793 

Jira
https://builder-io.atlassian.net/browse/ENG-7950
https://builder-io.atlassian.net/browse/ENG-7901

_Screenshot_
Before
Screenshot 2025-01-02 at 6 43 43 PM
Screenshot 2025-01-02 at 6 43 50 PM
After
Screenshot 2025-01-02 at 6 42 52 PM
Screenshot 2025-01-02 at 6 45 39 PM


samijaber (approved):

samijaber:
Have these angular tests been failing sporadically before this PR?

sidmohanty11:
@samijaber no these are related to changes I've made here. Turns out the text component is triggering `ngAfterViewChecked` if the parent dom element has a mouseenter event listener calling the getter frequently when we hover and `sanitizer.bypassSecurityTrustHtml(processedText)` is creating a new object everytime its called resulting in non-interactivity. This I think should be fixed from #3812 because it removes the getter and the text will only update when its updated from the block itself

packages/sdks/mitosis.config.cjs
```diff
@@ -718,6 +718,55 @@ const ANGULAR_COMPONENT_REF_UPDATE_TEMPLATE_SSR = () => ({
...
+          throw new Error(
+            'constructor found which should not be here. If you see this, please fix the ANGULAR_TEXT_MARK_SAFE_HTML Plugin.'
+          );
+        }
+
+        const variableName = code.match(/\[innerHTML\]="([^"]+)"/)?.[1];
+        if (variableName) {
+          code = code.replace(
+            `[innerHTML]="${variableName}"`,
+            `[innerHTML]="sanitizer.bypassSecurityTrustHtml(${variableName})"`
```
sidmohanty11, line 761: @samijaber this was needed for Embed and Custom Code block as well so I extended this PR to fix them as well

samijaber: long-term I think this should be baked into mitosis: the angular generator should always use `sanitizer.bypassSecurityTrustHtml` for `innerHTML` (maybe with a flag to force-disable the sanitizer if needed)

sidmohanty11: great idea, i've created a ticket for it

packages/sdks/mitosis.config.cjs
```diff
@@ -718,6 +718,55 @@ const ANGULAR_COMPONENT_REF_UPDATE_TEMPLATE_SSR = () => ({
...
 });
 
+const ANGULAR_MARK_SAFE_INNER_HTML = () => ({
+  code: {
+    post: (code) => {
+      if (
+        code.includes('selector: "builder-text"') ||
+        code.includes('selector: "builder-embed"') ||
+        code.includes('selector: "custom-code"')
+      ) {
```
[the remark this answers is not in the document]
sidmohanty11: done!

PR #3814, opened 2025-01-13
perf[react-native]: Memoized Blocks Component to free up UI thread.

yash-builder:
## Description

`Suspense`:
Added `React.Suspense` to defer the rendering of the `Content` component until it is fully loaded.

`Memoization`:
`React.memo` to memoize computationally expensive operations within the `Blocks` component. To prevents unnecessary recalculations of processed blocks during re-renders, reducing the load on the UI thread.

_Screenshot_
Before:
Screenshot 2025-01-13 at 12 20 38 PM

After:
Screenshot 2025-01-13 at 12 37 41 PM


packages/sdks/overrides/react-native/src/components/blocks/blocks.tsx
```diff
@@ -0,0 +1,96 @@
...
+      key={block.id}
+      block={block}
+      linkComponent={linkComponent}
+      context={context}
+      registeredComponents={registeredComponents}
+    />
+  ));
+  
+  // Optional: Give the component a display name for better debugging
+  RenderBlock.displayName = 'RenderBlock';
```
midhunadarvin, line 36: 🎯  **suggestion:** ‏
Should we append `block.id` to the displayName as a identifier ?

packages/sdks/overrides/react-native/src/components/blocks/blocks.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+type RenderBlockProps = {
+    item: any;  // Replace 'any' with your block type
+    linkComponent?: React.ComponentType<any>;
+    context?: any;  // Replace with your context type
+    registeredComponents?: any;  // Replace with your components type
+  };
+  
+  // Moved outside and memoized
+  const RenderBlock = memo(({
```
midhunadarvin, line 20: ❔**question:** 
We have used `memo` over Block component ( to get RenderBlock ) and then composed the `RenderBlock` with `React.useCallback`. Could you explain the logic here and how it works internally ?

- what happens if we use `memo` / `useCallback` in both places ?

examples/nextjs-app-dir-v2/package.json
```diff
@@ -9,7 +9,7 @@
     "lint": "next lint"
   },
   "dependencies": {
-    "@builder.io/sdk-react": "^1.0.23",
+    "@builder.io/sdk-react": "^3.0.1",
```
midhunadarvin, line 12: ❔**question:** ‏
Why do we have changes in the `nextjs-app-dir-v2` ?

midhunadarvin (commented):
Awesome work @yash-builder

packages/sdks/overrides/react-native/src/components/blocks/blocks.tsx
yash-builder: We can surely do that. But in this scenario I used this identifier for inspecting the component in React Devtools

packages/sdks/overrides/react-native/src/components/blocks/blocks.tsx
yash-builder: Great question!
- If Only `memo` Is Used:
`RenderBlock` will avoid re-renders, but `FlatList` may still recreate the renderItem function on every render. This may lead to potential performance bottleneck.
- If Only `useCallback` Is Used:
The `renderItem` function will be stable, but `RenderBlock` will re-render if its props are shallowly different, even when not required.

examples/nextjs-app-dir-v2/package.json
yash-builder: Ah, nice catch!  
I was testing a few other things and might have accidentally pushed the changes.

samijaber (changes requested):
Very well done! I'm Rewriting here what I sent you in Slack so we don't lose track of it.

One important thing is not to use `overrides/*` unless as a last resort because the file will quickly go out of sync with the original `Blocks.lite.tsx`, forcing us to maintain multiple copies of the same logic.

Here is a Loom showing how I'd tackle the ~3-4 changes in your PR and incorporate them into Mitosis plugins instead: 

https://www.loom.com/share/94f1767796bd487bb48be41d9046e01f


Also, you need to add a changesets entry, which you can do by following https://github.com/BuilderIO/builder/blob/main/packages/sdks/PUBLISHING.md

packages/sdks/mitosis.config.cjs
```diff
@@ -163,6 +163,76 @@ const ADD_IS_STRICT_STYLE_MODE_TO_CONTEXT_PLUGIN = () => ({
...
+    post: (code, json) => {
+      if (json.name === 'Blocks') {
+        return code.replace(
+          'export default Blocks',
+          'export default memo(Blocks)'
+        );
+      }
+      return code;
+    },
+  },
```
samijaber, line 233: Initially in this PR, you were memoizing `Block`, but now you're memoizing `Blocks` instead.

Was this intended? Do we still get the same performance benefits from memoizing the parent component instead of the individual items?

examples/nextjs-app-dir-v2/package.json
samijaber: All good, we should be updating these examples to latest SDK versions anyways

packages/sdks/mitosis.config.cjs
yash-builder: Hey Sami, In the earlier PR also we were memoizing `Blocks.tsx` and not `Block.tsx`

samijaber: Ah indeed, I missed that you memoized `Blocks.tsx` in your original PR commit.


but you were _also_ creating a memoized version of Block called RenderBlock : [https://github.com/BuilderIO/builder/pull/3814/commits/e264c120fd38851208875a9915b55e26e952528f#diff-b9cbe94aa51d49deda0[…]fd13d38a1f56e5641441f5R20](https://github.com/BuilderIO/builder/pull/3814/commits/e264c120fd38851208875a9915b55e26e952528f#diff-b9cbe94aa51d49deda00d636a12ece211317400d8ffd13d38a1f56e5641441f5R20)

CleanShot 2025-01-21 at 07 25 26@2x

Which is now gone from this PR. Was it unnecessary?

yash-builder: Ah my bad I missed this part. Also one question I want to ask by using mitosis how can I write code outside the `Blocks` function but inside the `Blocks.tsx`

samijaber: in my Loom I explan that in this case, you probably want to add `export default Memo(Block)` to `Block.tsx`, and remove this `RenderBlock` from `Blocks.tsx` to achieve the same outcome

yash-builder: gotcha!
Asking out of curiosity. Is there a way where we can declare variables/functions outside Functional/Class component using `mitosis.config.js`?

samijaber: Yes actually, there's a `preComponent` hook in Mitosis whose code will be printed right after import statements (and therefore, outside React components)

.changeset/early-pillows-jog.md
```diff
@@ -0,0 +1,7 @@
+---
+"@builder.io/sdk-react-native": major
```
samijaber, line 2: Not major, that's reserved for big breaking changes. You'll want to go with a patch change here:

```suggestion
"@builder.io/sdk-react-native": patch
```

samijaber (approved):
🎉

PR #3829, opened 2025-01-15
fix[reactSdk][gen1]: ENG-7900 Dynamic Symbol in Visual Editor loads the wrong Symbol type

midhunadarvin:
## Description

- When 2 or more dynamic symbols are used in a content, it was showing the same preview for all the symbols on initial load in the content editor
- For dynamic symbols we were passing the React key as `undefined` hence on initial render all the symbol blocks had the same key, now we are setting the key as the block id.
- This issue is only on the gen 1 react sdk.

**Jira Ticket**
https://builder-io.atlassian.net/browse/ENG-7900

**Loom**
https://www.loom.com/share/831e4bc4816845269769b323b5696765

### DIFF
--- a/.changeset/famous-camels-breathe.md
+++ b/.changeset/famous-camels-breathe.md
@@ -2,4 +2,4 @@
 "@builder.io/react": patch
 ---
 
-Fixed issue with dynamic symbols not rendering properly on visual editor
+Fix: Dynamic symbols not rendering properly in Visual Editor.
--- a/.changeset/famous-camels-breathe.md
+++ b/.changeset/famous-camels-breathe.md
@@ -1,0 +1,5 @@
+---
+"@builder.io/react": patch
+---
+
+Fixed issue with dynamic symbols not rendering properly on visual editor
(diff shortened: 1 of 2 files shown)


samijaber (commented):
Awesome fix @midhunadarvin.

- You need to add changesets in this PR. See https://github.com/BuilderIO/builder/blob/main/packages/sdks/PUBLISHING.md
- Is it possible to add an integration test to avoid regressions? See https://github.com/BuilderIO/builder/blob/main/packages/sdks/docs/DEVELOP.md for how to add an integration test to our SDKs

.changeset/famous-camels-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@builder.io/react": patch
+---
+
+Fixed issue with dynamic symbols not rendering properly on visual editor
```
samijaber, line 5:
```suggestion
Fix: Dynamic symbols not rendering properly in Visual Editor.
```

samijaber (approved):

PR #3855, opened 2025-01-24
fix[react]: ENG-8091 add check around builder block id

sanyamkamat:
## Description

We are getting block errors whenever we try to access `startsWith` method without checking if parent `id` field is there or not. This PR adds checks around that.

### DIFF
--- a/packages/react/src/components/builder-block.component.tsx
+++ b/packages/react/src/components/builder-block.component.tsx
@@ -664,12 +664,12 @@ export class BuilderBlock extends React.Component<
     );
   }
 
-  get id() {
+  get id(): string {
     const { block } = this;
-    if (!block.id!.startsWith('builder')) {
+    if (!block.id?.startsWith('builder')) {
       return 'builder-' + block.id;
     }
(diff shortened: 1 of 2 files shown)


packages/react/src/components/builder-block.component.tsx
```diff
@@ -664,12 +664,12 @@ export class BuilderBlock extends React.Component<
     );
   }
 
-  get id() {
+  get id(): string {
     const { block } = this;
-    if (!block.id!.startsWith('builder')) {
+    if (block.id && !block.id.startsWith('builder')) {
```
shyam-builder, line 669: Not this?
```suggestion
    if (block.id?.startsWith('builder')) {
```

sanyamkamat: we need to add `builder-` only if its _not_ there.

shyam-builder (approved):
So to recap
- we need to check for block to have an id and startsWith, so we can't do the ?
- the '' should not break anything as it is mostly to account for the types

PR #3856, opened 2025-01-24
feat[webapp][breakpoints]: ENG-7984 Add another breakpoint

midhunadarvin:
**Description**

- add `xsmall` breakpoints for gen1 react sdk

**Link to JIRA ticket:**
https://builder-io.atlassian.net/browse/ENG-7984

**Loom**
https://www.loom.com/share/f5f262080f7040f89eb976c5066dd237


packages/react/src/components/builder-block.component.tsx
```diff
@@ -181,7 +181,15 @@ export class BuilderBlock extends React.Component<
     const reversedNames = sizeNames.slice().reverse();
     const styles: any = {};
     if (responsiveStyles) {
+      const contentHasXSmallBreakpoint = Boolean(
```
sanyamkamat, line 184: Suggestion: we could avoid this flag and the check in the for loop later. That way we will treat xsmall as existing medium or small sizes.

packages/react/src/constants/device-sizes.constant.ts
```diff
@@ -21,8 +21,8 @@ const sizes = {
     max: 991,
   },
   large: {
-    min: 990,
-    default: 991,
+    min: 992,
```
sanyamkamat, line 24: Given the default `large` having bigger implications, we should avoid this change for now.

midhunadarvin: Yes, I did not want to make this change in default sizes. But if you look at the medium max, it is 991px whereas in the large min, it is 990px. There is some overlap. So to avoid the overlap I had made the change.

On searching in the code, it seems the large min is not being used anywhere. So I'll revert here.

packages/react/src/components/builder-block.component.tsx
midhunadarvin: But in the loop also we'll need to have a separate if clause `size === 'xsmall'` right ? Currently it is treating the xsmall the same way as medium / small sizes, but with a early return in the case when content doesn't have breakpoints xsmall.

sanyamkamat: Can we add a Note/comment saying that this check can be removed once we make `xsmall` becomes GA

packages/react/src/constants/device-sizes.constant.ts
```diff
@@ -41,40 +41,49 @@ const sizes = {
...
+  if (xsmall) {
+    const xsmallMin = Math.floor(xsmall / 2);
+    newSizes.xsmall = {
+      max: xsmall,
+      min: xsmallMin,
+      default: xsmallMin + 1,
+    };
+  }
 
-  const largeMin = newSizes.medium.max + 1;
```
sanyamkamat, line 72: check if this has to be added back

packages/react/src/components/builder-block.component.tsx
midhunadarvin: done

packages/react/src/constants/device-sizes.constant.ts
midhunadarvin: It seems we are overriding the large min & default in this function. Also couldn't find anywhere where are using `newSizes.large`  in the code. I reverted the changes to keep changes minimal.

sanyamkamat (approved):
LGTM. It will be better if @sidmohanty11 or @samijaber gives a pass at this before merging.

samijaber (changes requested):
Requesting changes due to issues applying styles in my manual tests, and absence of automated testing.

-> I tried to set a background color on an xs size and it doesn't seem to work: https://www.loom.com/share/595042fccb994f1b838a846467015af3

I would like to see some e2e tests added to make sure content renders as expected, and see what the source of my issue above is:

- see https://github.com/BuilderIO/builder/blob/main/packages/sdks/docs/DEVELOP.md on how to write integration tests in this repo
- You can likely reuse or build upon the custom breakpoints tests we have, adding some `xsmall` content in there and testing that it also renders as expected: https://github.com/BuilderIO/builder/blob/main/packages/sdks-tests/src/e2e-tests/custom-breakpoints.spec.ts

Since this PR only introduces `xsmall` to gen1 SDKs, see how we skip tests for certain SDKs using https://github.com/BuilderIO/builder/blob/4a345f86b4e034a1ccd7579d4f01d0a79111b5c4/packages/sdks-tests/src/helpers/index.ts#L156

midhunadarvin:
Hi @samijaber I have updated the PR with the e2e test cases. Also the issue mentioned in your loom was not reproducible for me. https://www.loom.com/share/08c7d737385346b7bf8149ccc86d4a50 . 

Could you share your content URL so that I can also debug.

packages/react/src/constants/device-sizes.constant.ts
```diff
@@ -71,7 +87,7 @@ export const getSizesForBreakpoints = ({ small, medium }: Breakpoints) => {
 
   const largeMin = newSizes.medium.max + 1;
   newSizes.large = {
-    max: 2000, // TODO: decide upper limit
+    max: largeMin + 1,
```
samijaber, line 90: I would undo the change here to avoid any surprises.

samijaber (approved):
@midhunadarvin LGTM 👍🏽 

PS: Don't forget to follow-up with another PR for gen2 SDKs

packages/react/src/constants/device-sizes.constant.ts
midhunadarvin: done

packages/react/src/components/builder-block.component.tsx
```diff
@@ -181,7 +181,16 @@ export class BuilderBlock extends React.Component<
...
+        this.privateState.context.builderContent?.meta?.breakpoints?.xsmall
+      );
       for (const size of reversedNames) {
+        if (!contentHasXSmallBreakpoint && size === 'xsmall') {
+          // Only apply xsmall styles if xsmall breakpoint is enabled on content
+          // TODO: remove this once xsmall is made GA
```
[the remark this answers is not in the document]
midhunadarvin: done

PR #3858, opened 2025-01-24
fix[angular-gen2]: form block rendering children multiple times

sidmohanty11:
## Description

Jira
https://builder-io.atlassian.net/browse/ENG-8042

_Screenshot_
Tested it out and it works:
Screenshot 2025-01-24 at 6 36 53 PM


sanyamkamat (approved):

kylefowler:
@sidmohanty11 any chance you could explain why this fixes things? It's very unclear why this would make a difference for angular, why it worked for other sdks and why this does impact other sdks.

packages/sdks/src/blocks/form/form/form.lite.tsx
```diff
@@ -320,17 +319,7 @@ export default function FormComponent(props: FormProps) {
...
-            <Block
-              key={`form-block-${idx}`}
-              block={block}
-              context={props.builderContext}
-              registeredComponents={props.builderComponents}
-              linkComponent={props.builderLinkComponent}
-            />
-          )}
-        </For>
+        {props.children}
```
samijaber, line 322: - the wrapping `<Show>` is still using `props.builderBlock.children`. Is that `Show` even needed at this point?
- IIUC: we went from iterating over the `BuilderBlock` array of children and calling `<Block>` on each, to now rendering `props.children`, which is being rendered by the parent component `ComponentRef`: https://github.com/BuilderIO/builder/blob/9f29317c0a4e8b8024630d0a5f66817fb691e9f1/packages/sdks/src/components/block/components/component-ref/component-ref.lite.tsx#L52-L62

So this PR's change is totally fine since it removes redundancy from `Form.lite.tsx` anyways. The part I'm confused by is: why did Angular render twice? Did Angular do a thing where it automatically renders form's children even if there's no `<ng-content>` in `form.ts`? The `For` loop you removed looks exactly like the one that `ComponentRef` has.

samijaber (approved):

packages/sdks/src/blocks/form/form/form.lite.tsx
samijaber: Oh, is it something to do with how `component-ref.ts` works in Angular? Does the following logic:

```ts
@Component({
  selector: "component-ref",
  template: `
    <ng-template #wrapperTemplate>
      <ng-container *ngIf="componentRef">
        <ng-container *ngFor="let child of blockChildren; trackBy: trackByChild0">
          <block
            [block]="child"
            [context]="context"
            [registeredComponents]="registeredComponents"
            [linkComponent]="linkComponent"
          ></block>
        </ng-container>
      </ng-container>
    </ng-template>
   //...
  `,
})

export default class ComponentRef {
  //...
  ngOnInit() {
    //...
    const wrapperTemplate = this.vcRef.createEmbeddedView(this.wrapperTemplateRef);
    wrapperTemplate.detectChanges();
    this.myContent = [wrapperTemplate.rootNodes];
  }

  //...
}
```

Always render the children content? Is that why we had duplicates?

sidmohanty11: yes exactly, if it doesn't get a `<ng-content />` slot it gets projected outside as an outside block (outside of form)

kylefowler: angular novice question... so why do we not need to fix / change how this template works to account for that?

sidmohanty11:
@kylefowler why this happened:
in other frameworks, if the `{props.children}` equivalent is not provided, we don’t render the children content. As a result, this `For` loop is being passed as children even though we never use it:

```tsx
<Show if children>
  <For each={props.blockChildren}>
    {(child) => (
      <Block
        key={child.id}
        block={child}
        context={props.context}
        registeredComponents={props.registeredComponents}
        linkComponent={props.linkComponent}
      />
    )}
  </For>
</Show>
```

However, in Angular, when using dynamic components, the projected template (the code inside ng-template) acts as the props.children equivalent. Whenever it doesn’t find a slot, Angular projects the template outside of interactive-element because the template still has content. At that point, we don’t know whether the slot should be projected or not (i.e., whether the user intends to use props.children), which makes it tricky to handle.

A potential fix is to conditionally render the template by introducing a meta variable for Angular, `shouldRenderChildren`. We could default `shouldRenderChildren` to `true`, but if users explicitly set it to `false`, then we wouldn’t render the children.

```html
<ng-container *ngIf="componentRef && shouldRenderChildren">
        <ng-container *ngFor="let child of blockChildren; trackBy: trackByChild0">
          <block
            [block]="child"
            [context]="context"
            [registeredComponents]="registeredComponents"
            [linkComponent]="linkComponent"
          ></block>
        </ng-container>
      </ng-container>
```

PR #3863, opened 2025-01-24
update Node version

samijaber:
## Description

Next is complaining about needing v18.18+


sidmohanty11 (approved):
will this require a re-install? maybe we can post it in dev-cms to move to this version

PR #3865, opened 2025-01-27
fix[qwik-sdk]: ENG-6695 visual editor hangs if editable area (<Content />) is not in the iframe viewport

yash-builder:
## Description
The `CustomEvent` `"initeditingbldr"` was not triggering as expected by ensuring proper event dispatch timing and listener registration. The `useOnDocument("readystatechange")` handler was updated to dispatch the event only after the DOM is fully loaded (`document.readyState === "complete"`)

**JIRA**:
https://builder-io.atlassian.net/browse/ENG-6695

**Loom**
https://www.loom.com/share/0ab5296adbe24bb2b22319f0588f0bbf?sid=4e4b4066-4b18-459e-bc62-7dae1e45c1f9

### DIFF
--- a/.changeset/nice-planets-roll.md
+++ b/.changeset/nice-planets-roll.md
@@ -2,5 +2,4 @@
 "@builder.io/sdk-qwik": patch
 ---
 
-- Fix: Content editor hangs if not in the iframe viewport
-- Chore: Added test case to trigger event when content component is not in the viewport
+Fix: Content editor hangs if not in the iframe viewport.
--- a/.changeset/nice-planets-roll.md
+++ b/.changeset/nice-planets-roll.md
@@ -1,0 +1,6 @@
+---
+"@builder.io/sdk-qwik": patch
+---
+
+- Fix: Content editor hangs if not in the iframe viewport
+- Chore: Added test case to trigger event when content component is not in the viewport
--- a/packages/sdks/mitosis.config.cjs
+++ b/packages/sdks/mitosis.config.cjs
@@ -107,104 +107,18 @@ const IMPORT_USE_ON_DOCUMENT_PLUGIN = () => ({
       return json;
     },
   },
-});
-
-const REPLACE_USE_ON_WITH_USE_ON_DOCUMENT_PLUGIN = () => ({
   code: {
     post: (code, json) => {
-      if (json.name === 'EnableEditor') {
-        code = code.replace(
-          `
-  useOn(
-    "qvisible",
-    $((event, element) => {
-      if (isBrowser()) {
-        if (isEditing() && !props.isNestedRender) {
-          if (element) {
-            element.dispatchEvent(new CustomEvent("initeditingbldr"));
-          }
-        }
-        const shouldTrackImpression =
-          element.attributes.getNamedItem("shouldTrack")?.value === "true";
-        if (shouldTrackImpression) {
-          const variationId =
-            element.attributes.getNamedItem("variationId")?.value;
-          const contentId = element.attributes.getNamedItem("contentId")?.value;
-          const apiKeyProp = element.attributes.getNamedItem("apiKey")?.value;
-          _track({
-            apiHost: props.apiHost,
-            type: "impression",
-            canTrack: true,
-            contentId,
-            apiKey: apiKeyProp!,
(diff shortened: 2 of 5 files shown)


packages/sdks/mitosis.config.cjs
```diff
@@ -95,6 +95,116 @@ ${restOfCode.join('\n').replace(/<(\/?)Text(.*?)>/g, '<$1BaseText$2>')}
   },
 });
 
+const IMPORT_USE_ON_DOCUMENT_PLUGIN = () => ({
+  json: {
+    post: (json) => {
+      if (json.name !== 'EnableEditor') return;
+      json.imports.push({
+        imports: { useOnDocument: 'useOnDocument' },
```
sidmohanty11, line 103: is it possible to check if there is a way to still use `useOn` with some event that can be called eagerly? If we use `useOnDocument` this will attach the listeners on the document itself and not that div. So we'll be losing all benefits of this approach. If this is a last resort, we could use this plugin: `INJECT_ENABLE_EDITOR_ON_EVENT_HOOKS_PLUGIN`

yash-builder: Hey @sidmohanty11,
I’ve replaced `useOnDocument` with `useOn` for our custom events and attempted to use `useOn("DOMContentLoaded")` instead of `useOnDocument("readystatechange")`. However, it seems that `useOn("DOMContentLoaded")` is not being triggered.

samijaber: @sidmohanty11 I added a brain dump of how editor initialization works here: https://github.com/BuilderIO/builder/pull/3873

essentially, using `useOnDocument` for the "root" event listener of the SDK isn't a problem. The perf benefits are gained by keeping all usage of `props.*` and `state.*` encapsulated within the `initeditingbldr` and `initpreviewingbldr` custom events. This PR does not undo that work.

What it's doing is moving the `useOn('qvisible')` listener to be against another event, such that it fires even if the element is not yet visible according to Qwik, which we need to do for the SDK to start working in the Visual Editor.

PS: `"readystatechange"` ended up working for this, but `useOnDocument` didn't since the latter doesn't hold a reference to the SDK Component element, which it needs to dispatch the subsequent editing/previewing custom events

samijaber: Followup: shockingly, `useOnDocument`'s `element` argument is the host of the current component...which means we can both listen to and send custom events using `useOnDocument`.

This is surprising to me, and goes against my understanding of Qwik's docs, but as long as it works...

CleanShot 2025-01-29 at 16 55 35@2x

packages/sdks/mitosis.config.cjs
```diff
@@ -95,6 +95,33 @@ ${restOfCode.join('\n').replace(/<(\/?)Text(.*?)>/g, '<$1BaseText$2>')}
   },
 });
 
+const IMPORT_USE_ON_DOCUMENT_PLUGIN = () => ({
```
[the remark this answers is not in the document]
yash-builder: Sure, but which approach would you like me to follow? Since `onEvent` uses `useOn` by default and needs to be changed to `useOnDocument`, I understand that the `eventName` will need to be updated to `readystatechange` as well.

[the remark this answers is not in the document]
samijaber: yup, keep your approach as-is, just combine the two plugins to be one plugin. 

I'm suggesting that you 

- update the `eventName` logic here to be `readystatechange`, instead of setting it to `qvisible` and then updating it. 
- keep the rest of your plugin logic (update `useOn` to `useOnDocument`). Copy/paste into this plugin so it's all in one place

.changeset/nice-planets-roll.md
```diff
@@ -0,0 +1,6 @@
+---
+"@builder.io/sdk-qwik": patch
+---
+
+- Fix: Content editor hangs if not in the iframe viewport
+- Chore: Added test case to trigger event when content component is not in the viewport
```
samijaber, line 6: We usually only list user-facing changes in our release notes. Changes to testing and other infra can be ignored, as they do not impact the end-user.

```suggestion
Fix: Content editor hangs if not in the iframe viewport.
```

samijaber (approved):

PR #3868, opened 2025-01-27
release @builder.io/plugin-medusajs 1.0.0 version

seeed-jgomez:
# Builder.io Medusa JS plugin

Easily connect your Medusa Js Commerce API to your Builder.io content!

Medusa Js platform  [https://medusajs.com/](https://medusajs.com/)


samijaber:
Thanks for making this plugin contribution to Builder!

We are revamping our submission process for community plugins to maintain our ecosystem better. As part of that change, we now ask contributors to create their own repos and self-publish their plugins, and then make a submission using the following form: https://forms.gle/hq9ybaBDAdB4dvzy6

I am going to go ahead and close this PR as it is no longer necessary.

PR #3869, opened 2025-01-28
fix[sdk][gen-2] : ENG-8049 support for lazy loading in video element

shubham-builder:
## Description

Added lazy loading for video element using Intersection observer.

_Screenshot_
https://www.loom.com/share/d6abb89423e3486d98246611f7473fc3


packages/sdks/src/blocks/video/video.lite.tsx
```diff
@@ -25,11 +25,54 @@ export default function Video(props: VideoProps) {
...
+  const videoRef = useRef<HTMLVideoElement>(null);
+  let lazyVideoObserver: IntersectionObserver | undefined;
+
+  onMount(() => {
+    if(props.lazyLoad){
+      lazyVideoObserver = new IntersectionObserver(function(entries) {
+        entries.forEach(function(entry) {
+          if (!entry.isIntersecting) return;
+
+          const videoElement = entry.target as HTMLVideoElement;
```
sidmohanty11, line 37: you don't need `.target` when using mitosis element refs, they are directly available,

check: https://github.com/sidmohanty11/builder/blob/feat/variant-container-gen2/packages/sdks/src/blocks/embed/embed.lite.tsx/#L22

https://mitosis.builder.io/docs/hooks/#useref

packages/sdks/src/blocks/video/video.lite.tsx
```diff
@@ -25,11 +25,54 @@ export default function Video(props: VideoProps) {
     },
   });
 
+  const videoRef = useRef<HTMLVideoElement>(null);
+  let lazyVideoObserver: IntersectionObserver | undefined;
```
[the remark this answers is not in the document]
yash-builder: Hey Sid,
Asking out of curiosity.
Since `useStore` is used for state management, could you clarify the need for defining `useRef` inside `useStore`?

Please correct me if I am missing something out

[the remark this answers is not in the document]
yash-builder: Yep I am with you on this for `lazyVideoObserver` @shubham-builder I got confused since `useRef` is also highlighted.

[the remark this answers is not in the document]
sidmohanty11: yes its just for the `lazyVideoObserver`

sidmohanty11 (commented):
can you check why the angular tests are failing?

samijaber (changes requested):
Could we add an e2e test? See https://github.com/BuilderIO/builder/blob/main/packages/sdks/docs/DEVELOP.md for instructions on how to get setup

A good test would do exactly what your loom checks for:

- make a content JSON with a lot of padding above the video, so that it's below the viewport. enable lazyloading for the video.
- have your playwright test check network asset requests
- make sure no video asset request is made initially
- scroll to the bottom
- make sure the video asset request is made

samijaber:
@shubham-builder for the angular test failures: could you run one of the 3 angular e2e test apps locally, build the angular SDK and try to load the `/video` path to see why they're all failing to render the video block now? https://github.com/BuilderIO/builder/blob/main/packages/sdks/docs/DEVELOP.md#run-tests should walk you through how to do that, but tl;dr 

- `cd packages/sdks/e2e/angular-16` (or one of the other 2 angular tests)
- `yarn g:nx build` (to build the angular SDK and the server that's importing it in question)
- `yarn serve` to launch the server

packages/sdks/mitosis.config.cjs
```diff
@@ -864,6 +864,29 @@ const ANGULAR_MARK_SAFE_INNER_HTML = () => ({
...
+
+        code = code.slice(0,startIndex)+ code.slice(endIndex+9);
+
+        const ngAfterViewInitIndex = code.indexOf('this.setAttributes(this.elRef1?.nativeElement, this.node_2_source);')+67;
+        code = code.slice(0,ngAfterViewInitIndex+1)+ codeToMove + code.slice(ngAfterViewInitIndex+1);
+      }
+      return code;
+    },
+  },
+});
```
samijaber, line 888: These hardcoded indexes are too brittle and will quickly backfire.

I dug into our mitosis angular generator logic, and found that we have an undocumented way to move hooks from one place to another:

```suggestion
const ANGULAR_MOVE_DOM_MANIPULATION_CODE_TO_AFTERVIEWINIT = () => ({
  json: {
    post: (json) => {
      if (json.name === 'BuilderVideo') {
        json.compileContext = {
          angular: {
            hooks: {
              ngAfterViewInit: json.hooks.onMount[0],
            },
          },
        };

        delete json.hooks.onMount;
      }
    },
  },
});
```

This should be much cleaner and still accomplish the same task

packages/sdks-tests/src/e2e-tests/blocks.spec.ts
```diff
@@ -313,6 +313,41 @@ test.describe('Blocks', () => {
...
+      const videoLocator = page.locator('video'); 
+
+      await expect(videoLocator).not.toBeInViewport();
+      expect(videoRequestMade).toBeFalsy();
+
+      await videoLocator.scrollIntoViewIfNeeded()
+      await page.waitForTimeout(1000); // allow time for the network request
+
+      await expect(videoLocator).toBeInViewport();
+      expect(videoRequestMade).toBeTruthy();
```
samijaber, line 349: we want to avoid using `waitForTimeout` as much as possible. It should never be needed unless under extremely unusual edge-cases.

Adding such timeouts will make our tests unreliable and flaky.

In this case, you want to scroll and then wait for the video request to be made. A better way to do that and not need a timeout is to use `page.waitForResponse()`:

```suggestion
      const requestPromise = page.waitForRequest(request => request.url().includes(VIDEO_CDN_URL))
      await videoLocator.scrollIntoViewIfNeeded()
      await page.waitForTimeout(1000); // allow time for the network request

      await expect(videoLocator).toBeInViewport();
      await requestPromise;
```

The `await requestPromise` will wait until (exactly) one request is made that matches `VIDEO_CDN_URL`, or will throw otherwise. It will take care of waiting the appropriate time so you don't have to

packages/sdks/src/blocks/video/video.lite.tsx
```diff
@@ -45,11 +89,10 @@ export default function Video(props: VideoProps) {
               }
             : null),
         }}
-        src={props.video || 'no-src'}
         poster={props.posterImage}
         class="builder-video"
       >
-        {!props.lazyLoad && <source type="video/mp4" src={props.video} />}
+        <source type="video/mp4" {...props.lazyLoad ? { 'data-src': props.video } : { src: props.video }} />
```
sidmohanty11, line 95: props.lazyLoad is a boolean right? Did you mean:
```suggestion
        <source type="video/mp4" {...(props.lazyLoad ? { 'data-src': props.video } : { src: props.video })} />
```

packages/sdks-tests/src/e2e-tests/blocks.spec.ts
```diff
@@ -313,6 +313,42 @@ test.describe('Blocks', () => {
...
+
+      await page.goto('/video-lazy-load');
+      const videoLocator = page.locator('video'); 
+
+      await expect(videoLocator).not.toBeInViewport();
+      expect(videoRequestMade).toBeFalsy();
+
+      const requestPromise = page.waitForRequest(request => request.url().includes(VIDEO_CDN_URL))
+      await videoLocator.scrollIntoViewIfNeeded()
+      await page.waitForTimeout(1000); // allow time for the network request
```
sidmohanty11, line 347: is this required? we should avoid using `waitForTimeout` wherever possible. We are waiting for the request from `requestPromise` right? I think this can be removed

sidmohanty11 (approved):
LGTM! 🚀

samijaber (approved):
🎉

PR #3871, opened 2025-01-28
Update Bynder plugin README.md with installation steps

mix3d:
(Now that we published the plugin)

## Description

We're submitting the plugin to Bynder's marketplace, and without a docs page for it (may not happen), we want to point to the Github readme. It was missing installation instructions post-npm-publish.


gustavocodess (approved):
Looks good, thanks Madison

PR #3872, opened 2025-01-29
fix[angular-gen2+solid]: allow center aligning blocks inside a column

sidmohanty11:
## Description

`flexGrow: 1` was passed as an object so we were not able to use setAttribute, now we are checking if typeof style is object then we are attaching styles using `this.renderer.setStyle` (this will auto handle case sensitivity, for ex: if passed flexGrow -> it will convert to flex-grow) else (in case of string) we are using the setAttribute method

For Solid,
it needed 'flex-grow' not flexGrow: https://docs.solidjs.com/reference/jsx-attributes/style

Jira
https://builder-io.atlassian.net/browse/ENG-7929

Have tested the function in the [angular playground](https://angular.dev/playground), code snippet:
```ts
import {Component,Renderer2,ViewChild,ElementRef} from '@angular/core';
import {bootstrapApplication} from '@angular/platform-browser';

@Component({
  selector: 'app-root',
  template: `
    <div #v>
Hello world!
    </div>
  `,
})
export class PlaygroundComponent {
    @ViewChild('v', { read: ElementRef })
  v!: ElementRef;
  constructor(private renderer: Renderer2) {}
  
   private handleStyleProp(el: any, style: any) {
    if (typeof style === 'object') {
      Object.entries(style).forEach(([key, value]) => {
        this.renderer.setStyle(el, key, value);
      });
    } else {
      this.renderer.setAttribute(el, 'style', style)
    }
  }

  ngAfterViewInit() {
    const el = this.v?.nativeElement;
    if (!el) {
      return;
    }
    this.handleStyleProp(el, {backgroundColor: 'red', flexGrow: 1})
    //this.handleStyleProp(el, 'color: blue;')
    }
  
}

bootstrapApplication(PlaygroundComponent);
```

_Screenshot_


packages/sdks/src/blocks/columns/columns.lite.tsx
```diff
@@ -41,6 +41,19 @@ export default function Columns(props: ColumnProps) {
     get stackAt() {
       return props.stackColumnsAt || 'tablet';
     },
+    get blocksStyle() {
+      return useTarget({
+        solid: {
+          'flex-grow': '1',
+        },
```
sidmohanty11, line 48: @samijaber I've made this as a state variable, but it looks like it should be handled in mitosis.

if style prop is provided as a state variable or object, we can wrap it up inside `convertCamelToKebab(styles)`

samijaber: Yeah, could be a good mitosis improvement to have ootb.

To avoid unnecessary state, could we inline this `useTarget` in `styleProp={useTarget(/**/)}`?

sidmohanty11:
Vercel deployment was failing because https://builder.io/content/34dba31c5147402287b09f21e58cbf57/edit content did not return us a URL path. Have updated the condition to use urlPath "is" containers. Thanks to @rima-builder and @sanyamkamat for helping find this.

PR #3879, opened 2025-01-30
fix[gen1][core]: ENG-7585 ensure canTrack has the right value set

clyde-builderio:
## Description

builderSessionId still set in cookies after setting canTrack to false

**JIRA Ticket**
https://builder-io.atlassian.net/browse/ENG-7585

**Loom**
https://www.loom.com/share/5b9b291177784255ac770c57d932151c

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -1294,6 +1294,7 @@ export class Builder {
   }
 
   set canTrack(canTrack) {
+    this.hasOverriddenCanTrack = true;
     if (this.canTrack !== canTrack) {
       this.canTrack$.next(canTrack);
     }
@@ -1326,6 +1327,7 @@ export class Builder {
   private apiEndpoint$ = new BehaviorSubject<'content' | 'query'>('query');
   private apiVersion$ = new BehaviorSubject<ApiVersion | undefined>(undefined);
   private canTrack$ = new BehaviorSubject(!this.browserTrackingDisabled);
+  private hasOverriddenCanTrack = false;
   private apiKey$ = new BehaviorSubject<string | null>(null);
   private authToken$ = new BehaviorSubject<string | null>(null);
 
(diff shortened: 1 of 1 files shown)


packages/core/src/builder.class.ts
```diff
@@ -2119,7 +2121,9 @@ export class Builder {
     if (res) {
       this.response = res;
     }
-    this.canTrack = canTrack;
+    if (!this.hasOverriddenCanTrack) {
+      this.canTrack = canTrack;
```
samijaber, line 2125: Ah, good catch! `this.defaultCanTrack` is overriding `builder.canTrack()` calls made before `builder.init()`.

Could you add a comment explaining that here? Something like:

> If `builder.canTrack` was set before the `builder.init()` call, we want to preserve the user's choice and not override it with a default.

samijaber (changes requested):
Can you add an e2e test to catch this?

Right now we test like this: 

https://github.com/BuilderIO/builder/blob/81ea5d997e54e0bc516ef51076aaf1df40657478/packages/react-tests/next14-pages/pages/%5B%5B...index%5D%5D.tsx#L36-L43

Note the `builder.canTrack = true` is made _after_ `builder.init()`. 


Could you add another test pathname, `/can-track-false-pre-init`, which will set `builder.canTrack = true` before `builder.init()`? Then you can reuse the tests for `/can-track-false` from: https://github.com/BuilderIO/builder/blob/81ea5d997e54e0bc516ef51076aaf1df40657478/packages/sdks-tests/src/e2e-tests/tracking.spec.ts#L14-L43

but pointing to this new path

.changeset/thin-doors-pull.md
```diff
@@ -0,0 +1,6 @@
+---
+'@builder.io/sdk': patch
+'@builder.io/react': patch
+---
+
+Fix: The value of `canTrack` that is set before `builder.init()` will ensure that `builderSessionId` is not saved
```
samijaber, line 6:
```suggestion
Fix: ability to set `builder.canTrack` before calling `builder.init()`
```

samijaber (approved):

PR #3883, opened 2025-01-31
update to production react 19

kylefowler:
## Description
Bump the peer dependency to the production react 19


samijaber:
As discussed in https://builder-internal.slack.com/archives/C07ARKMLD0S/p1738279699533949 , there is no technical issue at play here, merely educating the customer that we are trying to support as many React versions as possible.

Safe to close this as we want to keep `v19.0.0-rc` support

PR #3891, opened 2025-02-04
Vue SDK: stop passing down unused props

samijaber:
## Description

https://builder-io.atlassian.net/browse/ENG-7615


packages/sdks/mitosis.config.cjs
```diff
@@ -773,8 +773,8 @@ const ANGULAR_NOWRAP_INTERACTIVE_ELEMENT_PLUGIN = () => ({
           `
         );
         code = code.replaceAll(
-          'attributes: this.attributes,',
-          '...(this.hasAttributesInput(this.Wrapper) ? { attributes: this.attributes } : {})'
+          'attributes: this.wrappedPropsWithAttributes.attributes,',
+          '...(this.hasAttributesInput(this.Wrapper) ? { attributes: this.wrappedPropsWithAttributes.attributes } : {})'
```
samijaber, line 777: @sidmohanty11 I think I'm breaking this code with this PR. My change here is fine but the `ngOnChanges` overrides below in this same plugin will stop working:


```ts
ngOnChanges(changes: SimpleChanges) { if (changes["attributes"] && !this.hasAttributesInput(this.Wrapper)) { this.ngAfterViewInit(); }
```

what do you suggest we do about this? Also, is there anything else that will start to break with this approach?

if it's too hard to make the angular SDK happy with not having an `attributes` prop, i can look into bringing it back in this plugin instead

sidmohanty11: @samijaber breaking up attributes will mean that we'll lose access to builder-id, style etc so now we'll have to attach all props as HTML attributes and not pass them to be reactive (as inputs). Is it possible to keep the attributes key? because here:
https://github.com/BuilderIO/builder/blob/f9b85528015839287fbf092470bc57c38344d329/packages/sdks/mitosis.config.cjs#L723-L731
we check if that element doesn't have `attributes` as an input then we spread it out in the element

sidmohanty11: else we can look for only these four: id, style, href and className as attributes and send everything else as inputs

sidmohanty11 (commented):
@samijaber this needs a changeset

as the accordion VE is actually failing: https://www.loom.com/share/036fda4959c9473aa0604299097b7a64 we can skip it for now

sidmohanty11 (approved):
🚀

PR #3892, opened 2025-02-04
fix: ENG-8176 fallback UI for unknown elements

shubham-builder:
## Description

Added fallback UI for unknown element in angular dynamic renderer

_Screenshot_
LOOM - https://www.loom.com/share/c65be89e96194d988c1cd015a3047f70


packages/sdks-tests/src/e2e-tests/duplicate-attributes.spec.ts
```diff
@@ -3,7 +3,7 @@ import { excludeTestFor, test } from '../helpers/index.js';
 
 test.describe('Duplicate Attributes', () => {
   test('wrapped block has no duplicate attributes', async ({ page, packageName, sdk }) => {
-    test.fail(
+    test.expect(
```
samijaber, line 6: I don't understand this change. `test.fail(condition)` means that the test is _guaranteed to fail_ when `condition` is `true`, but `test.expect()` is used for making test assertions.

I think in this case, what you want to do is remove this `test.fail()` command altogether. No need for a replacement

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
```diff
@@ -202,6 +204,49 @@ export default class DynamicRenderer {
...
+    private _element!: HTMLElement;
+    private _listenerFns = new Map<string, () => void>();
+  
+    constructor(private hostRef: ElementRef, private renderer: Renderer2) {}
+  
+    ngOnInit() {
+      this._element = this.renderer.createElement(this.tagName);
+      this.renderer.appendChild(this.hostRef.nativeElement, this._element);
+      this.setAttributes(this._element, this.attributes);
+    }
```
samijaber, line 228: @sidmohanty11 can you remind me: did you attempt this before we went down the route of explicitly generating mappings for each known HTML element?

if this fallback works, then why can't we use it for all known HTML elements too?

samijaber (commented):
@shubham-builder can you add a test case with an example of such an "unknown element" so we can confirm that this PR fixes it?

packages/sdks-tests/src/e2e-tests/dynamic-unknown-element.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+
+test.describe('Unknown Element', () => {
+  test.only('unknown element is rendered', async ({ page, packageName }) => {
+    test.skip(packageName === 'react-native');
+    await page.goto('/dynamic-unknown-element');
+
+    const testTagName = await page.locator('test');
+
+    const testTagNameId = await testTagName?.getAttribute('builder-id', { timeout: 10000 });
+    const testTagNameClass = await testTagName?.getAttribute('class', { timeout: 10000 });
```
samijaber, line 12: As per https://playwright.dev/docs/api/class-locator#locator-get-attribute

> If you need to assert an element's attribute, prefer [expect(locator).toHaveAttribute()](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-have-attribute) to avoid flakiness

Also, unless there's a very specific reason, we want to avoid providing manual `timeout` values to any of our tests. Playwright usually takes care of waiting for things to happen. Having to provide these values manually is a sign that something is wrong.

packages/sdks-tests/src/e2e-tests/dynamic-unknown-element.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+
+    expect(testTagNameId).toBe('builder-493100cea9504d56886045462b65b481');
+    expect(testTagNameClass?.includes('builder-493100cea9504d56886045462b65b481')).toBe(true);
+
+  });
```
samijaber, line 17: can we also add some checks that the child of the unknown element renders correctly? in this case there should be a text block inside of `test` with content `111111`

samijaber (commented):
This LGTM (once the integration test is improved), but I'll let @sidmohanty11 give it a ✅  as well since he's very familiar with this code.

packages/sdks-tests/src/e2e-tests/dynamic-unknown-element.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { expect } from '@playwright/test';
+import { test } from '../helpers/index.js';
+
+test.describe('Unknown Element', () => {
+  test('unknown element is rendered', async ({ page, packageName }) => {
+    test.skip(packageName === 'react-native');
```
samijaber, line 6:
```suggestion
  test('unknown element is rendered', async ({ page, sdk }) => {
    test.skip(sdk === 'react-native');
```

https://github.com/BuilderIO/builder/blob/1396fb4cc5ff9db2369f0cff12f42fbb910071c9/packages/sdks-tests/src/helpers/sdk.ts#L2-L40

`sdk` is `react-native`

`packageName` is `react-native-74` or `react-native-76-fabric` (the specific server that is running the SDK)

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
```diff
@@ -202,6 +209,53 @@ export default class DynamicRenderer {
...
+      constructor(private hostRef: ElementRef, private renderer: Renderer2) {}
+    
+     ngOnInit() {
+      this._element = this.renderer.createElement(this.tagName);
+      this.renderer.appendChild(this.hostRef.nativeElement, this._element);
+      this.setAttributes(this._element, this.attributes);
+    }
+
+    ngAfterViewInit(){
+      this.renderer.appendChild(this.hostRef.nativeElement.children[1], this.hostRef.nativeElement.children[0]);
```
sidmohanty11, line 236: how does this work? Could you please explain? Can we somehow do this inside ngOnInit? I'm afraid if we'll lose SSRing children with this

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
```diff
@@ -190,6 +195,8 @@ export default class DynamicRenderer {
       switch (this.TagName) {
         ${htmlElements.map((el) => `case '${el}': this.TagName = Dynamic${capitalize(el)}; break;`).join('\n        ')}
```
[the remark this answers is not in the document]
sidmohanty11: This should also fix Angular 16 tests throwing Input annotation errors saying TagName is not defined in other dynamic components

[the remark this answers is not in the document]
shubham-builder: That would mean using the dynamic-unknown-element for all the known elements as well
So in this case even dynamic-p will render as dynamic-unkonwn-element unless it's code is updated accordingly with support of void elements.
Do we want that in same pr or I think we can take it as separate task to eliminate this switch case and use single function to handle known and unknown elements.

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
shubham-builder: Parent component populates the <ng-content></ng-content> and the <dynamic-unknown-element> is created inside the ngOnInit (which is necessary as input are initialised there). This changes the order in which the elements  should be inserted.
So once the view is created this fixes the order of elements( moving the element to be child of unknown element)
Unfortunately it will be issue for SSR rendering... I am looking into that part

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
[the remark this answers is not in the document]
sidmohanty11: IMO we should remove these hard-coded dynamic elements from dynamic renderer and use this `dynamic-unknown-element` to render dynamic elements. We could rename this to `dynamic-element` also if it makes more sense or even better we could refactor it so that we handle this inside the dynamic renderer itself (because that's the role of dynamic-renderer) but I think for now removing the switch and handling all the elements inside dynamic-element will be nice. What do you think?

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
sidmohanty11: what happens when we have multiple children?

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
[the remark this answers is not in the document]
shubham-builder: That totally makes sense

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
shubham-builder: @sidmohanty11 It will work even in case of muliple children since each children will have it's own dynamic-renderer and this append works within the dynamic-renderer element

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
```diff
@@ -174,22 +180,31 @@ export default class DynamicRenderer {
...
+  className: string = '';
   myContent?: any[][];
 
   useTypeOf(obj: any): string {
     return typeof obj;
   }
 
+  getClassName(): string {
+    return this.className;
+  }
```
sidmohanty11, line 196: as we're now separating known and unknown elements can we remove the className checks and just stick to the switch case?

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
```diff
@@ -198,9 +213,60 @@ export default class DynamicRenderer {
...
+        if(this.tagName){
+          this._element = this.renderer.createElement(this.tagName);
+          this.renderer.appendChild(this.hostRef.nativeElement, this._element);
+          this.setAttributes(this._element, this.attributes);
+        }
+      }
+
+      ngAfterViewInit(){
+        if(this.hostRef.nativeElement.children.length > 1){
+          this.renderer.appendChild(this.hostRef.nativeElement.children[1], this.hostRef.nativeElement.children[0]);
```
sidmohanty11, line 247: note for future: children of unknown dynamic elements will not be ssred

sidmohanty11 (commented):
Overall LGTM, just some minor comments then we're good to go

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
shubham-builder: @sidmohanty11 this check is not for the known and unknown element. It's still required. It's for the tagname input for the child element.

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
shubham-builder: this is on client side

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
sidmohanty11: local tagName should only be used in case of default case right? so we can just check if tagName is present -> it's a dynamic element no?

sidmohanty11 (approved):
LGTM!

PR #3896, opened 2025-02-07
fix: ensure correct variation tracking for impressions

yash-builder:
#### **What Changed?**  
- Fixed an issue where **two impressions** were being tracked—one for the default content and another for the variation.  
- Updated the logic to ensure that **only the variation impression is tracked**, preventing duplicate tracking.  
- Added a **Playwright test** to verify that impressions are logged correctly with the right `variationId`.  

#### **Why This Change?**  
- Previously, the analytics data incorrectly counted extra impressions for the default content.  
- Now, only the actual **winning variation** sends an impression event, ensuring accurate A/B testing results.  

_Screenshot_
**Before**: _Impression for variation getting send without variationId_
Screenshot 2025-02-07 at 4 29 48 PM

**After**: _Sending only winning variation_
Screenshot 2025-02-07 at 4 32 39 PM

Screenshot 2025-02-07 at 4 33 22 PM


samijaber (changes requested):
The way our A/B testing works is that it renders `<Content>` for each variant:

https://github.com/BuilderIO/builder/blob/79e188874b68695f27878c88aef0cf8443106183/packages/sdks/src/components/content-variants/content-variants.lite.tsx#L109-L174

And then each `<Content>` will render `<EnableEditor>`.

IIUC, with your change, each of those `<EnableEditor>` is still making a separate impression call but they are at least now using the winning ID each time. 

But we're still having duplicate impressions (one per `EnableEditor`), and I don't think we want that.

so I think you should make some additional changes to your PR:

- update `if (shouldTrackImpression)` to `if (shouldTrackImpression && variationId === winningVariantId)`. This should mean that we  _only_ call `/track` when the current variationId is the winning one, bringing the impression calls down to one.
- update the `/track` tests you added to `ab-test.spec.ts` to fail if more than 1 `/track` call is made. This way we guarantee the above

packages/sdks-tests/src/e2e-tests/ab-test.spec.ts
```diff
@@ -104,7 +103,30 @@ test.describe('A/B tests', () => {
...
+            if (payload.events[0].data.contentId !== CONTENT_ID) {
+              throw new Error('ContentId does not match expected default contentId');
+            }
+          }
+          await route.continue();
+        });
+
+        await page.goto('/ab-test', { waitUntil: 'networkidle' });
+
+        if (trackCalls > 0) {
```
samijaber, line 127: Why do we have this check? This means that if our tracking is completely broken and never sends any impressions, then we will always pass this test. That's not ideal.

packages/sdks/mitosis.config.cjs
```diff
@@ -1180,11 +1180,18 @@ module.exports = {
           code: {
             post: (code, json) => {
               if (json.name === 'EnableEditor') {
+                code = code.replace('const elementRef = useSignal<Element>();', 'const elementRef = useSignal<Element>();\nconst hasExecuted = useSignal(false);');
```
samijaber, line 1183: This is going to be a very brittle plugin to maintain. I would add the `hasExecuted` state to `enable-editor.lite.tsx`. It won't negatively impact the other SDKs to have this one extra state in there.

if we want, we can use `useTarget()` so that we only get and set this state for Qwik since it's unnecessary for the others

packages/sdks-tests/src/e2e-tests/ab-test.spec.ts
yash-builder: Based on this comment. That the `/track` call does actually fire _*(your test will pass even if no `/track` calls are made)*_

I added this line to pass the test even if no `/track` calls are made. Maybe I misunderstood it

samijaber (commented):
Awesome. your pr is missing the changesets still

.changeset/moody-rocks-mix.md
```diff
@@ -0,0 +1,11 @@
...
+"@builder.io/sdk-angular": patch
+"@builder.io/sdk-react-nextjs": patch
+"@builder.io/sdk-qwik": patch
+"@builder.io/sdk-react": patch
+"@builder.io/sdk-react-native": patch
+"@builder.io/sdk-solid": patch
+"@builder.io/sdk-svelte": patch
+---
+
+Fix: Ensure `/track` call validation handles default and variant scenarios correctly.
```
samijaber, line 11:
```suggestion
Fix: duplicate `/track` API calls made for losing A/B test variants.
```

samijaber (commented):
I would also add another changeset specifically for the Qwik SDK fix you made:

> Fix: duplicate `/track` impression calls.

because it's a separate bugfix to the one about A/B test variants

samijaber (approved):
🎉

PR #3898, opened 2025-02-10
feat[angular-gen2]: A/B tests support

sidmohanty11:
## Description

Jira
https://builder-io.atlassian.net/browse/ENG-8217
https://builder-io.atlassian.net/browse/ENG-8383

Looms
https://www.loom.com/share/93a87a08e48f468180f9110bb434aa4c
https://www.loom.com/share/035299b7d96149b78f2f31bb42be7804

### DIFF
--- a/packages/sdks-tests/src/e2e-tests/editing.spec.ts
+++ b/packages/sdks-tests/src/e2e-tests/editing.spec.ts
@@ -6,7 +6,7 @@ import {
 } from '../specs/columns.js';
 import { NEW_TEXT } from '../specs/helpers.js';
 import { HOMEPAGE } from '../specs/homepage.js';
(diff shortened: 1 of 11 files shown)


packages/sdks-tests/src/e2e-tests/editing.spec.ts
```diff
@@ -150,6 +150,33 @@ test.describe('Visual Editing', () => {
     }
   });
 
+  test('removal of styles should work properly', async ({ page, packageName, sdk, basePort }) => {
```
sidmohanty11, line 153: ran the test on main branch as well and it was failing so we can be sure that styles now work as expected,

Screenshot 2025-02-18 at 7 37 48 PM

samijaber (approved):
🎉

PR #3907, opened 2025-02-12
Live Previewing Data Model – Add Support for Latest subscribeToEditor() Change in React Gen2 Snippets

uttej-vsk:
## Description

As per the ticket [EDU-459](https://builder-io.atlassian.net/browse/EDU-459), this PR updates the React Gen2 code snippet to align with the latest breaking change to `subscribeToEditor()`. This ensures that the docs remain up to date and work without any issues. 

ENG changelog: https://github.com/BuilderIO/builder/blob/main/packages/sdks/output/react/CHANGELOG.md#400 

This PR adds the code for live previewing data models in the code snippets. I just wanted to let you know that more to come. 
@samijaber Please review it.

[EDU-459]: https://builder-io.atlassian.net/browse/EDU-459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/sdks-tests/src/snippet-tests/live-previewing.spec.ts
+++ b/packages/sdks-tests/src/snippet-tests/live-previewing.spec.ts
@@ -22,7 +22,9 @@ test.describe('LivePreviewBlogData Component', () => {
     await expect(blogPreview).toContainText('Authored by: John Doe');
     await expect(blogPreview).toContainText('Handle: john_doe');
 
-    const expectedDate = new Date().toDateString();
+    const expectedDate = new Date().toDateString().split(' ').slice(0, 4).join(' ');
+
+    console.log('expectedDate', expectedDate);
 
     const actualText = await blogPreview.textContent();
 
--- a/packages/sdks-tests/src/snippet-tests/live-previewing.spec.ts
+++ b/packages/sdks-tests/src/snippet-tests/live-previewing.spec.ts
@@ -14,7 +14,6 @@ test.describe('LivePreviewBlogData Component', () => {
     test.skip(!['react'].includes(packageName));
 
     await page.goto('/live-preview');
-    await page.waitForLoadState('networkidle');
 
     const blogPreview = page.locator('.blog-data-preview');
     await expect(blogPreview).toBeVisible();
@@ -37,9 +36,8 @@ test.describe('LivePreviewBlogData Component', () => {
       );
 
       await launchEmbedderAndWaitForSdk({ path: '/live-preview', basePort, page, sdk });
-      await page.waitForLoadState('networkidle');
 
-      const initialContent = {
+      const INITIAL_CONTENT = {
         data: {
           title: 'Welcome to Builder.io',
           author: 'John Doe',
@@ -48,18 +46,18 @@ test.describe('LivePreviewBlogData Component', () => {
         },
       };
       await expect(
-        page.frameLocator('iframe').getByText(`Blog Title: ${initialContent.data.title}`)
+        page.frameLocator('iframe').getByText(`Blog Title: ${INITIAL_CONTENT.data.title}`)
       ).toBeVisible();
       await expect(
-        page.frameLocator('iframe').getByText(`Authored by: ${initialContent.data.author}`)
+        page.frameLocator('iframe').getByText(`Authored by: ${INITIAL_CONTENT.data.author}`)
       ).toBeVisible();
       await expect(
-        page.frameLocator('iframe').getByText(`Handle: ${initialContent.data.handle}`)
+        page.frameLocator('iframe').getByText(`Handle: ${INITIAL_CONTENT.data.handle}`)
       ).toBeVisible();
       await expect(
         page
           .frameLocator('iframe')
-          .getByText(`Published date: ${initialContent.data.publishedDate}`)
+          .getByText(`Published date: ${INITIAL_CONTENT.data.publishedDate}`)
       ).toBeVisible();
 
       const UPDATED_CONTENT = {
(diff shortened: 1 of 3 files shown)


packages/sdks-tests/src/snippet-tests/live-previewing.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+    const blogPreview = page.locator('.blog-data-preview');
+    await expect(blogPreview).toBeVisible();
+
+    //assert the blog details coming from builder data model
+    await expect(blogPreview).toContainText('Blog Title: Welcome to Builder.io');
+    await expect(blogPreview).toContainText('Authored by: John Doe');
+    await expect(blogPreview).toContainText('Handle: john_doe');
+    await expect(blogPreview).toContainText('Published date: Tue Feb 11 2025');
+  });
+});
```
samijaber, line 27: Since this is snippet for live previewing, we want to make sure that editing the content will work.

Can you try to replicate this Visual Editor snippet test we already have for integrating Pages?

https://github.com/BuilderIO/builder/blob/1396fb4cc5ff9db2369f0cff12f42fbb910071c9/packages/sdks-tests/src/snippet-tests/integrating-pages.spec.ts#L46-L97

this code uses `launchEmbeddedAndWaitForSdk` to launch the test, and then `sendPatchOrUpdateMessage` (instead of `sendContentUpdateMessage` so that react gen1 SDK is also supported in the future) to send the new versions of the data model to the SDK

uttej-vsk: Yeah, I was expecting this type of behavior from the test cases. I tried a few things to dynamically assert to mock the behavior which didn't work and i didn't know we had one already. Thanks for pointing me towards that.

uttej-vsk: @samijaber Added the test cases. Can you please review it?

packages/sdks-tests/src/snippet-tests/live-previewing.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+      const NEW_CONTENT = {
+        data: {
+          title: 'Welcome to Builder.io',
+          author: 'John Doe',
+          handle: 'john_doe',
+          publishedDate: 'Tue Feb 11 2025',
+        },
+      };
+
+      await sendPatchOrUpdateMessage({
```
samijaber, line 52: Can we add `expect()` checks for the values _before_ this call is made? to make sure the values are actually changing from A to B.

packages/sdks-tests/src/snippet-tests/live-previewing.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    test.skip(!['react'].includes(packageName));
+
+    await page.goto('/live-preview');
+    await page.waitForLoadState('networkidle');
+
+    const blogPreview = page.locator('.blog-data-preview');
+    await expect(blogPreview).toBeVisible();
+
+    //assert the blog details coming from builder data model
+    await expect(blogPreview).toContainText('Blog Title: Welcome to Visual Copilot');
```
samijaber, line 23: this test seems to be failing

uttej-vsk: Screenshot 2025-02-13 at 2 56 43 PM

@samijaber All tests are passing in my local. Am i doing anything incorrectly?

samijaber: oh, false warning. My Github UI showed me a failure for some reason, might have been from one of your old commits

packages/sdks-tests/src/snippet-tests/live-previewing.spec.ts
uttej-vsk: @samijaber I have added this assertion.

packages/sdks-tests/src/snippet-tests/live-previewing.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    const response = await page.goto('/live-preview');
+    expect(response?.status()).toBeLessThan(400);
+  });
+
+  test('should display blog details correctly', async ({ page, packageName }) => {
+    test.skip(!['react'].includes(packageName));
+
+    await page.goto('/live-preview');
+    await page.waitForLoadState('networkidle');
```
samijaber, line 17: We shouldn't need this check. https://playwright.dev/docs/api/class-page#page-wait-for-load-state

> Most of the time, this method is not needed because Playwright [auto-waits before every action](https://playwright.dev/docs/actionability).

```suggestion
```

Any particular reason you needed it?

packages/sdks-tests/src/snippet-tests/live-previewing.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+      test.skip(
+        packageName === 'nextjs-sdk-next-app' ||
+          packageName === 'gen1-next14-pages' ||
+          packageName === 'gen1-remix' ||
+          !['react'].includes(packageName),
+        'Skipping test: incompatible package or framework.'
+      );
+
+      await launchEmbedderAndWaitForSdk({ path: '/live-preview', basePort, page, sdk });
+      await page.waitForLoadState('networkidle');
```
samijaber, line 40: We shouldn't need this check. https://playwright.dev/docs/api/class-page#page-wait-for-load-state

> Most of the time, this method is not needed because Playwright [auto-waits before every action](https://playwright.dev/docs/actionability).

```suggestion
```

Any particular reason you needed it?

samijaber (approved):
Awesome, Thanks @uttej-vsk !

packages/sdks-tests/src/snippet-tests/live-previewing.spec.ts
uttej-vsk: Oh, I used it for the very same reason. It looks like I don't need it anymore. I will remove it.

PR #3911, opened 2025-02-12
[shopify][Plugin]: Increasing limit for storefront shopify products api

anaghav2023:
## Description
This api is by default having a limit of 20 products, increasing it to a maximum of 250.

### DIFF
--- a/plugins/shopify/src/plugin.ts
+++ b/plugins/shopify/src/plugin.ts
@@ -86,6 +86,7 @@ registerCommercePlugin(
           return client.collection.fetchQuery({
             query: search ? `title:*${search}*` : '',
             sortKey: 'TITLE',
+            first: 250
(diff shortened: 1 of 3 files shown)


sanyamkamat (approved):

plugins/shopify/src/plugin.ts
```diff
@@ -51,6 +51,7 @@ registerCommercePlugin(
             (await client.product.fetchQuery({
               query: search ? `title:*${search}*` : '',
               sortKey: 'TITLE',
+              first: 250
```
sanyamkamat, line 54: Do we need this for `collection` too?

anaghav2023: Might need in future. Added to collection as well

PR #3916, opened 2025-02-17
fix[plugin][algolia]: ENG-8338 fix index and private key checks in algolia plugin

sanyamkamat:
## Description

Return early if user decline permanent sync. Also, correctly override Algolia plugin configuration


midhunadarvin (commented):
LGTM

sidmohanty11 (approved):

PR #3915, opened 2025-02-15
EDU-465: Angular live previewing data models

uttej-vsk:
## Description

This is similar to a previous PR on React that focused on live previewing the data model.

@samijaber Whenever you are back on Tuesday, please review this.

PR #3919, opened 2025-02-18
Vue and Nuxt - live previewing data models

uttej-vsk:
## Description

This is similar to previous other PRs on adding live previewing data models that focused on live previewing the data model. It has Nuxt and Vue 3 examples for Live preview.

@samijaber One more today and a few more to come; please review it :D

### DIFF
--- a/packages/sdks/snippets/nuxt/pages/live-preview/[...app].vue
+++ b/packages/sdks/snippets/nuxt/pages/live-preview/[...app].vue
@@ -1,69 +1,41 @@
 <script setup lang="ts">
-import { ref, computed, watch, onMounted, onBeforeUnmount } from 'vue'
+import { computed, watchEffect, onMounted, onUnmounted } from 'vue'
 import { useAsyncData, useRoute } from '#app'
 import { fetchOneEntry, subscribeToEditor } from '@builder.io/sdk-vue'
 
-const route = useRoute()
+const slug = computed(() => useRoute().path)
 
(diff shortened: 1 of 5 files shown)


uttej-vsk:
> If you could go over the feedback in #3915: I believe all of it can be applied here as well.

I tried doing it for Vue as well. Along with angular.

packages/sdks/snippets/vue/src/views/LivePreview.vue
```diff
@@ -0,0 +1,71 @@
...
+import { fetchOneEntry, subscribeToEditor } from '@builder.io/sdk-vue'
+
+const content = ref<any>(null)
+const loading = ref(true)
+
+const route = useRoute()
+
+function getSlug() {
+  return route.path
+}
```
samijaber, line 24: can we remove this helper? it's only used once and adds noise. did you mean to make it a `computed()` and use it in `watch()` like in the nuxt snippet?

packages/sdks/snippets/nuxt/pages/live-preview/[...app].vue
```diff
@@ -0,0 +1,69 @@
...
+    },
+  })
+)
+
+watch(slug, () => {
+  refresh()
+})
+
+const unsubscribe = ref<() => void>()
+if (process.client) {
```
Comment (samijaber) [inline packages/sdks/snippets/nuxt/pages/live-preview/[...app].vue:30]: i dont think this check is needed.

samijaber (approved):

PR #3923, opened 2025-02-20
feat[TextEditor]: Handling <p> and <br> tags in text editor

anaghav2023:
## Description

Issue: https://www.loom.com/share/778be6445389441db7188e83bdc752cb

**Current behavior:**
Pressing Enter creates a line break (<br> tag) without spacing (aka without creating new <p> tags)
Shift+Enter has no different effect than Enter

**Expected behavior:**
Pressing Enter should create a new paragraph (P tag) which will give it the expected amount of spacing
Shift+Enter should create a line break (BR tag) without spacing

**Solution Loom**
https://www.loom.com/share/a29637870e84432daa4493af6bed6181

### DIFF
--- a/packages/react/src/blocks/Text.tsx
+++ b/packages/react/src/blocks/Text.tsx
@@ -22,7 +22,38 @@ class TextComponent extends React.Component<TextProps> {
     if (this.textRef && !/{{([^}]+)}}/.test(this.props.text)) {
       this.textRef.innerHTML = this.props.text;
     }
+
+    if (this.textRef) {
+      this.textRef.addEventListener('keydown', this.handleKeyDown);
+    }
   }
+
+  handleKeyDown = (event: KeyboardEvent) => {
+    if (event.key === 'Enter') {
+      event.preventDefault(); // Prevent default Enter behavior
+
+      const selection = window.getSelection();
+      if (!selection || !selection.rangeCount) return;
+
+      const range = selection.getRangeAt(0);
+      const newNode = document.createElement(event.shiftKey ? 'br' : 'p');
+
+      if (!event.shiftKey) {
+        // For <p> tags, add a line break to avoid collapsing
+        newNode.innerHTML = '<br>';
+      }
+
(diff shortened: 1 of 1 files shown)


builder-adi (commented):
LGTM

yash-builder (approved):

midhunadarvin (commented):
❔**question:** ‏
These changes are for gen1 sdk right ? Do we need the same changes in gen2 ?

anaghav2023:
> ❔**question:** ‏ These changes are for gen1 sdk right ? Do we need the same changes in gen2 ?

Yes, I was wondering if this is the right approach to this problem. If yes then I'll make similar change in gen2 as well.

packages/react/src/blocks/Text.tsx
```diff
@@ -22,8 +22,39 @@ class TextComponent extends React.Component<TextProps> {
...
+      range.deleteContents();
+      range.insertNode(newNode);
+
+      // Move caret after the newly inserted element
+      range.setStartAfter(newNode);
+      range.setEndAfter(newNode);
+      selection.removeAllRanges();
+      selection.addRange(range);
+    }
+  };
```
sidmohanty11, line 56: shouldn't we handle this inside our webapp? I think we can handle/update the logic here:
https://github.com/BuilderIO/builder-internal/blob/54c7e608f5f4599428919ce3bb49c2d115e4474d/packages/app/components/BlocksOverlay.tsx#L700

I think SDKs should only handle the rendering part and this event listener is an extra addition that we'll never be able to use without adding `contentEditable=true` that only gets added in the visual editor. What do you think?

sidmohanty11 (changes requested):
marking this as request changes so that @samijaber can take a look once about this approach before merging this

packages/react/src/blocks/Text.tsx
anaghav2023: Closing this PR, will raise on the webapp side

PR #3932, opened 2025-02-26
fix[angular-gen2]: new blocks getting added at the top while editing

sidmohanty11:
## Description

https://builder-io.atlassian.net/browse/ENG-7718

_Screenshot_


packages/sdks-tests/src/e2e-tests/editing.spec.ts
```diff
@@ -455,4 +456,92 @@ test.describe('Visual Editing', () => {
...
+      basePort,
+      sdk,
+      packageName,
+    }) => {
+      test.skip(checkIsGen1React(sdk));
+      test.skip(packageName === 'nextjs-sdk-next-app');
+      test.skip(
+        packageName === 'vue',
+        `Failing on the CI: TypeError: Cannot read properties of null (reading 'namespaceURI')`
+      );
```
sidmohanty11, line 472: this works on local
Screenshot 2025-02-28 at 3 16 36 PM

sidmohanty11: Screenshot 2025-02-28 at 3 17 15 PM

sidmohanty11:
@shubham-builder `ngAfterContentInit` gets called only once at the start of the render which places/adds blocks in the array that are already present in the content JSON and this is the reason why if we start from scratch, we're unable to repro this issue.

This never gets called again and the myContent array doesn't update the nodes when blocks changes. We need to use the `ngAfterContentChecked` hook because this runs every time change detection is triggered and runs after the children have been initialized. This helps us to project them after they are completely ready.

packages/sdks/mitosis.config.cjs
```diff
@@ -400,6 +400,20 @@ const ANGULAR_RENAME_NG_ONINIT_TO_NG_AFTERCONTENTINIT_PLUGIN = () => ({
     post: (code) => {
       if (code?.includes('selector: "blocks-wrapper"')) {
         code = code.replace('ngOnInit', 'ngAfterContentInit');
+        const ngOnChangesIndex = code.indexOf('ngOnChanges');
+        if (ngOnChangesIndex > -1) {
+          code = code.replace(/^\s*\/\/\s*@ts-expect-error.*$/gm, '');
+          code = code.replace(
+            'ngOnChanges(changes: SimpleChanges) {',
+            // trigger a re-render of the view when blocks got updated and the new children content has been fully initialized
```
samijaber, line 408: Could we document here in more detail? A brief summary of why it's needed, essentially a tl;dr of your looms. You'll know how to phrase this in the most accurate way, but something along the lines of:

> Since the angular SDK manually handles the creation of the dynamic blocks and attaching them as children of `BlocksWrapper` in the DOM, it must also manually handle their re-renders on content change in the visual editor. `ngAfterContentChecked` runs after children were checked for changes, which is the earlier point we can safely append new blocks, otherwise they are appended outside the existing array of Blocks, therefore pushed to the top of the list.

samijaber (approved):
Approving since all my comments were about needing additional comments and wider test coverage, but nothing blocking the fix itself

samijaber:
@sidmohanty11 I'm going to merge this as-is given the urgency of getting the fix in, feel free to spin up another PR for my feedback 👍🏽

PR #3937, opened 2025-02-27
fix[dynamic-renderer]: ENG-8440 Button links cannot be used in Angular Gen 2 SDK

yash-builder:
## Description

This PR replaces the switch-case logic with a Map for dynamically selecting components in `dynamic-renderer`. The previous approach caused an assertion error when users typed.

_Loom_
https://www.loom.com/share/1229e7974db44743921df588ca1bd1bb?sid=ea3ccc9c-41de-4efe-b499-f61ca9244cbd

### DIFF
--- a/.changeset/loud-jars-sleep.md
+++ b/.changeset/loud-jars-sleep.md
@@ -2,4 +2,4 @@
 "@builder.io/sdk-angular": patch
 ---
 
-Resolved assertion error encountered when dynamically switching components
+Fix: crashes when visually editing blocks (encountered when SDK dynamically switched HTML elements)
--- a/.changeset/loud-jars-sleep.md
+++ b/.changeset/loud-jars-sleep.md
@@ -1,0 +1,5 @@
+---
+"@builder.io/sdk-angular": patch
+---
+
+Resolved assertion error encountered when dynamically switching components
--- a/packages/sdks-tests/src/e2e-tests/dynamic-button.spec.ts
+++ b/packages/sdks-tests/src/e2e-tests/dynamic-button.spec.ts
@@ -7,7 +7,7 @@ import {
 } from '../helpers/visual-editor.js';
 import { DYNAMIC_BUTTON } from '../specs/dynamic-button.js';
 
-test.describe.only('Dynamic Button', () => {
+test.describe('Dynamic Button', () => {
   test('should render a button', async ({ page, sdk, basePort }) => {
     test.skip(sdk !== 'angular');
 
--- a/packages/sdks-tests/src/e2e-tests/dynamic-button.spec.ts
+++ b/packages/sdks-tests/src/e2e-tests/dynamic-button.spec.ts
@@ -1,0 +1,55 @@
+import { expect } from '@playwright/test';
+import { test } from '../helpers/index.js';
+import {
+  cloneContent,
+  launchEmbedderAndWaitForSdk,
+  sendContentUpdateMessage,
+} from '../helpers/visual-editor.js';
+import { DYNAMIC_BUTTON } from '../specs/dynamic-button.js';
+
+test.describe.only('Dynamic Button', () => {
+  test('should render a button', async ({ page, sdk, basePort }) => {
+    test.skip(sdk !== 'angular');
+
+    await launchEmbedderAndWaitForSdk({
+      path: '/dynamic-button',
+      basePort,
+      page,
+      sdk,
+    });
+
+    const buttonLocator = page
+      .frameLocator('iframe')
+      .locator('[builder-id="builder-b53d1cc2bcbb481b869207fdd97ee1db"]');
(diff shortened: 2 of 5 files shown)


samijaber (commented):
Can we add an integration test for this feature? It would be a visual editing test where the user is updating the value of the link

packages/sdks-tests/src/e2e-tests/dynamic-button.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+import {
+  cloneContent,
+  launchEmbedderAndWaitForSdk,
+  sendContentUpdateMessage,
+} from '../helpers/visual-editor.js';
+import { DYNAMIC_BUTTON } from '../specs/dynamic-button.js';
+
+test.describe('Dynamic Button', () => {
+  test('should render a button', async ({ page, sdk, basePort }) => {
+    test.skip(sdk !== 'angular');
```
sidmohanty11, line 12: this test should work for all the gen2 sdks

packages/sdks-tests/src/e2e-tests/dynamic-button.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      page,
+      newContent,
+      model: 'page',
+    });
+
+    const updatedButtonLocator = page
+      .frameLocator('iframe')
+      .locator('[builder-id="builder-b53d1cc2bcbb481b869207fdd97ee1db"]');
+
+    await expect(updatedButtonLocator).toBeVisible();
```
sidmohanty11, line 53: also can you check the href value?

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
```diff
@@ -196,14 +197,21 @@ export default class DynamicRenderer {
...
     if (typeof this.TagName === 'string') {
-      switch (this.TagName) {
-        ${htmlElements.map((el) => `case '${el}': this.TagName = Dynamic${capitalize(el)}; break;`).join('\n        ')}
-        default:
-          this.tagName = this.TagName;
-          this.TagName = DynamicElement;
-          break;
+      if (this.tagComponentMap[this.TagName]) {
+        this.TagName = this.tagComponentMap[this.TagName];
+      } else {
```
sidmohanty11, line 212: If I understand it correctly, map lookups are faster than switch case statements? and is that why angular was unable to createComponent quickly when we type? 🤯

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
```diff
@@ -196,14 +197,21 @@ export default class DynamicRenderer {
 
   constructor(private vcRef: ViewContainerRef) {}
 
+  private tagComponentMap: { [key: string]: any } = {
+    ${htmlElements.map((el) => `'${el}': Dynamic${capitalize(el)}`).join(',\n    ')}
```
sidmohanty11, line 201: also can you check the `dynamic-a` for this part of the jira:
```
Additionally, if you use anchor tags by setting the link to #stuff and place a block with html attribute of id = stuff it will open a new page by appending #stuff to the url rather than focusing on the html element named stuff.
```

sidmohanty11 (commented):
LGTM! 🚀 

needs a changeset though

packages/sdks-tests/src/e2e-tests/dynamic-button.spec.ts
yash-builder: Gotcha! let me remove that statement. I thought we have to intentionally make this test case only for angular. Please do correct me if I am wrong 🤔

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
yash-builder: Yess!!

packages/sdks/output/angular/scripts/generate-dynamic-renderer.mjs
yash-builder: Oh I missed this part. Thank you for adding it! Let me check it

.changeset/loud-jars-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"@builder.io/sdk-angular": patch
+---
+
+Resolved assertion error encountered when dynamically switching components
```
samijaber, line 5: This changeset is too low-level. I prefer changesets to focus on what customer-facing impact they have. In this case, the fact that Visual Editing could cause a crash:

```suggestion
Fix: crashes when visually editing blocks (encountered when SDK dynamically switched HTML elements)
```

samijaber (approved):
✅ , just update the changeset and good to go

PR #3939, opened 2025-02-27
Add snippets for Querying Cheatsheet

bwreid:
## Description

Adds snippets to be used within the [Querying Cheatsheet](https://www.builder.io/c/docs/querying) doc. The only files that will be linked within the docs are the ones within the `queries/` directory -- not the `querying-cheatsheet/index.tsx` file.

As this if my first commit to this repo, please let me know if I've gotten anything wrong! 🙏


packages/sdks/snippets/gen1-react/src/routes/query-cheatsheet/style.css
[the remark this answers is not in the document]
uttej-vsk: @bwreid The goal is to add only the bare minimum items you will use in the docs. If you believe that any file that would not break your code or tests, you can safely remove it.

packages/sdks/snippets/react/src/routes/query-cheatsheet/queries/gte.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+export const getProduct = async (options: {
+  apiKey: string;
+  [key: string]: unknown;
+}) =>
+  fetchOneEntry({
+    model: 'product',
+    query: { 'data.price': { $gte: 700 } },
+    ...options,
+  });
```
samijaber, line 11: you should rely on `GetContentOptions` instead of `unknown` to let types flow from the SDK

```suggestion
  import { fetchOneEntry, GetContentOptions } from '@builder.io/sdk-react';
  
  const getProduct = async (options: GetContentOptions) =>
    fetchOneEntry({
      query: { 'data.price': { $gte: 700 } },
      ...options,
    });
```

same applies to gen1 SDK (remove `unknown`). You'll have to rework the logic a bit to provide `model` from the component for your tests

samijaber: Also, the `async` are redundant here and can be removed. They are only useful if you use `await` in the body.

bwreid: Thanks @samijaber. Those changes have been made -- please let me know if you have any other thoughts!

packages/sdks/snippets/gen1-react/src/routes/query-cheatsheet/queries/regex.tsx
```diff
@@ -0,0 +1,7 @@
+import { builder } from '@builder.io/react';
+
+export const getProduct = (options: { [key: string]: any }) =>
```
samijaber, line 3: similarly for all gen1 code:

```suggestion
import type { GetContentOptions } from '@builder.io/sdk';

export const getProduct = (options: GetContentOptions) =>
```

bwreid: Thank you. 🙏 I should have assumed the type would be somewhere when I couldn't find it in `@builder.io/react`. Apologies. It's been updated.

samijaber (approved):

PR #3944, opened 2025-03-03
Add targeting snippets

bwreid:
## Description

Adds snippets to be used within the [Targeting Cheatsheet](https://www.builder.io/c/docs/targeting-cheatsheet) doc. The only files that will be linked within the docs are the non-index files within the `targeted-page/` directory.

### DIFF
--- a/packages/sdks/snippets/gen1-react/src/routes/targeted-page/custom-targeting-request.tsx
+++ b/packages/sdks/snippets/gen1-react/src/routes/targeted-page/custom-targeting-request.tsx
@@ -1,0 +1,9 @@
+import { builder } from '@builder.io/react';
+
+export default ({ model, urlPath }: { model: string; urlPath: string }) =>
+  builder.get(model, {
+    userAttributes: {
+      audience: ['mens-fashion'],
+      urlPath,
+    },
+  });
--- a/packages/sdks/snippets/gen1-react/src/routes/targeted-page/index.tsx
+++ b/packages/sdks/snippets/gen1-react/src/routes/targeted-page/index.tsx
@@ -1,0 +1,58 @@
+import { BuilderComponent, builder, useIsPreviewing } from '@builder.io/react';
+import { useEffect, useState } from 'react';
+import { useSearchParams } from 'react-router-dom';
+import FourOhFour from '../../components/404';
+
+import customTargetingMultiRequest from './custom-targeting-multi-request';
+import customTargetingRequest from './custom-targeting-request';
+import desktopRequest from './desktop-request';
+import loggedInRequest from './is-logged-in-request';
+import mobileRequest from './mobile-request';
+import noTargetRequest from './no-target-request';
+
+const MODEL = 'targeted-page';
+builder.init('ee9f13b4981e489a9a1209887695ef2b');
+
+export default function TargetedPage() {
+  const isPreviewingInBuilder = useIsPreviewing();
+  const [notFound, setNotFound] = useState(false);
+  const [content, setContent] = useState();
+  const [searchParams] = useSearchParams();
+
+  const fnMap: { [key: string]: any } = {
+    desktop: desktopRequest,
+    mobile: mobileRequest,
+    'mens-fashion': customTargetingRequest,
+    'multi-target': customTargetingMultiRequest,
+    'is-logged-in': loggedInRequest,
+  };
+
+  useEffect(() => {
+    async function fetchContent() {
+      const options = { model: MODEL, urlPath: window.location.pathname };
+      const target: string = searchParams.get('target') || '';
+      const targetFn = target ? fnMap[target] : noTargetRequest;
+      const content = await targetFn(options);
+
+      setContent(content);
+      setNotFound(!content);
+
+      if (content?.data.title) {
+        document.title = content.data.title;
+      }
+    }
+    fetchContent();
+  }, [window.location.pathname]);
+
+  if (notFound && !isPreviewingInBuilder) {
+    return <FourOhFour />;
+  }
+
+  return (
+    <>
+      <h1>Targeting Snippet</h1>
+      <hr />
+      <BuilderComponent model={MODEL} content={content} />
+    </>
+  );
+}
--- a/packages/sdks/snippets/gen1-react/src/routes/targeted-page/is-logged-in-request.tsx
+++ b/packages/sdks/snippets/gen1-react/src/routes/targeted-page/is-logged-in-request.tsx
@@ -1,0 +1,9 @@
+import { builder } from '@builder.io/react';
+
+export default async ({ model, urlPath }: { model: string; urlPath: string }) =>
+  builder.get(model, {
+    userAttributes: {
+      isLoggedIn: true,
+      urlPath,
+    },
+  });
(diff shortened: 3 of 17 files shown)


packages/sdks/snippets/gen1-react/src/routes/targeted-page/is-logged-in-request.tsx
```diff
@@ -0,0 +1,9 @@
+import { builder } from '@builder.io/react';
+
+export default async ({ model, urlPath }: { model: string; urlPath: string }) =>
```
samijaber, line 3: - unnecessary async
- for consistency all requests fns should have the same name. I see some don't have a name at all, and some are called `request`. I find `request` to be a bit of a special keyword in JavaScript land. I'd opt for a name like `targettedRequest`

bwreid: Thank you, @samijaber. I've updated the code to remove the last few `async` keywords. I've also used the `GetContentOptions` type in Gen 1 now that I know where it is.

packages/sdks/snippets/gen1-react/src/routes/targeted-page/custom-targeting-request.tsx
```diff
@@ -0,0 +1,11 @@
...
+import type { GetContentOptions } from '@builder.io/sdk';
+
+export default function targetedRequest({ model, url }: GetContentOptions) {
+  return builder.get(model!, {
+    userAttributes: {
+      audience: ['mens-fashion'],
+      urlPath: url,
+    },
+  });
+}
```
samijaber, line 11: - Recently added `!` to the list of things we want to avoid in our docs: https://www.notion.so/builderio/Docs-code-snippet-guidelines-1283d7274be580d1a96cdafcf6cab708?pvs=4#1ab3d7274be58085b038e2b7a609c435

- `GetContentOptions` is specific to the second argument of `builder.get`, so it's more consistent to use it only for that part
- `url` in gen1 can be used to replace `userAttributes.urlPath`, as per its code comment: 
CleanShot 2025-03-04 at 14 53 43@2x

Putting all that together:
```suggestion
export default function targetedRequest(modelName: string, options: GetContentOptions) {
  return builder.get(modelName, {
    userAttributes: {
      audience: ['mens-fashion'],
    },
    ...options,
  });
}
```

samijaber (commented):
Just took a look at the doc: https://www.builder.io/c/docs/targeting-cheatsheet#examples-of-custom-targeting-attributes

Are we reworking the doc completely? Right now that page uses `builder.setUserAttributes()` to illustrate how those attributes can be set, whereas this PR is using `builder.get()` calls

packages/sdks/snippets/react/src/routes/targeted-page/is-logged-in-request.tsx
```diff
@@ -0,0 +1,15 @@
...
+) {
+  return fetchOneEntry({
+    apiKey,
+    model,
+    userAttributes: {
+      isLoggedIn: true,
+      urlPath,
+    },
+  });
+}
```
samijaber, line 15: The arguments of `targetedRequest` add a layer of indirection that can complicate the code snippet for the end-user.

I would stick with a philosophy of "spread everything" to keep this wrapper function as unobtrusive as possible:

```suggestion
import { fetchOneEntry, GetContentOptions } from '@builder.io/sdk-react';

export default function targetedRequest(options: GetContentOptions) {
  return fetchOneEntry({
    ...options,
    userAttributes: {
      isLoggedIn: true,
      ...options.userAttributes,
    },
  });
}
```

bwreid:
> Just took a look at the doc: https://www.builder.io/c/docs/targeting-cheatsheet#examples-of-custom-targeting-attributes
> 
> Are we reworking the doc completely? Right now that page uses builder.setUserAttributes() to illustrate how those attributes can be set, whereas this PR is using builder.get() calls

I am trying to make some significant changes to the doc. There are more details in the [Jira ticket](https://builder-io.atlassian.net/browse/EDU-169?focusedCommentId=28441), but in short I would like to separate out the code blocks from the more generalized information about targeting.

I also worked on [the Querying Cheatsheet](https://www.builder.io/c/docs/querying), which has a similar name but is structured very differently. This page seems focused around being able to quickly find the code snippet you need and copy it to your code, while the current Targeting Cheatsheet acts a bit more like a tutorial.

If you disagree with this approach, however, please feel free to let me know. I'm very much still getting a handle on all the documentation.

packages/sdks/snippets/react/src/routes/targeted-page/is-logged-in-request.tsx
bwreid: @samijaber Is your assumption then that the arguments provided to `targetedRequest()` will look something like this?

```javascript
const options = {
  apiKey: BUILDER_API_KEY,
  model: MODEL,
  userAttributes: {
    urlPath: window.location.pathname,
  },
};
```

As someone new to the code, I feel like having `urlPath` explicit within the code snippet feels important as the request will not work without it. Do you think something like this is too verbose?

```javascript
import { fetchOneEntry, GetContentOptions } from '@builder.io/sdk-react';

export default function targetedRequest(options: GetContentOptions) {
  return fetchOneEntry({
    ...options,
    userAttributes: {
      isLoggedIn: true,
      urlPath: options.userAttributes?.urlPath,
      ...options.userAttributes, // Include additional userAttributes
    },
  });
}
```

packages/sdks/snippets/gen1-react/src/routes/targeted-page/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+      setContent(content);
+      setNotFound(!content);
+
+      if (content?.data.title) {
+        document.title = content.data.title;
+      }
+    }
+    fetchContent();
+  }, [window.location.pathname]);
```
samijaber, line 48: heads up that `useEffect` cannot listen to changes in `window.location.pathname`.

this is used for our own testing, so as long as you get the desired behaviour it's fine i guess, but more of a heads up in general

samijaber (commented):
hmm, i see these as two separate things:

- _setting_ a user attribute: 
  - gen1: `builder.setUserAttribute()` 
  - gen2: `setClientUserAttributes()`

- `using` a user attribute when fetch content: `userAttributes: {}` argument to `builder.get` and `fetchOneEntry`/`fetchEntries`

The doc currently has the former, you're replacing it with a latter. I think we'd like both ideally!

packages/sdks/snippets/gen1-react/src/routes/targeted-page/index.tsx
bwreid: Thanks, I did not know this but it makes sense! 👍

packages/sdks/snippets/react/src/routes/targeted-page/setUserAttributes.tsx
```diff
@@ -0,0 +1,8 @@
+import { setClientUserAttributes } from '@builder.io/sdk-react';
+
+export default function setTargetingAttributes() {
+  setClientUserAttributes({
+    device: 'desktop',
+    audience: ['recent-shopper', 'womens-fashion'],
+  });
+}
```
bwreid, line 8: @samijaber Here is the option that's failing. I am using this function and file similarly in the Gen 1 directory.

This function _does_ set the cookies appropriately but `fetchOneEntry()` does not seem to pick up the details in the cookie.

bwreid: @samijaber I was able to get this passing with a change to the content entry rather than a change to the code.

samijaber (approved):

PR #3946, opened 2025-03-05
React Gen 2- Product Details - Blueprints snippet 

uttej-vsk:
## Description

This PR aims to add Product Details blueprints code to GitHub for React Gen 2 SDK as part of this [epic in the JIRA-138](https://builder-io.atlassian.net/browse/EDU-138)

@samijaber Please review it.

I have added more code to the `useEffect()`, which is only for cleanup. I know this isn't very significant for this scale of code, but I thought it would be better to use some best practices so that customers would know our code is resilient.


packages/sdks/snippets/react/src/routes/blueprints/ProductDetails.tsx
```diff
@@ -0,0 +1,47 @@
...
+      if (isMounted) {
+        setProductDetails(productData);
+      }
+    };
+    fetchProductDetails();
+
+    return () => {
+      isMounted = false;
+    };
+  }, []);
```
samijaber, line 28: I don't see the point of `isMounted`. What does it achieve that can't be done without it, like so:

```suggestion
  useEffect(() => {
    fetchOneEntry({
        model: 'product-details',
        apiKey: 'ee9f13b4981e489a9a1209887695ef2b',
        query: {
          'data.handle': 'jacket',
        },
      }).then(productData => {
        setProductDetails(productData);
      }
  }, []);
```

uttej-vsk: @samijaber I was reading about abort controllers, and then the react doc suggested the below:
Screenshot 2025-03-05 at 10 28 27 AM

I'm totally with you on simplifying the code, and I had the same code as yours before pushing it to the PR. The only reason I added a flag is to stop fetching twice and updating the state after unmounts.

samijaber: I'd ignore this. since your `useEffect` has an empty dependencies array, it will only run once no matter what, so there's no point in aborting the fetch or ignoring the result.

The advice you screenshotted is relevant when `useEffect` has items in its dependencies array, which means it will execute more than once.

samijaber: As for async/await: it adds bloat to the code to declare an async function and then call it. One of our most important rules around snippets in docs is simplicity: remove any code that is not absolutely necessary

uttej-vsk:
I totally agree to this rule, Sami! I have already committed the change and will push it and removed as much as I can. Thanks for sharing your two cents :)
Regards,
Uttej.

On Mar 5, 2025, at 12:19 PM, Sami Jaber ***@***.***> wrote:

﻿

@samijaber commented on this pull request.

________________________________

In packages/sdks/snippets/react/src/routes/blueprints/ProductDetails.tsx<https://github.com/BuilderIO/builder/pull/3946#discussion_r1981935035>:

> +  useEffect(() => {
+    let isMounted = true;
+    const fetchProductDetails = async () => {
+      const productData = await fetchOneEntry({
+        model: 'product-details',
+        apiKey: 'ee9f13b4981e489a9a1209887695ef2b',
+        query: {
+          'data.handle': 'jacket',
+        },
+      });
+      if (isMounted) {
+        setProductDetails(productData);
+      }
+    };
+    fetchProductDetails();
+
+    return () => {
+      isMounted = false;
+    };
+  }, []);


As for async/await: it adds bloat to the code to declare an async function and then call it. One of our most important rules around snippets in docs is simplicity: remove any code that is not absolutely necessary

packages/sdks/snippets/react/src/routes/blueprints/ProductDetails.tsx
uttej-vsk: Thanks for the explanation, sami! i have updated the code.

packages/sdks/snippets/react/src/routes/blueprints/ProductDetails.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const [productDetails, setProductDetails] = useState<BuilderContent | null>(
+    null
+  );
+
+  useEffect(() => {
+    fetchOneEntry({
+      model: 'product-details',
+      apiKey: 'ee9f13b4981e489a9a1209887695ef2b',
+      query: {
+        'data.handle': 'jacket',
```
samijaber, line 14: IMO `'jacket'` should be fetched dynamically from the URL instead:
```ts
  let { productName } = useParams();

//...
        'data.handle': 'productName',
```

and in `main.tsx`:

```ts
,
  {
    path: '/product/category/:productName',
    element: <ProductDetails />,
  },
```

packages/sdks/snippets/react/src/routes/blueprints/ProductDetails.tsx
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  return (
+    <div className="product-details">
+      <h1>{productDetails.data?.name}</h1>
+      <img
+        src={productDetails.data?.image}
+        alt={productDetails.data?.name}
+        width="400"
+        height="500"
```
samijaber, line 32: I'd consider these unnecessary layout attributes
```suggestion
```

packages/sdks/snippets/react/src/routes/blueprints/ProductDetails.tsx
uttej-vsk: @samijaber Oh yeah, in all the other frameworks in the second PR, I fetched the handle dynamically, but somehow, I didn't do it here. I will push the correct one.

packages/sdks/snippets/react/src/routes/blueprints/ProductDetails.tsx
uttej-vsk: @samijaber Yeah, but the image is taking up more space, so to restrict that, I have added width and height. 

**If i keep those attributes:**
Screenshot 2025-03-10 at 1 02 43 PM

**If i remove them:**
Screenshot 2025-03-10 at 1 02 32 PM

samijaber: I'm ok with that, as we want to focus on only having the bare minimum needed for the example to work and make sense.

uttej-vsk: @samijaber Pushed the code

PR #3947, opened 2025-03-05
fix[qwik]: ENG-7299 Default value not updating for custom components on ContentEditor

yash-builder:
## Description

This PR addresses a critical reactivity issue in the Qwik SDK where deeply nested content updates weren't properly triggering. 

The core problem was that when content was updated through the Builder.io editor, the changes were correctly stored in state but weren't causing components to re-render.

 I've implemented a key-based approach for the `InteractiveElement` component that uses a dynamic key derived from the component options, forcing `Qwik` to update it's component when `options` changes. This effectively bypasses limitations in Qwik's resumability model where prop changes to dynamic components aren't always detected. While using `JSON.stringify` in keys isn't ideal for **_performance_**, but it provides a reliable solution that ensures content updates are immediately reflected in the UI without requiring page refreshes.

JIRA
https://builder-io.atlassian.net/browse/ENG-7299

_Loom_
https://www.loom.com/share/2b4d1b2872fe4151b1a9f2f131bab780?sid=7fc0f143-0623-41f4-aa94-e947530dde55


packages/sdks/e2e/qwik-city/src/routes/[...index]/index.tsx
```diff
@@ -67,7 +69,23 @@ export default component$(() => {
...
+            name: 'Description',
+            component: Description,
+            inputs: [
+              {
+                name: 'text',
+                type: 'string',
+                defaultValue: 'Hello',
+              },
+            ],
+          },
```
Comment (clyde-builderio) [inline packages/sdks/e2e/qwik-city/src/routes/[...index]/index.tsx:87]: Preferably make this also a variable defined above and add it to the array. In future we may have more custom components and this code may get hard to read

packages/sdks/mitosis.config.cjs
```diff
@@ -1195,6 +1195,15 @@ module.exports = {
     "readystatechange"`
                 );
               }
+              if (json.name === 'InteractiveElement') {
+                console.log('code', code);
```
clyde-builderio, line 1199: Remove the console.log()

clyde-builderio (changes requested):

packages/sdks/e2e/qwik-city/src/routes/[...index]/index.tsx
Comment (yash-builder) [reply packages/sdks/e2e/qwik-city/src/routes/[...index]/index.tsx:87]: Will make the changes. Thank you @clyde-builderio

samijaber (changes requested):
- to avoid `JSON.stringify()`'ing the entire set of props, I played with using the `track` Qwik helper and managed to get something working: https://qwik.dev/examples/introduction/hello-world/#f=7VZNS8NAEL37K4YiJGJYUryUkNSLXjzYg0cRqkmqscVKRTEs%2B999M7NJ86HixYNgL002s7NvPnbe6zZNfBJ%2F1jVRy7nyJHNN13RufddUBxq%2BRpD1msrLB0ur7fY0afSBG%2FUay46UNcIchphWlPMoxkZxYLDoDEISC2xTnYJ%2BxOE%2FaWmvQWT8Z22koQVHATOum4B0GGEjqBGd1RizVd7SSeIjxemNZ9BpjmstxKj%2BJWthjPzDqBUgOFHlxD4DjYt1Wb9g7%2BLuEQxh%2BC0UvGYPEdGyPX9DRnbnt%2FlDGOKt44vUvw996OAaxjeMmn8dzAZjDgR2nNFUPkpknNx%2BgbSv%2BEmLhHGXUCr%2FcsezCU%2BpCQ%2FNV6ir4elSRQgp0W2HmQ3Lt4jKjSAdxcrGSEe5UWxHDl3uq96gaFScJRACokgomD6%2FE4i0KggMlK%2BDCARaFKhjQrMIiml3Xz3BbEZxQM7NyadCmH2J5Czp4mpxabT21aqmVEttjcFiH6HjQmS2k0bG6O%2FeLwFFr4N3oX34WiD3ig6p%2BirbQ0D68D%2FJ%2F%2Bok%2FwA

The idea is to use `useEffect$()`'s `track()` to listen to changes individually on each object value, and increment a render count that is then used in `key`.  we actually used to do something like this in the past: https://github.com/BuilderIO/builder/pull/3064/files

- since this bugfix only impacts the visual editing experience, it would be a shame for it to impact production content. the solution I offered here should help with that: use a counter signal that only updates when we receive a new value from the visual editor to force re-renders. Make sure to only do it for Qwik and not other frameworks.


- no matter what the solution is, let's try to put it in the `.lite.tsx` file directly (using `useTarget`). If that's not possible, then let's use a json plugin and add it programmatically via `json.hooks.*`. Using a string replace is a last resort as it is particularly brittle

samijaber:
That's fair, my main concern is avoiding unnecessary `String.replace`:

> no matter what the solution is, let's try to put it in the .lite.tsx file directly (using useTarget). If that's not possible, then let's use a json plugin and add it programmatically via json.hooks.*. Using a string replace is a last resort as it is particularly brittle

I suggest you:
- write `key={useTarget({ qwik: /**/, default: undefined })}`
- add a mitosis `json` plugin that does `delete json.children[0].children[0].bindings.key`, and apply it to all non-qwik

this way we don't do brittle `.replace()`.

samijaber:
Ah, sorry! I missed a special scenario here: you also have to delete the relevant `useTarget`-related code from `json.targetBlocks`, otherwise the generators will take that and apply it to the `key` binding regardless.

packages/sdks/mitosis.config.cjs
```diff
@@ -1235,6 +1235,14 @@ module.exports = {
...
                 );
               }
+              if (json.name === 'InteractiveElement') {
+                code = code.replace(`<props.Wrapper {...props.wrapperProps} attributes={attributes.value}>
+            <Slot></Slot>
+          </props.Wrapper>`, 
+                `<props.Wrapper key={'wrapper-'+state.forceRenderCount} {...props.wrapperProps} attributes={attributes.value}>
+            <Slot></Slot>
+          </props.Wrapper>`)
+              }
```
samijaber, line 1245: - use `json` plugin instead of `code` plugin
- move this to its own plugin with some good naming/documentation so its explained why we're doing this

packages/sdks-tests/src/e2e-tests/editing.spec.ts
```diff
@@ -332,13 +343,18 @@ test.describe('Visual Editing', () => {
         model: 'page',
       });
 
-      await item1.click();
+      // Add a small delay to allow the component to fully update
+      await page.waitForTimeout(500);
```
samijaber, line 347: Adding timeouts is a slippery slope for e2e testing. It shouldn't be done unless absolutely necessary. There is almost always a better approach. It's better to ask playwright to wait for an element, or text, or event to occur.

https://playwright.dev/docs/api/class-page#page-wait-for-timeout
CleanShot 2025-03-12 at 12 37 29@2x

packages/sdks/src/components/block/components/interactive-element.lite.tsx
```diff
@@ -51,6 +59,16 @@ export default function InteractiveElement(props: InteractiveElementProps) {
...
 
+  // Use onUpdate to track prop changes (Mitosis equivalent of useEffect/useTask)
+  onUpdate(() => {
+    useTarget({
+      qwik: () => {
+        state.forceRenderCount = state.forceRenderCount + 1;
+      },
+      default: () => {},
+    });
+  }, [props.wrapperProps, props.block?.component?.options]);
```
samijaber, line 70: oh that's a lot simpler than what I tried. This works fine and reacts to any deeply nested change in `component.options`?

packages/sdks-tests/src/e2e-tests/editing.spec.ts
yash-builder: 😲

packages/sdks/mitosis.config.cjs
yash-builder: Okie. Let me try with json instead of removing or deleting data let me try to add the code

packages/sdks/mitosis.config.cjs
```diff
@@ -960,6 +960,21 @@ const QWIK_ONUPDATE_TO_USEVISIBLETASK = () => ({
   },
 });
 
+const QWIK_FORCE_RENDER_COUNT_FOR_RENDERING_CUSTOM_COMPONENT_DEFAULT_VALUE = () => ({
+  json: {
+    post: (json) => {
+      if (json.name === 'InteractiveElement') {
+        console.log('json', JSON.stringify(json, null, 2));
```
samijaber, line 967:
```suggestion
```

packages/sdks-tests/src/e2e-tests/editing.spec.ts
```diff
@@ -332,13 +343,19 @@ test.describe('Visual Editing', () => {
...
       });
 
-      await item1.click();
+      await expect(
+        page.frameLocator('iframe').getByText('Item 1')
+      ).toBeVisible({ timeout: 5000 });
+      
+      // Re-query the item1 element as it might have been recreated
+      const updatedItem1 = page.frameLocator('iframe').getByText('Item 1');
+      await updatedItem1.click();
```
samijaber, line 352: we want to avoid hardcoded timeouts unless there's a special reason for them. Playwright has automated waiting and retrying so that we don't need to provide hardcoded timeout values all the time. Also, in this case, the default is already 5000 so this does nothing: https://github.com/BuilderIO/builder/blob/b351bd82895c40394e9c022af2c9762cbb1a8add/packages/sdks-tests/playwright.config.ts#L69

you also don't need to redefine the locator twice:
```suggestion
      // Re-query the item1 element as it might have been recreated
      const updatedItem1 = page.frameLocator('iframe').getByText('Item 1');
      await expect(updatedItem1).toBeVisible();      
      await updatedItem1.click();
```

packages/sdks-tests/src/e2e-tests/custom-components.spec.ts
```diff
@@ -9,7 +9,7 @@ const LAST_COMPONENT_REGISTERED_MESSAGE =
 
 test.describe('Custom components', () => {
   test('correctly renders custom component', async ({ page, packageName, sdk }) => {
-    test.skip(!['angular', 'react'].includes(sdk));
+    test.skip(!['angular', 'react', 'qwik-city'].includes(sdk));
```
sidmohanty11, line 12: this line doesn't have any impact, `sdk` can be `qwik`. `packageName`s are the different e2e/snippet server names and can be `qwik-city`

packages/sdks-tests/src/e2e-tests/editing.spec.ts
```diff
@@ -332,14 +344,19 @@ test.describe('Visual Editing', () => {
...
+      await expect(updatedItem1).toBeInViewport();
+      await updatedItem1.click();
 
       const detailElement = page.frameLocator('iframe').getByText(NEW_DETAILS_TEXT);
-      await detailElement.waitFor();
+      await expect(detailElement).toBeVisible();
 
       const [titleBox, detailBox] = await Promise.all([
-        item1.boundingBox(),
+        updatedItem1.boundingBox(),
```
sidmohanty11, line 359: no need to update this right?

packages/sdks/e2e/qwik-city/src/routes/[...index]/index.tsx
```diff
@@ -46,6 +48,25 @@ const builderBlockWithClassNameCustomComponent = {
   ],
 };
 
+const CUSTOM_COMPONENT = [
```
Comment (sidmohanty11) [inline packages/sdks/e2e/qwik-city/src/routes/[...index]/index.tsx:51]:
```suggestion
const CUSTOM_COMPONENTS = [
```

packages/sdks/e2e/qwik-city/src/routes/[...index]/index.tsx
```diff
@@ -46,6 +48,25 @@ const builderBlockWithClassNameCustomComponent = {
...
+  {
+  name: 'Hello',
+  component: Hello,
+  inputs: [],
+},
```
Comment (sidmohanty11) [inline packages/sdks/e2e/qwik-city/src/routes/[...index]/index.tsx:56]: I think you added the Hello component for this test:https://github.com/sidmohanty11/builder/blob/general-support-vc/packages/sdks-tests/src/e2e-tests/custom-components.spec.ts#L11 right? you will also need to update the content of it from just hello to `hello World`

packages/sdks-tests/src/e2e-tests/editing.spec.ts
```diff
@@ -152,6 +153,16 @@ test.describe('Visual Editing', () => {
...
 
+  test('correctly updating custom components when default value is not set', async ({ page, basePort, sdk, packageName }) => {
+    test.skip(!['react', 'qwik-city'].includes(packageName));
+    await launchEmbedderAndWaitForSdk({ path: '/custom-components-no-default-value', basePort, page, sdk });
+    const newContent = cloneContent(CUSTOM_COMPONENT_NO_DEFAULT_VALUE);
+    newContent.data.blocks[0].component.options.text = "Hello";
+    await sendContentUpdateMessage({ page, newContent, model: 'page' });
+    const helloWorldText = page.frameLocator('iframe').locator('[builder-id="builder-d01784d1ca964e0da958ab4a4a891b08"]')
+    await expect(helloWorldText).toHaveText('Hello');
+  })
```
sidmohanty11, line 164: by default it should render `Hello` right? because you're passing it as defaultValue? without needing to send the content update message?

packages/sdks-tests/src/e2e-tests/custom-components.spec.ts
yash-builder: Ahh nice catch....I was testing somethings...I forgot to remove this

packages/sdks-tests/src/e2e-tests/editing.spec.ts
yash-builder: Didn't get you...🤔
I am answering this questions based on my understanding. Yes, by default it should render `hello`. `defaultValue` has not been passed yet if you see `CUSTOM_COMPONENT_NO_DEFAULT_VALUE` the object will look like this

```js
component: {
    name: 'Description',
    options: {},
}
```
we will be requiring sendContentUpdateMessage to update the text message from `""` to `"Hello"`

sidmohanty11: shouldn't this: https://github.com/BuilderIO/builder/pull/3947/files#diff-f17fc2e9e6719ba501c3ab4df9c1fef742dc2c5dd3f709c622f6c79cdffcc1bdR99 add the defaultValue? i'm confused because on initial render if there is no initial value provided and we are providing a default value - we should be fallbacking to that

samijaber: @sidmohanty11 your confusion stems from the fact that default values are only used when the block is created in the visual editor to determine the initial value for inputs: they are not used by the SDK at run-time.


As a general rule, when we add a test that involves editing content, i think it's best that it looks like this:

- load initial content
- check the initial value of a block
- make an update to that block
- check the value of the same block, and see that it changed as expected

In this test you're not checking the initial state of the block. You could add a `toBeEmpty()` check:

```ts
    test.skip(!['react', 'qwik-city'].includes(packageName));
    await launchEmbedderAndWaitForSdk({ path: '/custom-components-no-default-value', basePort, page, sdk });

    // check initial value here
    const helloWorldText = page.frameLocator('iframe').locator('[builder-id="builder-d01784d1ca964e0da958ab4a4a891b08"]')
    await expect(helloWorldText).toBeEmpty();

    const newContent = cloneContent(CUSTOM_COMPONENT_NO_DEFAULT_VALUE);
    newContent.data.blocks[0].component.options.text = "Hello";
    await sendContentUpdateMessage({ page, newContent, model: 'page' });
    const helloWorldText = page.frameLocator('iframe').locator('[builder-id="builder-d01784d1ca964e0da958ab4a4a891b08"]')
    await expect(helloWorldText).toHaveText('Hello');
```

Not a huge deal though, more of a general rule of thumb for these types of tests

samijaber (approved):

clyde-builderio (approved):

samijaber (approved):

PR #3953, opened 2025-03-06
chore: review comment updates of 3932

sidmohanty11:
## Description

Followup of #3932 

_Screenshot_


samijaber (approved):

samijaber (commented):
i would add a changeset for this, even if it's a refactor of how the SDK is built. just in case it does somehow introduce a change in behaviour

PR #3957, opened 2025-03-10
feat[vue+svelte]: support of variant containers

sidmohanty11:
## Description

Adds support of Variant Containers in Gen2 Vue and Svelte SDK updating the approach to be synonymous with how we handle A/B tests in Gen2 SDKs (general support) so that we can expand this approach and support VCs in other gen2 SDKs in the future.

Jira
https://builder-io.atlassian.net/browse/ENG-7677

Demo
https://www.loom.com/share/678e5b23a666445d87e3758f32a81b3a


packages/sdks/src/blocks/personalization-container/helpers/inlined-fns.ts
```diff
@@ -194,6 +174,65 @@ export function filterWithCustomTargeting(
...
+    }
+    const winningVariantIndex = variants?.findIndex(function (variant) {
+      return (window as any).filterWithCustomTargeting(
+        attributes,
+        variant.query,
+        variant.startDate,
+        variant.endDate
+      );
+    });
+
```
sidmohanty11, line 219: one idea here to optimize finding the winner again and again is to set a cookie - something like `builder.pblock.id` to the winning index or -1 and then pick that instead in the personalization script? but it will be an extra cookie addition for each block

packages/sdks/src/components/blocks/blocks-wrapper.lite.tsx
```diff
@@ -132,27 +133,29 @@ export default function BlocksWrapper(props: BlocksWrapperProps) {
...
+            flexDirection: 'column',
+          },
+          default: {
+            display: 'flex',
+            flexDirection: 'column',
+            alignItems: 'stretch',
+          },
+        }),
+        ...props.styleProp,
+      }}
```
sidmohanty11, line 150: Mitosis generation of React component creates a <style> block that gets outputted outside of the block/parent div - either we move to style else we can still use CSS and specifically for react delete any subsequent <style> block

sidmohanty11: we will not be able to delete any next element sibling when running the script, so had to write a plugin to update the placement of the style element. the inling of styles didn't feel right. let me know your thoughts. i think a better approach would be to add a useMetadata option for react - injectStyleAtTop that injects style elements at the top of the fragment

packages/sdks/src/blocks/personalization-container/helpers/inlined-fns.ts
sidmohanty11: @samijaber

packages/sdks/src/components/blocks/blocks-wrapper.lite.tsx
sidmohanty11: @samijaber

packages/sdks/src/components/content-variants/content-variants.lite.tsx
```diff
@@ -92,6 +96,13 @@ export default function ContentVariants(props: VariantsProviderProps) {
           id="builderio-init-variants-fns"
           nonce={props.nonce || ''}
         />
+        {SDKS_SUPPORTING_PERSONALIZATION.includes(TARGET) && (
+          <InlinedScript
+            nonce={props.nonce || ''}
+            scriptStr={getInitPersonalizationVariantsFnsScriptString()}
+            id="builderio-init-personalization-variants-fns"
```
samijaber, line 103: existing issue in a/b tests potentially as well: 

these scripts set stuff on `window.*`, which makes me think things might break if users render:

```tsx
<Content /> 
<Content />
```

multiple `Content` components in the same page.

packages/sdks/src/blocks/personalization-container/helpers.ts
```diff
@@ -91,30 +126,46 @@ export function getBlocksToRender({
...
-      return {
-        blocks: winningVariant.blocks,
-        path: `component.options.variants.${variants?.indexOf(winningVariant)}.blocks`,
-      };
+    if (winningVariant && variants) {
+      const variantIndex = variants.indexOf(winningVariant);
+      if (variantIndex !== -1) {
+        return {
+          blocks: winningVariant.blocks,
+          path: `component.options.variants.${variantIndex}.blocks`,
```
samijaber, line 134: we prefix `component.options` in Blocks, so you can remove it from all the paths in this file

packages/sdks/mitosis.config.cjs
```diff
@@ -942,6 +948,30 @@ const QWIK_ONUPDATE_TO_USEVISIBLETASK = () => ({
   },
 });
 
+/**
+ * Updates the placement of the style element in the BlocksWrapper component. Injects it at the top of the fragment.
+ * This is necessary because when we run the inlined script, we only have access to previous sibling elements for deletion.
+ */
+const UPDATE_PLACEMENT_OF_STYLE_ELEMENT_IN_BLOCKS_WRAPPER = () => ({
```
samijaber, line 955: if we can replace this with mitosis logic here: https://github.com/BuilderIO/mitosis/blob/0a49334818422d76efd89f9cdde1af5ab3bedcef/packages/core/src/generators/react/generator.ts#L423-L434

`state.stylesLocation: 'top' | 'bottom'`

packages/sdks/src/components/blocks/blocks-wrapper.lite.tsx
```diff
@@ -33,6 +33,7 @@ export type BlocksWrapperProps = {
   children?: any;
 
   classNameProp?: string;
+  extraAttributesForBlocksWrapper?: Record<string, any>;
```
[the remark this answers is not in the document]
sidmohanty11: added, please take a look

packages/sdks/src/blocks/personalization-container/helpers.ts
sidmohanty11: done!

packages/sdks/mitosis.config.cjs
sidmohanty11: done!

packages/sdks/src/components/content-variants/content-variants.lite.tsx
sidmohanty11: @samijaber just checked, we don't use ids of the scripts anywhere in the inlined scripts. These global window fn initialisations will be defined twice if we use two Contents and it will override the function again in window but not affect the contents. 

https://github.com/sidmohanty11/builder/blob/general-support-vc/packages/sdks/src/components/content-variants/inlined-fns.ts/#L142 - we attach cookies to a specific content id and content ids will be different in each case if multiple Content are using A/B tests feature

samijaber (approved):

packages/sdks/src/components/blocks/blocks.types.ts
```diff
@@ -15,4 +15,5 @@ export type BlocksProps = Partial<
   registeredComponents?: RegisteredComponents;
   linkComponent?: any;
   className?: string;
+  extraAttributesForBlocksWrapper?: Record<string, any>;
```
samijaber, line 18: discussed in slack about how to handle these extra attrs. we want to avoid the extra prop and instead rely on BlocksWrapperProps


- `Blocks` should now receive `BlocksWrapperProps`
- if its provided by user, override the context value completely

samijaber: and lets make sure to properly document this behaviour in JSDoc and the changeset

samijaber (approved):

PR #3968, opened 2025-03-12
Custom Components - Child blocks - svelte & svelteKit all examples

uttej-vsk:
## Description

@samijaber This is part 2 of the Custom Components - Child Blocks [ticket-133](https://builder-io.atlassian.net/browse/EDU-133). It contains Svelte code for all three example snippets. Whenever you have the chance, please review this.

Update: I have pushed svelte and sveltekit to this PR.


packages/sdks-tests/playwright.config.ts
```diff
@@ -66,7 +66,7 @@ export default defineConfig({
     /**
      * Increase the default timeout for snippet tests because they're not deterministic.
      */
-    timeout: testType === 'snippet' ? 30000 : 5000,
+    timeout: testType === 'snippet' ? 60000 : 5000,
```
yash-builder, line 69: curious question:
why do we have to ramp up the time for snippets?

uttej-vsk: @yash-builder I tried debugging this issue and found this answer in [Stackoverflow](https://stackoverflow.com/a/77068413).

> You currently don't specify a timeout when you call toBeVisible(), and your code defaults to 5000ms, which is 5 seconds.
> Perhaps it is possible that the page locater is taking longer than 5 seconds to find the logout button.

> One solution you can try; from https://playwright.dev/docs/test-timeouts:
> await expect(logoutLink).toBeVisible({ timeout: 10000 });
> To increase the timeout to a large enough value that it might be able to pick it up.
> Await can take a really long time especially looking for elements so I think this would be a good first step. 


So i came up with this assumption of increasing the timeout in our `playwright.config.ts` file like this:
> expect: {
>     /**
>      * Increase the default timeout for snippet tests because they're not deterministic.
>      */
>     timeout: testType === 'snippet' ? 60000 : 5000,
>   },

The comments even mentioned that we can increase the default timeout for snippet tests. But we didn't have this issue until the day before yesterday, so we had to change. This is just one assumption.

samijaber (approved):

PR #3969, opened 2025-03-13
feat: ENG-8599 Add title attribute for images

midhunadarvin:
## Description

- add `title` attribute for images in gen 1 and gen 2 sdks

__Corresponding webapp PR__ :
https://github.com/BuilderIO/builder-internal/pull/8124

**Jira Ticket**
https://builder-io.atlassian.net/browse/ENG-8599

**Loom**
https://www.loom.com/share/2ebd32ee8c874332af2f49a2eea8e5ef

### DIFF
--- a/.changeset/forty-masks-serve.md
+++ b/.changeset/forty-masks-serve.md
@@ -4,7 +4,6 @@
 '@builder.io/sdk-react-nextjs': patch
 '@builder.io/sdk-qwik': patch
 '@builder.io/sdk-react': patch
-'@builder.io/sdk-react-native': patch
 '@builder.io/sdk-solid': patch
 '@builder.io/sdk-svelte': patch
 '@builder.io/sdk-vue': patch
--- a/.changeset/forty-masks-serve.md
+++ b/.changeset/forty-masks-serve.md
@@ -1,0 +1,14 @@
+---
+'@builder.io/react': patch
+'@builder.io/sdk': patch
+'@builder.io/sdk-angular': patch
+'@builder.io/sdk-react-nextjs': patch
+'@builder.io/sdk-qwik': patch
+'@builder.io/sdk-react': patch
+'@builder.io/sdk-react-native': patch
+'@builder.io/sdk-solid': patch
+'@builder.io/sdk-svelte': patch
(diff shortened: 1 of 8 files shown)


sidmohanty11 (approved):

.changeset/forty-masks-serve.md
```diff
@@ -0,0 +1,13 @@
+---
+'@builder.io/react': patch
+'@builder.io/sdk-angular': patch
+'@builder.io/sdk-react-nextjs': patch
+'@builder.io/sdk-qwik': patch
+'@builder.io/sdk-react': patch
+'@builder.io/sdk-react-native': patch
```
sidmohanty11, line 7: oh, for react native we use this override: https://github.com/sidmohanty11/builder/blob/general-support-vc/packages/sdks/overrides/react-native/src/blocks/image/image.tsx/#L29 can you please check if this supports title? else we can remove this addition from react-native

midhunadarvin: Done

packages/sdks/src/blocks/img/img.lite.tsx
```diff
@@ -18,6 +18,7 @@ export interface ImgProps {
   imgSrc?: string; // TODO(misko): I think this is unused
   image?: string;
   altText?: string;
+  title?: string;
```
samijaber, line 21: this is the raw `img` component. it's different from `image`. I don't think you actually want to add it here? If you did, then:

- you need to update its componentInfo file
- you need to update the `Raw:Img` in gen1 react as well

samijaber (commented):
can we throw in a quick test? add `title` to an image and use https://playwright.dev/docs/api/class-locator#locator-get-by-title to check if it's set properly

packages/sdks/src/blocks/img/img.lite.tsx
midhunadarvin: removed the change in this file

midhunadarvin:
> can we throw in a quick test? add title to an image and use https://playwright.dev/docs/api/class-locator#locator-get-by-title to check if it's set properly

Added test

PR #3971, opened 2025-03-13
EDU-Hydrogen: Custom components - child blocks 

uttej-vsk:
## Description

This is the final custom component for the child blocks PR. With it, we can update the custom components—child blocks doc.

@samijaber Please review this as well.

### DIFF
--- a/packages/sdks/snippets/hydrogen/app/components/CustomHero.tsx
+++ b/packages/sdks/snippets/hydrogen/app/components/CustomHero.tsx
@@ -1,8 +1,8 @@
 // app/components/CustomHero.tsx
-import React from 'react';
+import {type PropsWithChildren} from 'react';
 import type {RegisteredComponent} from '@builder.io/sdk-react';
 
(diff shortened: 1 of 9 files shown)


packages/sdks/snippets/hydrogen/app/components/CustomHero.tsx
```diff
@@ -0,0 +1,27 @@
+// app/components/CustomHero.tsx
+import React from 'react';
+import type {RegisteredComponent} from '@builder.io/sdk-react';
+
+export function CustomHero({children}: {children?: React.ReactNode}) {
```
samijaber, line 5: React has a helper type for this called `PropsWithChildren`:
```suggestion
import React, {  type PropsWithChildren } from 'react';
import type {RegisteredComponent} from '@builder.io/sdk-react';

export function CustomHero({children}: PropsWithChildren) {
```

samijaber (approved):

PR #3973, opened 2025-03-17
fix [editor] : ENG-8585 Change in column direction

shubham-builder:
## Description

Columns changes direction with change in screen width

LOOM - https://www.loom.com/share/2c21262c622642679c14502b3bc30112

### DIFF
--- a/packages/react/src/blocks/Columns.tsx
+++ b/packages/react/src/blocks/Columns.tsx
@@ -99,17 +99,7 @@ class ColumnsComponent extends React.Component<any> {
         <div
           className="builder-columns"
           css={{
-            display: 'flex',
-            ...(this.props.stackColumnsAt !== 'never' && {
-              [`@media (max-width: ${
-                this.props.stackColumnsAt !== 'tablet'
-                  ? breakpointSizes.small.max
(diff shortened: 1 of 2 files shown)


packages/react/src/blocks/Columns.tsx
```diff
@@ -99,17 +99,7 @@ class ColumnsComponent extends React.Component<any> {
         <div
           className="builder-columns"
           css={{
-            display: 'flex',
-            ...(this.props.stackColumnsAt !== 'never' && {
-              [`@media (max-width: ${
-                this.props.stackColumnsAt !== 'tablet'
```
midhunadarvin, line 105: I think this might not be the proper fix. The `stackColumnsAt` property can be either `tablet` or `mobile` .  So as per this code , the media query makes the flex direction to either `column` or `column-reverse` when the screen width is less than tablet / mobile max sizes. So it changes from row orientation to column orientation when the screen size changes to below the max width. 

The code here looks proper and removing it might be affecting `stackColumnsAt` feature. Could you debug some more on how the values given here are changing with respect to the screen size ?

midhunadarvin (changes requested):

PR #3977, opened 2025-03-19
fix[angular-gen2]: new blocks getting added on top in case of components that use children

sidmohanty11:
## Description

_Screenshot_

### DIFF
--- a/packages/sdks-tests/src/e2e-tests/editing.spec.ts
+++ b/packages/sdks-tests/src/e2e-tests/editing.spec.ts
@@ -19,6 +19,7 @@ import { ADD_A_TEXT_BLOCK } from '../specs/duplicated-content-using-nested-symbo
 import { EDITING_STYLES } from '../specs/editing-styles.js';
 import { ACCORDION_WITH_NO_DETAIL } from '../specs/accordion.js';
 import { NEW_BLOCK_ADD, NEW_BLOCK_ADD_2 } from '../specs/new-block-add.js';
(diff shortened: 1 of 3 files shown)


packages/sdks/src/components/block/components/component-ref/component-ref.lite.tsx
```diff
@@ -28,13 +29,52 @@ useMetadata({
...
   });
+
+  onUpdate(() => {
+    useTarget({
+      angular: () => {
+        if (
+          // @ts-expect-error - 'changes' comes from Angular's ngOnChanges hook
+          !changes['blockChildren']?.isFirstChange() &&
+          // @ts-expect-error - 'changes' comes from Angular's ngOnChanges hook
+          JSON.stringify(changes['blockChildren']?.previousValue) !==
```
samijaber, line 48: isn't this a massive cost, stringifying the entire block children on _every_ `ngOnChanges`?

samijaber: any way it can be avoided?

sidmohanty11: yup, updated the logic to only check for changes in blockChildren, had to add the previousValue && currVal !== 0 as getter will send a new empty array every time the change detection runs

loom of the approach:
https://www.loom.com/share/5c3c3952775442099ef5dc6ba626d340

samijaber (approved):

PR #3983, opened 2025-03-21
Fix: NextJS SDK: reduce network roundtrips for visual editing

samijaber:
## Description

https://www.loom.com/share/f1c0c50adb4c4566a00caeedd40f8fc5

- replace `router.refresh()` with `revalidateTag()` to avoid a redundant network->client->network roundtrip
- you have to specify the path, so i decided to restrict it to the current `location.url.pathname` to reduce its impact as much as possible

before:
- client: receives new data from editor
- client: POSTs new content to server action
- server: stores new content in cache
- client: calls `router.refresh()`
- server: re-renders page

after
- client: receives new data from editor
- client: POSTs new content to server action
- server: stores new content in cache and calls `revalidateTag()`, which re-renders the page

### DIFF
--- a/packages/sdks/overrides/rsc/src/helpers/preview-lru-cache/set.ts
+++ b/packages/sdks/overrides/rsc/src/helpers/preview-lru-cache/set.ts
@@ -1,5 +1,6 @@
 'use server';
 
+import { revalidatePath } from 'next/cache';
 import type { BuilderContent } from '../../types/builder-content.js';
 import { init } from './init.js';
 import type { GlobalWCache } from './types.js';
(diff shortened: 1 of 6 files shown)


packages/sdks/overrides/rsc/src/helpers/preview-lru-cache/set.ts
```diff
@@ -1,19 +1,22 @@
 'use server';
 
+import { revalidatePath } from 'next/cache';
```
sidmohanty11, line 3: @samijaber seems like the build is failing, did an initial research and found out that:
- we need to externalize `next/cache` so that it's not required to be bundled inside the build similar to `next/navigation`: https://github.com/BuilderIO/builder/blob/main/packages/sdks/output/nextjs/vite.config.ts#L45
- add a declaration module for types to export from next/cache.js for the `.js` implicit extension workaround: https://github.com/BuilderIO/builder/blob/main/packages/sdks/output/nextjs/module-augmentations.d.ts#L5

samijaber: Thanks for the help!

sidmohanty11 (approved):
🔥

PR #3985, opened 2025-03-24
fix[ENG-8003]: Changing the entry on a symbol does not update the preview window

yash-builder:
## Description

This PR fixes an issue where symbol content was not updating correctly. The fix compares the previous symbol entry with the incoming symbol entry—if they are not equal, it triggers `fetchSymbolContent` to ensure the latest content is loaded.  

**Changes Made:**  
- Added a comparison check between the previous and incoming symbol entries.  
- Triggered `fetchSymbolContent` if the entries differ to update the content.  

**Why This Change Was Made:**  
To ensure symbol content updates correctly when the symbol entry changes, improving accuracy and consistency in the preview.

_Loom_
https://www.loom.com/share/032123ec4a7f449db7690eb9ca022f3e?sid=3964bfa8-ac48-428a-a376-91568c447213


sidmohanty11 (commented):
you're missing a few things, i created a loom explaining what all changes you need to do so that you can properly mock the fetch calls

https://www.loom.com/share/62e673c3668342949f1df71b72b921f3

let me know if you have any questions

packages/sdks-tests/src/e2e-tests/editing.spec.ts
```diff
@@ -825,3 +829,45 @@ test.describe('Visual Editing', () => {
...
+  const newContent = cloneContent(MAIN_CONTENT);
+
+  await page.frameLocator('iframe').getByText('Green Potato').waitFor();
+
+  newContent.data.blocks[0].component.options.symbol.entry = 'aa024e6851e94b49b99f41a2294fd423';
+  await sendContentUpdateMessage({ page, newContent, model: 'page' });
+
+  await page.frameLocator('iframe').getByText('Red tomato').waitFor();
+  expect(fetchCount).toBe(2);
+});
```
samijaber, line 873: This test fails because the Symbol API call is made twice instead of once. It's an issue somewhere in our reactivity that we haven't completely fixed yet.

You're checking for the 1st VS 2nd fetch which is why you get the wrong symbol. Instead, check the `query.id` param received in the request to decide on the entry and it will work.

```suggestion
test('Symbol should update the data when nested values are updated', async ({
  page,
  sdk,
  packageName,
  basePort,
}) => {
  test.skip(excludeGen1(sdk));
  test.skip(packageName === 'nextjs-sdk-next-app');

  let fetchCount = 0;
  const urlMatch = /https:\/\/cdn\.builder\.io\/api\/v3\/content\/symbol/;

  await page.route(urlMatch, route => {
    fetchCount++;
    const url = new URL(route.request().url());
    const queryIdParam = url.searchParams.get('query.id');
    let symbolToReturn = null;

    if (queryIdParam === '"e2a166f7d9544ed9ade283abf9491af3"') {
      symbolToReturn = GET_CONTENT_SYMBOL_UPDATE_ENTRY_ONE;
    } else if (queryIdParam === '"aa024e6851e94b49b99f41a2294fd423"') {
      symbolToReturn = GET_CONTENT_SYMBOL_UPDATE_ENTRY__TWO;
    } else {
      throw new Error(`Unknown query id: ${queryIdParam}`);
    }

    return route.fulfill({
      status: 200,
      json: {
        results: [symbolToReturn],
      },
    });
  });

  await launchEmbedderAndWaitForSdk({ path: '/symbol-update-entries', basePort, page, sdk });

  const newContent = cloneContent(MAIN_CONTENT);

  await page.frameLocator('iframe').getByText('Green Potato').waitFor();

  newContent.data.blocks[0].component.options.symbol.entry = 'aa024e6851e94b49b99f41a2294fd423';
  await sendContentUpdateMessage({ page, newContent, model: 'page' });

  await page.frameLocator('iframe').getByText('Red tomato').waitFor();
  expect(fetchCount).toBe(2);
})
```

packages/sdks-tests/src/specs/index.ts
```diff
@@ -93,6 +93,7 @@ import { DYNAMIC_BUTTON } from './dynamic-button.js';
 import { SYMBOLS_WITH_LIST_CONTENT_INPUT } from './symbols-with-list-content-input.js';
 import { COLUMNS_VERTICAL_CENTERING } from './columns-vertical-centering.js';
 import { SECTION_CHILDREN } from './section-children.js';
+import { MAIN_CONTENT } from './get-content-symbol-update-entry.js';
```
sidmohanty11, line 96: we can follow the naming convention, keep this same as the route

```suggestion
import { MAIN_CONTENT as SYMBOL_UPDATE_ENTRIES } from './get-content-symbol-update-entry.js';
```

sidmohanty11 (commented):
LGTM! requires a changeset

sidmohanty11 (approved):

PR #3987, opened 2025-03-24
Range Slider Plugin

mix3d:
Adds a range slider input that mimics the Built-in private one for Image Aspect Ratios. 

See this thread: https://builder-io.atlassian.net/browse/PLAN-364?focusedCommentId=27085

Note: while some initial requests included the ability to set min/max values, AFAIK we cannot do that with plugins today.

### DIFF
--- a/plugins/range-input/.gitignore
+++ b/plugins/range-input/.gitignore
@@ -1,0 +1,12 @@
+node_modules
+dist
+compiled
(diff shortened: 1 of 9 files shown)


paprikaf (approved):
LGTM

paprikaf (approved):

PR #4013, opened 2025-04-07
feat[qwik-gen2]: support of variant containers

sidmohanty11:
## Description

Support of Variant Containers in Qwik SDK.

Jira
https://builder-io.atlassian.net/browse/ENG-8844

Continued from https://github.com/sidmohanty11/builder/pull/2

_Screenshot_

### DIFF
--- a/packages/sdks-tests/src/e2e-tests/personalization-container.spec.ts
+++ b/packages/sdks-tests/src/e2e-tests/personalization-container.spec.ts
@@ -4,8 +4,7 @@ import { isSSRFramework, test } from '../helpers/index.js';
 import { launchEmbedderAndWaitForSdk } from '../helpers/visual-editor.js';
 import type { Sdk } from '../helpers/sdk.js';
 
-const SELECTOR = 'div[builder-content-id]';
-const SDKS_SUPPORTING_PERSONALIZATION = ['react', 'vue', 'svelte'] as Sdk[];
+const SDKS_SUPPORTING_PERSONALIZATION = ['react', 'vue', 'svelte', 'qwik'] as Sdk[];
 
 const createContextWithCookies = async ({
   cookies,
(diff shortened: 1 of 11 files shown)


samijaber (approved):

packages/sdks-tests/src/e2e-tests/personalization-container.spec.ts
```diff
@@ -166,8 +169,8 @@ test.describe('Personalization Container', () => {
 
     await page.goto('/variant-containers');
 
-    await expect(page.getByText('My default content')).toBeVisible();
-    await expect(page.getByText('Default content 2')).toBeVisible();
+    await expect(page.getByText('My default content').locator('visible=true')).toBeVisible();
```
yash-builder, line 172: hey is there a reason why we added `.locator('visible=true')`? Because `toBeVisible()` should be doing the same job

packages/sdks/e2e/qwik-city/src/routes/[...index]/index.tsx
```diff
@@ -80,6 +84,18 @@ export const useBuilderContentLoader = routeLoader$(async (event) => {
 
 export default component$(() => {
   const contentProps = useBuilderContentLoader();
+
+  useOnDocument(
+    'qinit',
+    $(() => {
+      if (window.location.pathname === '/variant-containers/') {
+        setClientUserAttributes({
```
Comment (yash-builder) [inline packages/sdks/e2e/qwik-city/src/routes/[...index]/index.tsx:92]: what does `setClientUserAttributes` do?

Comment (sidmohanty11) [reply packages/sdks/e2e/qwik-city/src/routes/[...index]/index.tsx:92]: `setClientUserAttributes` sets cookies in the browser and also triggers a subscriber call that updates the Personalization Containers and shows the winning variant wrt to the cookie update

packages/sdks-tests/src/e2e-tests/personalization-container.spec.ts
sidmohanty11: great question, in qwik: https://qwik.dev/docs/components/slots/#unprojected-content there is a concept of unprojected content - keeps content at hand inside a <q:template hidden > that may be required in certain future
so as we are using hydration reset approach - this now has two contents now pointing to the same text in case of the default variant which resolves to the incorrect text element in the DOM.

This (`page.getByText('My default content')`) resolves to two locators and we only want to look for the locator that is visible

PR #4016, opened 2025-04-08
fix: Skipping flakey test for Qwik - CI pipeline

yash-builder:
## Description

This PR skips the test:  
`[qwik-city] › editing.spec.ts:552:5 › Visual Editing › New Block addition and deletion › should add new block in the middle`

The test is failing in CI but works fine locally. I'm skipping it for now to unblock the pipeline while we investigate the root cause.


sidmohanty11 (commented):
@yash-builder would it be possible to push ~10 empty commits in this PR to check that CI passes everytime so that we're sure that nothing else is flaky?

```bash
git commit --allow-empty -m "Trigger CI"
```

samijaber (approved):

PR #4017, opened 2025-04-09
fix[webapp][editor]:ENG-8571 add registerAction handler

shubham-builder:
## Description

Added postMessage on registerAction() call with the action's data.

Link to JIRA ticket:
https://builder-io.atlassian.net/browse/ENG-8571

_Screenshot_
LOOM - https://www.loom.com/share/948fa21e285e4cbab1d6b1667fd99dde

CMS pr link - https://github.com/BuilderIO/builder-internal/pull/8316

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -882,7 +882,17 @@ export interface Action {
   name: string;
   inputs?: readonly Input[];
   returnType?: Input;
-  action: Function | string;
+  action: (options: Record<string, any>) => string;
+   /**
+   * Is an action for expression (e.g. calculating a binding like a formula
+   * to fill a value based on locale) or a function (e.g. something to trigger
+   * on an event like add to cart) or either (e.g. a custom code block)
+   */
+   kind: 'expression' | 'function' | 'any';
+    /**
+   * Globally unique ID for an action, e.g. "@builder.io:setState"
+   */
+  id: string;
 }
 
 export class Builder {
@@ -984,6 +994,20 @@ export class Builder {
 
   static registerAction(action: Action) {
     this.actions.push(action);
+
+    if (Builder.isBrowser) {
+      const actionClone = JSON.parse(JSON.stringify(action));
+      if(action.action) {
+        actionClone.action = action.action.toString();
+      }
+      window.parent?.postMessage(
+        {
+          type: 'builder.registerAction',
+          data: actionClone,
+        },
+        '*'
+      );
+    }
   }
 
   static registerTrustedHost(host: string) {
--- a/packages/sdks-tests/src/e2e-tests/custom-action.spec.ts
+++ b/packages/sdks-tests/src/e2e-tests/custom-action.spec.ts
@@ -1,4 +1,4 @@
-import { test,excludeGen2 } from '../helpers/index.js';
+import { test, excludeGen2 } from '../helpers/index.js';
 import { launchEmbedderAndWaitForSdk } from '../helpers/visual-editor.js';
 
 
@@ -9,13 +9,10 @@ test.describe('Custom actions', () => {
   test('correctly captures registering of custom action',  async ({
     page,
     basePort,
-    packageName,
     sdk,
   }) => {
     test.skip(excludeGen2(sdk));
-    test.skip(
-      packageName !== 'gen1-react'
-    );
+    
     const customActionMsgPromise = page.waitForEvent('console', msg =>
        msg.text().includes(CUSTOM_ACTION_LOADED_MESSAGE)
     );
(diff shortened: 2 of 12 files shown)


packages/core/src/builder.class.ts
```diff
@@ -882,7 +882,17 @@ export interface Action {
...
+   /**
+   * Is an action for expression (e.g. calculating a binding like a formula
+   * to fill a value based on locale) or a function (e.g. something to trigger
+   * on an event like add to cart) or either (e.g. a custom code block)
+   */
+   kind: 'expression' | 'function' | 'any';
+    /**
+   * Globally unique ID for an action, e.g. "@builder.io:setState"
+   */
+  id: string;
```
midhunadarvin, line 895: ❔**question:** ‏
where is the `kind` and `id` being used ?

packages/core/src/builder.class.ts
```diff
@@ -984,6 +994,20 @@ export class Builder {
 
   static registerAction(action: Action) {
     this.actions.push(action);
+
+    if (Builder.isBrowser) {
+      const actionClone = JSON.parse(JSON.stringify(action));
+      if(action.action) {
+        actionClone.action = action.action.toString();
```
midhunadarvin, line 1001: In the `Action` type, the `action` property is a function type, but here we are setting it as a string. Is this intended ?

packages/core/src/builder.class.ts
```diff
@@ -882,7 +882,17 @@ export interface Action {
   name: string;
   inputs?: readonly Input[];
   returnType?: Input;
-  action: Function | string;
+  action: (options: Record<string, any>) => string;
```
midhunadarvin, line 885: nitpick: Could we give a better name for this property because both interface and the property are having the same name ?

shubham-builder: This name is already being used by client so, didn't wanted to change that part.

packages/core/src/builder.class.ts
shubham-builder: Kind is being used to filter the type of action and id is used for uniquely identifying each actions.

packages/core/src/builder.class.ts
shubham-builder: Yes... function cannot be passed in a postMessage()

yash-builder: Can you try these instead?
```suggestion
    const actionClone = {
      ...action,
      action: action.action?.toString(),
    };
```

packages/core/src/builder.class.ts
```diff
@@ -984,6 +994,20 @@ export class Builder {
 
   static registerAction(action: Action) {
     this.actions.push(action);
+
+    if (Builder.isBrowser) {
+      const actionClone = JSON.parse(JSON.stringify(action));
```
yash-builder, line 999: Can you try this instead?
```suggestion
      const actionClone = {
         ...action,
         action: action.action?.toString(),
    };
```

Reason: It's a performance hit for large objects

shubham-builder: @yash-builder This is shallow clone. It will cause issue for the postMessage.

clyde-builderio:
@shubham-builder would it be possible to add a test for this similar to `packages/sdks-tests/src/e2e-tests/custom-components.spec.ts`

clyde-builderio:
@shubham-builder  `packages/sdks/snippets/gen1-react/src/routes/custom-child.tsx` may be a good reference for a snippet-test
Here's a loom explaining how we could go about it - https://www.loom.com/share/1c7253414a774e09811d8a096e0854fc

packages/sdks-tests/src/e2e-tests/custom-action.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+  test('correctly captures registering of custom action',  async ({
+    page,
+    basePort,
+    packageName,
+    sdk,
+  }) => {
+    test.skip(excludeGen2(sdk));
+    test.skip(
+      packageName !== 'gen1-react'
+    );
```
clyde-builderio, line 18: Why not the other gen1 ?

shubham-builder: "registerAction" action is done only in this package.

shubham-builder: Let me check for all packages

clyde-builderio: You can implement the same for the other gen1-* too
Same for the snippet test

clyde-builderio (approved):

PR #4022, opened 2025-04-11
WIP. feat[RSC]: [ENG-8936] Add `LiveEdit` support for client-side updates using `BuilderContext.Provider`

yash-builder:
**Description:**  
This PR introduces client-side live editing support by extending the existing context and rendering logic in the SDK.

**What changes were made:**  
- Added a new `BuilderContext.Provider` inside `EnableEditor` to consume a derived state from the `builderContextSignal` prop in `Content`.
- Introduced a `LiveEdit` wrapper component that:
  - Finds a block by ID from the current context content.
  - Applies updated component options dynamically to render changes in real time.
- Updated `InteractiveElement` to conditionally render `LiveEdit` when in editing mode.
- Implemented logic to differentiate between `editType: 'server'` and `editType: 'client'`:
  - For `'server'`, retain existing behavior—push updates to LRU cache and re-render on the server.
  - For `'client'`, trigger a context update that re-renders `LiveEdit` components on the client side.

**Why these changes were made:**  
To support hybrid editing in the SDK where developers can opt into either server-side or client-side editing behavior, depending on the use case.

**Additional context:**  
- These changes are foundational for enabling smoother, near-instant updates during visual editing sessions—especially when using `editType: 'client'`.
- These changes will also require handling on `builder-internal`

_Screenshot_
OTW

### DIFF
--- a/packages/core/src/builder.class.ts
+++ b/packages/core/src/builder.class.ts
@@ -770,6 +770,11 @@ export interface Component {
   models?: readonly string[];
 
   /**
+   * Whether the component is a React Server Component
+   */
+  isRSC?: boolean;
+
+  /**
    * Specify restrictions direct children must match
    */
   childRequirements?: {
--- a/packages/sdks/src/components/content/components/enable-editor.lite.tsx
+++ b/packages/sdks/src/components/content/components/enable-editor.lite.tsx
@@ -28,7 +28,7 @@ import { getInteractionPropertiesForEvent } from '../../../functions/track/inter
 import { getDefaultCanTrack } from '../../../helpers/canTrack.js';
 import { getCookieSync } from '../../../helpers/cookie.js';
 import { postPreviewContent } from '../../../helpers/preview-lru-cache/set.js';
-import { createEditorListener } from '../../../helpers/subscribe-to-editor.js';
+import { createEditorListener, type EditType } from '../../../helpers/subscribe-to-editor.js';
 import {
   registerInsertMenu,
   setupBrowserForEditing,
@@ -72,6 +72,7 @@ export default function EnableEditor(props: BuilderEditorProps) {
    */
   const elementRef = useRef<HTMLDivElement>();
   const [hasExecuted, setHasExecuted] = useState<boolean>(false);
+  const [contextValue, setContextValue] = useState<any>(props.builderContextSignal.value);
   const state = useStore({
     mergeNewRootState(newData: Dictionary<any>) {
       const combinedState = {
@@ -85,7 +86,7 @@ export default function EnableEditor(props: BuilderEditorProps) {
         props.builderContextSignal.value.rootState = combinedState;
       }
     },
-    mergeNewContent(newContent: BuilderContent) {
+    mergeNewContent(newContent: BuilderContent, editType?: EditType) {
       const newContentValue = {
         ...props.builderContextSignal.value.content,
         ...newContent,
@@ -104,11 +105,17 @@ export default function EnableEditor(props: BuilderEditorProps) {
 
       useTarget({
         rsc: () => {
-          postPreviewContent({
-            value: newContentValue,
-            key: newContentValue.id!,
-            url: window.location.pathname,
-          });
+          console.log('DEBUG: SDK: RECEIEVED editType: ', editType);
+
+          if (editType === 'server') {
+            postPreviewContent({
+              value: newContentValue,
+              key: newContentValue.id!,
+              url: window.location.pathname,
+            });
+          } else {
+            setContextValue({...contextValue, content: newContentValue});
+          }
         },
         default: () => {
           props.builderContextSignal.value.content = newContentValue;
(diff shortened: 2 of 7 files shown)


packages/sdks/mitosis.config.cjs
```diff
@@ -949,7 +949,7 @@ const QWIK_FORCE_RENDER_COUNT_FOR_RENDERING_CUSTOM_COMPONENT_DEFAULT_VALUE =
     json: {
       post: (json) => {
         if (json.name === 'InteractiveElement') {
-          json.children[0].meta.else.bindings['key'] = {
+          json.children[0].meta.else.meta.else.bindings['key'] = {
```
clyde-builderio, line 952: Would it be possible to record a loom to explain this bit?

packages/sdks/src/components/live-edit.lite.tsx
```diff
@@ -0,0 +1,68 @@
...
+      for (const block of blocks) {
+        if (block.id === id) return block;
+
+        if (block.children) {
+          const child = this._findBlockById(block.children, id);
+          if (child) return child;
+        }
+      }
+      return null;
+    },
```
clyde-builderio, line 37: We could use just `findBlockById()` instead of having a separate `_findBlockById()`. Also by removing `_findBlockById()` we reduce the future possibility of a developer calling `state._findBlockById()`

clyde-builderio: If `_findBlockById()` was a private function in `useStore` then we could keep it.

packages/sdks/src/components/live-edit.lite.tsx
```diff
@@ -0,0 +1,68 @@
...
+  onMount(() => {
+    useTarget({
+      angular: () => {
+        /** this is a hack to include unused props */
+        const _ = {
+          a: props.id,
+          b: props.Wrapper,
+          c: props.attributes,
+          d: props.children,
+        };
```
clyde-builderio, line 58: Would it be possible to record a loom to explain this part?

packages/sdks/src/components/content/components/enable-editor.lite.tsx
```diff
@@ -104,11 +105,17 @@ export default function EnableEditor(props: BuilderEditorProps) {
 
       useTarget({
         rsc: () => {
-          postPreviewContent({
-            value: newContentValue,
-            key: newContentValue.id!,
-            url: window.location.pathname,
-          });
+          // console.log('DEBUG: SDK: RECEIEVED editType: ', editType);
```
clyde-builderio, line 108: can remove this

packages/sdks/src/components/content/components/enable-editor.lite.tsx
```diff
@@ -242,7 +249,7 @@ export default function EnableEditor(props: BuilderEditorProps) {
     },
   });
 
-  setContext(builderContext, props.builderContextSignal);
+  setContext(builderContext, props.builderContextSignal)
```
clyde-builderio, line 252: nitpick but we can add the ; back

clyde-builderio:
@yash-builder Would it be possible to add tests for the changes?

packages/core/src/builder.class.ts
```diff
@@ -769,6 +769,11 @@ export interface Component {
    */
   models?: readonly string[];
 
+  /**
+   * Whether the component is a React Server Component
+   */
+  isRSC?: boolean;
```
samijaber, line 775: On second thought i would prefer we don't have this type here. Some customers might see this and set it to `true` in their React Gen1 SDK implementation, and then they'll be confused as to why it doesn't do anything

samijaber (commented):
with this change, many existing visual editing tests (like the ones that edit text blocks) should now work on nextjs SDK.

- For `'client'` updates (like editing texts, images): Can we enable those tests for nextjs sdk and make any other changes needed to confirm that editing works as expected?
- For `'server'` updates (like editing Columns, which is an RSC), can we dig into getting those tests working for nextjs SDK too to make sure the server updates also work as expected? I never got around to doing so, but it would be useful as part of this work to make sure all editing works well going forward.

samijaber (commented):
Another note: would be good to see & test whole flow end-to-end for both PRs (webapp + SDK) combined and make sure it works well before merging SDK changes, otherwise those might themselves in an SDK release if someone else has changes queued up.

packages/sdks/mitosis.config.cjs
yash-builder: Noted! Will surely record a loom around it before merging this PR

packages/sdks/src/components/live-edit.lite.tsx
yash-builder: Noted!

sanyamkamat (approved):
Approving PR so that it unblocks other tasks related to RSC epic

PR #4044, opened 2025-04-25
Handle null origin

STRd6:
https://builder-io.atlassian.net/browse/ENG-9229

### DIFF
--- a/.changeset/proud-moles-punch.md
+++ b/.changeset/proud-moles-punch.md
@@ -1,0 +1,5 @@
+---
+"@builder.io/sdk": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/proud-moles-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@builder.io/sdk": patch
```
samijaber, line 2: so the release note shows up in the React CHANGELOG.md too:

```suggestion
"@builder.io/sdk": patch
"@builder.io/react": patch
```

samijaber (approved):

PR #4049, opened 2025-04-29
fix[gen2][bindings]: ENG-8483 Fix for localized bindings not working

anaghav2023:
## Description
If we have bindings for a localized text, they don't work as expected.

https://www.loom.com/share/e6e4ce4c9a784a7da01e15d058c10d17?focus_title=1&muted=1&from_recorder=1

https://builder-io.atlassian.net/browse/ENG-8483

**Loom for solution**
https://www.loom.com/share/562dfc3a2994443cb72aaa9fcb97dc26

**Test for Angular Gen2**
https://www.loom.com/share/d062f76dff4d49af8c78f0cfcc8d548e

### DIFF
--- a/.changeset/perfect-pandas-hunt.md
+++ b/.changeset/perfect-pandas-hunt.md
@@ -1,0 +1,13 @@
+\---
+"@builder.io/sdk-angular": patch
+"@builder.io/sdk-react-nextjs": patch
+"@builder.io/sdk-qwik": patch
+"@builder.io/sdk-react": patch
+"@builder.io/sdk-react-native": patch
+"@builder.io/sdk-solid": patch
+"@builder.io/sdk-svelte": patch
+"@builder.io/sdk-vue": patch
+
+---
+
+fix: localized data bindings issue
--- a/.changeset/perfect-pandas-hunt.md
+++ b/.changeset/perfect-pandas-hunt.md
@@ -1,13 +1,12 @@
-\---
-"@builder.io/sdk-angular": patch
-"@builder.io/sdk-react-nextjs": patch
-"@builder.io/sdk-qwik": patch
-"@builder.io/sdk-react": patch
-"@builder.io/sdk-react-native": patch
-"@builder.io/sdk-solid": patch
-"@builder.io/sdk-svelte": patch
-"@builder.io/sdk-vue": patch
-
+---
+'@builder.io/sdk-angular': patch
+'@builder.io/sdk-react-nextjs': patch
+'@builder.io/sdk-qwik': patch
+'@builder.io/sdk-react': patch
+'@builder.io/sdk-react-native': patch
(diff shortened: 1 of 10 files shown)


packages/sdks/src/blocks/helpers.ts
```diff
@@ -53,3 +53,13 @@ export function setAttrs(
...
+
+export const resolveLocalizedText = (
+  text: string | Record<string, string> | undefined,
+  locale?: string
+): string => {
+  if (typeof text === 'object') {
+    if (locale && text[locale]) return text[locale];
+  }
+  return text?.toString() || '';
+};
```
sidmohanty11, line 65: @anaghav2023 we should already be doing this inside `block.lite.tsx` when we process a block, so text passed to `text.lite.tsx` should always be a string, not sure if the extraction logic is not running in some cases? Here I can see that it's inside a repeat? Does this only happen while using repeat bindings or is it happening for normal bindings with localized fields as well?

https://github.com/BuilderIO/builder/blob/59cf58a0d7c828de1166f4dcadbbaec3c81cecd4/packages/sdks/src/components/block/block.lite.tsx#L94-L102

https://github.com/BuilderIO/builder/blob/59cf58a0d7c828de1166f4dcadbbaec3c81cecd4/packages/sdks/src/functions/get-processed-block.ts#L112-L115

https://github.com/BuilderIO/builder/blob/main/packages/sdks/src/functions/extract-localized-values.ts

anaghav2023: @sidmohanty11 Thanks, I have updated the code now. Added a few tests also.

.changeset/perfect-pandas-hunt.md
```diff
@@ -0,0 +1,13 @@
+\---
```
sidmohanty11, line 1: i think the `\` got added here accidentally, this might result into skipping this changeset. 
```suggestion
---
```

anaghav2023: removed it !

packages/sdks/src/functions/get-processed-block.ts
```diff
@@ -109,18 +107,17 @@ export function getProcessedBlock({
...
     localState,
     rootState,
     rootSetState,
     context,
   });
+  const localizedBlock = resolveLocalizedValues(
+    transformedBlock,
+    rootState.locale as string | undefined
+  );
+  return localizedBlock;
```
sidmohanty11, line 122: minor thing but could we just use `transformedBlock` variable instead of creating a new variable?
```suggestion
  transformedBlock = resolveLocalizedValues(
    transformedBlock,
    rootState.locale as string | undefined
  );
  return transformedBlock;
```

sidmohanty11 (approved):
🙌

PR #4050, opened 2025-04-30
fix[gen1][react]: ENG-8739 change to makeFn() for handling errors when serialising functions

clyde-builderio:
## Description

Changed makeFn() for handling errors when serialising functions

**PR that caused the hydration error**
https://github.com/BuilderIO/builder/pull/3396

**Screenshot**
| Before | After |
| ---- | --- |
| Screenshot 2025-04-30 at 11 19 52 AM | Screenshot 2025-04-30 at 11 19 14 AM |

**Loom**
https://www.loom.com/share/8b961a80726d4fe4ad8bdf52e389f72a

### DIFF
--- a/packages/react/src/functions/string-to-function.ts
+++ b/packages/react/src/functions/string-to-function.ts
@@ -209,7 +209,12 @@ export const makeFn = (code: string, useReturn: boolean, args?: string[]) => {
             }
             const val = obj.getSync(key);
             if (typeof val?.copySync === 'function') {
+              try {
                 return JSON.parse(stringify(val));
+              } catch (e) {
+               log('Error:', e);
+                return refToProxy(val);
+              }
             }
             return val;
         },


packages/react/src/functions/string-to-function.ts
```diff
@@ -209,7 +212,12 @@ export const makeFn = (code: string, useReturn: boolean, args?: string[]) => {
             }
             const val = obj.getSync(key);
             if (typeof val?.copySync === 'function') {
+              try {
                 return JSON.parse(stringify(val));
+              } catch (e) {
+                log('Error:', e);
+                return refToProxy(val);
+              }
```
sidmohanty11, line 220: i will let @teleaziz or @samijaber review this PR. I am not fully grasping this change, I am thinking if instead of a try catch we should handle passing of functions correctly. I can see that here we return the value itself if the obj is not an object:
https://github.com/BuilderIO/builder/blob/804a4587ccefa385bea9ac5f401d9b5baadd07c3/packages/react/src/functions/string-to-function.ts#L202-L204
and I am guessing that's fixing this issue because we're re-calling it to handle this? and I can see some error logs that you're printing in the server in your loom that fn couldn't get parsed.

What if we check if the value (`const val = obj.getSync(key);`) is not an object then just return it without parsing it?

clyde-builderio: Screenshot 2025-05-02 at 4 05 53 PM

Tried this. I am getting hydration-error
`typeof val` will always be object as `val` is a [`Reference`](https://github.com/laverdet/isolated-vm?tab=readme-ov-file#class-reference-transferable)

Screenshot 2025-05-02 at 4 05 37 PM

PR #4054, opened 2025-05-02
fix[gen1][core]: ENG-9321 only serialize fn if not plugin and not onSave()

clyde-builderio:
## Description

JTI is having an issue when trying to save the Magento plugin 

**Root cause of the regression**
https://github.com/BuilderIO/builder/pull/3492

**JIRA ticket**
https://builder-io.atlassian.net/browse/ENG-9321

**Loom**
https://www.loom.com/share/738964aade954c5ab15c60b0f0703248

### DIFF
--- a/.changeset/fair-dryers-report.md
+++ b/.changeset/fair-dryers-report.md
@@ -1,0 +1,5 @@
+---
+"@builder.io/sdk": patch
+---
+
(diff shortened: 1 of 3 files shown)


midhunadarvin (approved):
LGTM

PR #4078, opened 2025-05-18
 Snippets: remove all references to getBuilderSearchParams ref to PR #3954

uttej-vsk:
## Description

As per https://builder-io.atlassian.net/browse/ENG-8635?focusedCommentId=28672 we want to remove this helper from our snippets as it's unneeded.

### DIFF
--- a/packages/sdks/snippets/angular-16/src/app/live-preview/live-preview.component.ts
+++ b/packages/sdks/snippets/angular-16/src/app/live-preview/live-preview.component.ts
@@ -2,31 +2,54 @@ import { CommonModule } from '@angular/common';
 import type { OnDestroy, OnInit } from '@angular/core';
 import { Component } from '@angular/core';
 import type { BuilderContent } from '@builder.io/sdk-angular';
-import { subscribeToEditor } from '@builder.io/sdk-angular';
+import {
+  fetchOneEntry,
+  isPreviewing,
+  subscribeToEditor,
+} from '@builder.io/sdk-angular';
 
 @Component({
   standalone: true,
   selector: 'app-live-preview',
   imports: [CommonModule],
   template: `
-    <div *ngIf="content" class="blog-data-preview">
+    <div *ngIf="loading">Loading Data...</div>
+
+    <div *ngIf="!loading && !notFound && content" class="blog-data-preview">
       <div>Blog Title: {{ content.data?.title }}</div>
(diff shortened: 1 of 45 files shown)


packages/sdks/snippets/angular-16/src/app/live-preview/live-preview.component.ts
```diff
@@ -2,31 +2,54 @@ import { CommonModule } from '@angular/common';
...
 
 @Component({
   standalone: true,
   selector: 'app-live-preview',
   imports: [CommonModule],
   template: `
-    <div *ngIf="content" class="blog-data-preview">
+    <div *ngIf="loading">Loading Data...</div>
+
+    <div *ngIf="!loading && !notFound && content" class="blog-data-preview">
```
clyde-builderio: What are your thoughts on this? Would it have the same outcome?
Screenshot 2025-06-10 at 2 40 32 PM

packages/sdks/snippets/qwik-city/src/routes/custom-child/index.tsx
```diff
@@ -1,39 +1,41 @@
...
   });
+
+  return {
+    content,
+    searchParams,
+  };
 });
 
 export default component$(() => {
   const content = useCustomChild();
```
clyde-builderio: could we rename this variable to `customChild`?

packages/sdks/snippets/angular-16/src/app/live-preview/live-preview.component.ts
uttej-vsk: @clyde-builderio I agree on this abstraction. However, we have a redundant check with `content` on the `*ngIf`, which I have removed. Do you think this is better? But if you prefer to abstract this logic to a new function, I would do that. 

Screenshot 2025-06-10 at 11 39 08 AM

packages/sdks/snippets/qwik-city/src/routes/custom-child/index.tsx
uttej-vsk: Sure! I can do this.

packages/sdks/snippets/angular-16/src/app/live-preview/live-preview.component.ts
clyde-builderio: removing `content` from `*ngIf` is cool 👍

clyde-builderio (approved):

PR #4097, opened 2025-06-18
WIP! feat(Raw:Img): Added new inputs for raw:img to handle the metadata

yash-builder:
## Description

_Screenshot_

### DIFF
--- a/packages/sdks/src/blocks/img/component-info.ts
+++ b/packages/sdks/src/blocks/img/component-info.ts
@@ -16,6 +16,97 @@ export const componentInfo: ComponentInfo = {
       allowedFileTypes: IMAGE_FILE_TYPES,
       required: true,
     },
+    {
+      name: 'backgroundSize',
+      type: 'text',
+      defaultValue: 'cover',
+      enum: [
+        {
+          label: 'contain',
+          value: 'contain',
+          helperText: 'The image should never get cropped',
+        },
+        {
+          label: 'cover',
+          value: 'cover',
+          helperText: "The image should fill it's box, cropping when needed",
+        },
+      ],
+    },
+    {
+      name: 'backgroundPosition',
+      type: 'text',
+      defaultValue: 'center',
+      enum: [
+        'center',
+        'top',
+        'left',
+        'right',
+        'bottom',
+        'top left',
+        'top right',
+        'bottom left',
+        'bottom right',
+      ],
+    },
+    {
+      name: 'altText',
+      type: 'string',
+      helperText: 'Text to display when the user has images off',
+    },
+    {
+      name: 'title',
+      type: 'string',
+      helperText: 'Text to display when hovering over the asset',
+    },
+    {
+      name: 'height',
+      type: 'number',
+      hideFromUI: true,
+    },
+    {
+      name: 'width',
+      type: 'number',
+      hideFromUI: true,
+    },
+    {
+      name: 'sizes',
+      type: 'string',
+      hideFromUI: true,
+    },
+    {
+      name: 'srcset',
+      type: 'string',
+      hideFromUI: true,
+    },
+    {
+      name: 'lazy',
+      type: 'boolean',
+      defaultValue: true,
+      hideFromUI: true,
+    },
+    {
+      name: 'fitContent',
+      type: 'boolean',
+      helperText:
+        "When child blocks are provided, fit to them instead of using the image's aspect ratio",
+      defaultValue: true,
+    },
+    {
+      name: 'highPriority',
+      type: 'boolean',
+      advanced: true,
+      helperText:
+        'Mark this image as high priority compared to other images on the page. This prevents lazy loading of the image and tells the browser to load this image before others on the page.',
+    },
+    {
+      name: 'aspectRatio',
+      type: 'number',
+      helperText:
+        "This is the ratio of height/width, e.g. set to 1.5 for a 300px wide and 200px tall photo. Set to 0 to not force the image to maintain it's aspect ratio",
+      advanced: true,
+      defaultValue: 0.7041,
+    },
   ],
   noWrap: true,
   static: true,
--- a/packages/sdks/src/blocks/img/img.lite.tsx
+++ b/packages/sdks/src/blocks/img/img.lite.tsx
@@ -7,6 +7,7 @@ import { getSrcSet } from '../image/image.helpers.js';
  */
 
 import { setAttrs } from '../helpers.js';
+import { JSX } from '@builder.io/mitosis/jsx-runtime';
 
 useMetadata({
   rsc: {
@@ -30,6 +31,7 @@ export interface ImgProps {
     | 'top right'
     | 'bottom left'
     | 'bottom right';
+  aspectRatio?: number;
 }
 
 export default function ImgComponent(props: ImgProps) {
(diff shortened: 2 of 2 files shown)


.changeset/cuddly-pillows-battle.md
```diff
@@ -0,0 +1,6 @@
+---
+"@builder.io/react": patch
+"@builder.io/sdk": patch
```
sidmohanty11, line 3: we need it for gen2 SDKs too

packages/sdks/src/blocks/img/img.lite.tsx
```diff
@@ -55,6 +72,7 @@ export default function ImgComponent(props: ImgProps) {
       style={{
         objectFit: props.backgroundSize || 'cover',
         objectPosition: props.backgroundPosition || 'center',
+        ...state.aspectRatioCss,
```
sidmohanty11: do we need this change in gen1 React Raw Img?

packages/sdks/src/blocks/img/component-info.ts
```diff
@@ -16,6 +16,97 @@ export const componentInfo: ComponentInfo = {
...
+        'Mark this image as high priority compared to other images on the page. This prevents lazy loading of the image and tells the browser to load this image before others on the page.',
+    },
+    {
+      name: 'aspectRatio',
+      type: 'number',
+      helperText:
+        "This is the ratio of height/width, e.g. set to 1.5 for a 300px wide and 200px tall photo. Set to 0 to not force the image to maintain it's aspect ratio",
+      advanced: true,
+      defaultValue: 0.7041,
+    },
```
sidmohanty11: only some of the inputs are getting used inside the RawImg component do we need to accept everything that Image accepts?

packages/sdks/src/blocks/img/img.lite.tsx
yash-builder: Yes, implemented

packages/sdks/src/blocks/img/component-info.ts
yash-builder: removed unnecessary code 👍

packages/sdks/src/blocks/img/img.lite.tsx
```diff
@@ -47,18 +50,34 @@ export default function ImgComponent(props: ImgProps) {
...
+      const aspectRatioStyles = {
+        position: 'absolute',
+        height: '100%',
+        width: '100%',
+        left: '0px',
+        top: '0px',
+      } as const;
+      const out = props.aspectRatio ? aspectRatioStyles : undefined;
+      return out;
+    },
```
sidmohanty11: is it possible to add an e2e test to test all of the inputs, that they're working as expected?

you can copy one of the Image spec example and instead of Image use Raw:Img

packages/sdks/src/blocks/img/component-info.ts
```diff
@@ -2,21 +2,154 @@ import { IMAGE_FILE_TYPES } from '../../constants/file-types.js';
...
+      helperText:
+        "When child blocks are provided, fit to them instead of using the image's aspect ratio",
+      defaultValue: true,
     },
   ],
+
+  shouldReceiveBuilderProps: {
+    builderBlock: true,
+  },
   noWrap: true,
```
sidmohanty11: is the 'Raw:Img' exactly same as the 'Image' now?

yash-builder: TBH yes. I reviewed the [ticket](https://builder-io.atlassian.net/browse/ENG-9527), and the client specifically requested that raw:img align closely with the existing Image component. As of now, they behave similarly.
Let me know if there’s anything I might have overlooked or if you have suggestions for further improvements.

sidmohanty11: should we re-export the Image itself for Raw:Img then? instead of managing them both? tagging @samijaber here

sidmohanty11 (changes requested):
let's ask this in dev-cms first to confirm, if yes just re-export Image for Raw:Img

https://github.com/BuilderIO/builder/pull/4097#discussion_r2163042879

packages/sdks/src/blocks/img/component-info.ts
```diff
@@ -2,21 +2,79 @@ import { IMAGE_FILE_TYPES } from '../../constants/file-types.js';
...
 
   image:
     'https://firebasestorage.googleapis.com/v0/b/builder-3b0a2.appspot.com/o/images%2Fbaseline-insert_photo-24px.svg?alt=media&token=4e5d0ef4-f5e8-4e57-b3a9-38d63a9b9dc4',
+  defaultStyles: {
+    position: 'relative',
+    minHeight: '20px',
+    minWidth: '20px',
+    overflow: 'hidden',
+  },
+  canHaveChildren: true,
```
sidmohanty11: we should remove these right?

defaultStyles will add extra styles to the img component
canHaveChildren as we're not passing it any children builder block

packages/sdks/src/blocks/img/component-info.ts
```diff
@@ -2,21 +2,79 @@ import { IMAGE_FILE_TYPES } from '../../constants/file-types.js';
 import type { ComponentInfo } from '../../types/components.js';
 
 export const componentInfo: ComponentInfo = {
-  // friendlyName?
   name: 'Raw:Img',
+  static: true,
```
sidmohanty11: safe to remove too no?

packages/sdks/src/blocks/img/img.types.ts
```diff
@@ -0,0 +1,34 @@
+import type { BuilderBlock } from "../../types/builder-block.js";
+
+export interface ImgProps {
```
sidmohanty11: delete this file? as we're using the type defined in img.lite.tsx

packages/sdks/src/blocks/img/img.lite.tsx
```diff
@@ -47,29 +49,45 @@ export default function ImgComponent(props: ImgProps) {
...
+        return {};
+      }
+      const { class: _class, className: _className, ...rest } = props.attributes;
+      return rest;
+    },
+    get className() {
+      return ['builder-raw-img', props.attributes?.class, props.attributes?.className]
+        .filter(Boolean)
+        .join(' ');
+    },
```
sidmohanty11: you can use this helper to determine if it's either class or className:
https://github.com/sidmohanty11/builder/blob/main/packages/sdks/src/functions/get-class-prop-name.ts#L3

and i think it's safe to update this to a single state and pass class at the end so that we override it, for ex we can simplify this to:
```ts
 get imgAttrs() {
      return {
        ...useTarget({
          vue: filterAttrs(props.attributes, 'v-on:', false),
          svelte: filterAttrs(props.attributes, 'on:', false),
          default: props.attributes,
        }),
        ...useTarget({
          vue: filterAttrs(props.attributes, 'v-on:', true),
          svelte: filterAttrs(props.attributes, 'on:', true),
          default: {},
        }),
        [getClassPropName()]: `raw-img ${props.attributes[getClassPropName()] || ''}`,
      };
    },
```

packages/react/src/blocks/raw/Img.tsx
```diff
@@ -31,32 +36,110 @@ class ImgComponent extends React.Component<ImgProps> {
...
       <img
         loading="lazy"
-        {...this.props.attributes}
+        {...restAttributes}
         src={this.props.image || attributes.src}
         srcSet={srcset}
+        alt={this.props.altText}
+        title={this.props.title}
+        style={mergedStyle}
+        className="builder-raw-img"
```
sidmohanty11, line 62: will we get any classname from attributes here? (for ex a builder-id?)

sidmohanty11 (changes requested):
added some comments

packages/sdks/src/blocks/img/component-info.ts
yash-builder: yes

sidmohanty11 (approved):
🙌

PR #4102, opened 2025-06-25
fix[gen1][core]: ENG-9774 properly handle query when $-mongo-operator is present for content API

clyde-builderio:
## Description

`query` with `$or` is not part of Content API when we use `builder.get()` but Query API includes `query` with `$or`
It works on Gen-2 SDK which also uses Content API

**JIRA Ticket**
https://builder-io.atlassian.net/browse/ENG-9774

**Steps to test**
1. Clone repo https://github.com/BuilderIO/oncall-repro-cases.
2. Run `cd eng-9774-gen1-vite-app`.
3. change the `package.json` to `link://` the `@builder.io/react` and `@builder.io/sdk` dependencies to this branch.
4. Run `yarn`.
5. Run `yarn dev`.

**Loom**
https://www.loom.com/share/051b8af3d9344c218d678a20d28b0fc4

### DIFF
--- a/.changeset/dry-ghosts-doubt.md
+++ b/.changeset/dry-ghosts-doubt.md
@@ -1,5 +1,0 @@
----
-'@builder.io/sdk': patch
----
-
-Fix: Corrected the conversion of query-objects with $-mongo-operators which are passed to builder.get() with apiEndpoint is "content"


packages/core/src/builder.class.ts
```diff
@@ -2709,11 +2728,17 @@ export class Builder {
...
-        const flattened = this.flattenMongoQuery({ query: queue[0].query });
+        const objectToFlatten = { query: queue[0].query };
+        const flattened = this.flattenMongoQuery(objectToFlatten);
         for (const key in flattened) {
           queryParams[key] = flattened[key];
         }
-        delete queryParams.query;
+
+        if (this.isShallowEqual(objectToFlatten, flattened)) {
+          queryParams.query = JSON.stringify(flattened.query);
```
clyde-builderio: This is similar to Gen-2 implementation
https://github.com/BuilderIO/builder/blob/70fccea549322b20ffb544c798b249e7c2197d4e/packages/sdks/src/functions/get-content/generate-content-url.ts#L98-L103

packages/core/src/builder.class.test.ts
```diff
@@ -1021,6 +1021,81 @@ describe('flushGetContentQueue', () => {
       { headers: { Authorization: `Bearer ${AUTH_TOKEN}` } }
     );
   });
+
+  test('hits content url with query as the same object as the one passed in options.query if query contains $ mongo-operator', async () => {
```
rima-builder, line 1025: could you also look at the tests from the content API and see if we can add similar (if not present) ?

clyde-builderio: Content API did not fail. SDK deleted fields that should've been part of the query-params for Content API

packages/core/src/builder.class.ts
midhunadarvin: Could you add a comment on why we are checking for shallow equality ? What does `flattenMongoQuery` do ?

clyde-builderio: Good point. this implementation is incorrect. updated the code with a better implementation

packages/core/src/builder.class.ts
```diff
@@ -2709,11 +2716,12 @@ export class Builder {
...
     if (this.apiEndpoint === 'content') {
       if (queue[0].query) {
-        const flattened = this.flattenMongoQuery({ query: queue[0].query });
+        delete queryParams.query;
+        const objectToFlatten = { query: queue[0].query };
+        const flattened = this.flattenMongoQuery(objectToFlatten);
         for (const key in flattened) {
-          queryParams[key] = flattened[key];
+          queryParams[key] = JSON.stringify(flattened[key]);
         }
```
clyde-builderio, line 2724: This is consistent with the Gen-2 Implementation
https://github.com/BuilderIO/builder/blob/70fccea549322b20ffb544c798b249e7c2197d4e/packages/sdks/src/functions/get-content/generate-content-url.ts#L98-L103

packages/core/src/builder.class.ts
```diff
@@ -2709,11 +2716,12 @@ export class Builder {
 
     if (this.apiEndpoint === 'content') {
       if (queue[0].query) {
-        const flattened = this.flattenMongoQuery({ query: queue[0].query });
+        delete queryParams.query;
```
clyde-builderio, line 2719: If i didn't delete this, and we had query object
```typescript
{ id: 123 }
```
then `queryParams` would be (after the `flattenMongoQuery()`)
```typescript
{
'query.id': 123,
`query`: { id: 123 }
}
```

midhunadarvin (approved):

sidmohanty11 (approved):
LGTM!

i think this test - https://github.com/BuilderIO/builder/pull/4102/files#file-packages-sdks-tests-src-e2e-tests-hit-content-api-spec-ts-L74 needs to be updated as `query.id` will preserve the strings now right? so - `query.id=%22id%22`

also can you add '@builder.io/react' to the changeset too?

PR #4104, opened 2025-06-26
fix[gen1][core]: ENG-9774 backported fix from v6.1.1 into v5 upgrade

clyde-builderio:
For Manual upgrade only
Reference: https://github.com/BuilderIO/builder/pull/4102


clyde-builderio:
Closing to create a new PR

PR #4108, opened 2025-06-30
feat(react-native): access DOMNode directly from ref in React 19

kerwanp:
## Description

In react-dom@19 the function `findDOMNode` has been removed. Instead the `ref` actually contains the Element.
To support both 18 and 19 version we check wether `findDOMNode` is defined.

image


sidmohanty11 (approved):
thanks for the fix @kerwanp !! LGTM.

.changeset/strange-donuts-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+"@builder.io/sdk-react-native": patch
+---
+
+Add support for React 19
```
yash-builder, line 5:
```suggestion
Fix: Add support for React 19 by handling `findDOMNode` deprecation
```

PR #4114, opened 2025-07-11
fix[gen2][fetchOneEntry]: ENG-9110 make fetchOneEntry() handle options.locale just like builder.get() handles options.locale

clyde-builderio:
## Description

 FetchEntries - includes extra userAttribute: {locale: value} along with the locale payload

`builder.getAll()` hits the Content API without the addition of `locale` to `userAttributes`. See here: https://www.loom.com/share/7117e8725b69400f98e02d823a901b7a

The change in this [PR](https://github.com/BuilderIO/builder/pull/3719)  was added to ensure `locale` is added to `userAttributes` for `builder.get()` and not `builder.getAll()`

However, to ensure the same outcome In Gen2 SDK the [change](https://github.com/BuilderIO/builder/pull/3719/files#diff-b4fb78a40f76e70808858993da491c522b4f567feefeb6c01b0ab5defac58569) of adding `locale` to `userAttributes` is added in `packages/sdks/src/functions/get-content/generate-content-url.ts` which is used in **BOTH** `fetchOneEntry()` and `fetchEntries()`.

**PR that caused the bug**
https://github.com/BuilderIO/builder/pull/3719

**Link to JIRA ticket**
https://builder-io.atlassian.net/browse/ENG-9110

### DIFF
--- a/packages/sdks/src/functions/get-content/index.ts
+++ b/packages/sdks/src/functions/get-content/index.ts
@@ -20,6 +20,16 @@ const checkContentHasResults = (
 export async function fetchOneEntry(
   options: GetContentOptions
 ): Promise<BuilderContent | null> {
+  const finalLocale = options.locale || options.userAttributes?.locale;
+
+  if (finalLocale) {
+    options.locale = finalLocale;
+    options.userAttributes = {
+      locale: finalLocale,
+      ...options.userAttributes,
+    };
+  }
+
   const allContent = await fetchEntries({ ...options, limit: 1 });
 
   if (allContent) {
--- a/packages/core/src/builder.class.test.ts
+++ b/packages/core/src/builder.class.test.ts
@@ -1195,6 +1195,20 @@ describe('get', () => {
       { headers: { Authorization: `Bearer ${AUTH_TOKEN}` } }
     );
   });
+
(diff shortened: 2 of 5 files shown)


packages/sdks/src/functions/get-content/index.ts
```diff
@@ -20,6 +20,16 @@ const checkContentHasResults = (
...
 ): Promise<BuilderContent | null> {
+  const finalLocale = options.locale || options.userAttributes?.locale;
+
+  if (finalLocale) {
+    options.locale = finalLocale;
+    options.userAttributes = {
+      locale: finalLocale,
+      ...options.userAttributes,
+    };
+  }
```
samijaber, line 31: IMO all logic that manipulates the final URL should live in `generateContentUrl` which is used in multiple places.

also, more importantly: we already have very similar logic in `generateContentUrl`:
 https://github.com/BuilderIO/builder/blob/4ec6e0eac9d95cedc8c940fcd335a708aa85f9cc/packages/sdks/src/functions/get-content/generate-content-url.ts#L55-L65

so i think we should modify that existing logic no?

samijaber (changes requested):

samijaber (approved):
just watched the looms, makes sense.

PR #4121, opened 2025-07-22
upgrade node-fetch for admin-sdk

sanyamkamat:
## Description

Upgrades node-fetch to patched version which doesnt have vulnerability.

### DIFF
--- a/packages/admin-sdk/package.json
+++ b/packages/admin-sdk/package.json
@@ -56,7 +56,8 @@
   },
   "pnpm": {
     "overrides": {
-      "node-fetch": "^2.7.0"
+      "node-fetch": "2.7.0",
+      "isomorphic-fetch > node-fetch": "2.7.0"
     }
   }
 }
--- a/packages/admin-sdk/package.json
+++ b/packages/admin-sdk/package.json
@@ -53,5 +53,10 @@
   },
   "resolutions": {
     "node-fetch": "^2.7.0"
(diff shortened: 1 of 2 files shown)


packages/admin-sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@builder.io/admin-sdk",
-  "version": "0.0.8",
+  "version": "0.0.9-0",
```
kylefowler: Why the -0?

packages/admin-sdk/package.json
```diff
@@ -45,7 +45,13 @@
     "fs-extra": "^11.3.0",
     "graphql-typed-client": "^1.7.4",
     "lodash": "^4.17.21",
-    "node-fetch": "^3.3.2",
+    "node-fetch": "^2.7.0",
```
kylefowler, line 48: Did we down grade?

packages/admin-sdk/package.json
sanyamkamat: These are dev releases for the customer to test/verify. Actual merge will have the final non `-n` version.

packages/admin-sdk/package.json
sanyamkamat: We actually dont actually use node-fetch directly in the admin-sdk. Keeping all on same version felt safe.

PR #4122, opened 2025-07-22
Fix null translateable fields

chris-budd:
## Description

When trying to get translatable fields, if any of them are null (like `isRSC`) it fails. This adds a check to avoid that.

See SUPP-1396 internally.


chris-budd:
@clyde-builderio That makes sense! Thanks. I've removed the changeset now so the checks are passing.

clyde-builderio:
> @chris-budd this commit doesn't have a verified signature so merging is blocked. I will add a verified signature

@chris-budd I guess I have to add a signature to each of the 3 commits. Doing that right now

PR #3732, opened 2024-11-13
add support for binding responsive styles in gen1 react sdk

steve8708:
currently gen2 SDKs support binding to responsive styles, like 

```js
{ bindings: { 'responsiveStyles.medium.color': 'state.foo'}
```

but not gen1, this fixes that

PR #4128, opened 2025-08-14
fix: passing custom breakpoints down to symbol

sidmohanty11:
## Description

when we passed content of symbol down - we do a deepClone and here we were skipping `meta` field that contains the breakpoints object

i think it's safe to clone meta field too?

Jira
https://builder-io.atlassian.net/browse/ENG-10131

loom:
https://www.loom.com/share/c31a74790c6d4aac9d9375dcf8a3ce69

### DIFF
--- a/.changeset/nervous-rabbits-sing.md
+++ b/.changeset/nervous-rabbits-sing.md
@@ -1,0 +1,5 @@
+---
+"@builder.io/react": patch
+---
+
+fix passing down custom breakpoints to Symbol
--- a/packages/react/src/components/builder-block.component.tsx
+++ b/packages/react/src/components/builder-block.component.tsx
@@ -43,11 +43,7 @@ export function deepCloneWithConditions<T = any>(obj: T): T {
   for (const key in obj) {
     if (Object.prototype.hasOwnProperty.call(obj, key)) {
       if (key === 'meta') {
-        clonedObj.meta = {
-          breakpoints: {
-            ...obj['meta'].breakpoints,
-          },
-        };
+        clonedObj.meta = obj['meta'];
       } else {
         clonedObj[key] = deepCloneWithConditions(obj[key]);
       }
--- a/packages/react/src/components/builder-block.component.tsx
+++ b/packages/react/src/components/builder-block.component.tsx
@@ -41,7 +41,7 @@ export function deepCloneWithConditions<T = any>(obj: T): T {
   const clonedObj: any = {};
 
   for (const key in obj) {
-    if (key !== 'meta' && Object.prototype.hasOwnProperty.call(obj, key)) {
+    if (Object.prototype.hasOwnProperty.call(obj, key)) {
       clonedObj[key] = deepCloneWithConditions(obj[key]);
     }
   }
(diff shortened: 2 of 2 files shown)


packages/react/src/components/builder-block.component.tsx
```diff
@@ -41,7 +41,7 @@ export function deepCloneWithConditions<T = any>(obj: T): T {
   const clonedObj: any = {};
 
   for (const key in obj) {
-    if (key !== 'meta' && Object.prototype.hasOwnProperty.call(obj, key)) {
```
clyde-builderio, line 44: @steve8708 In this PR https://github.com/BuilderIO/builder/pull/3732 why was the `key !== 'meta'` added?

clyde-builderio:
@sidmohanty11 Snippet tests were failing for a duplicate content created with the same URL path. I have deleted the content

packages/react/src/components/builder-block.component.tsx
```diff
@@ -41,7 +41,7 @@ export function deepCloneWithConditions<T = any>(obj: T): T {
...
+    if (Object.prototype.hasOwnProperty.call(obj, key)) {
```
clyde-builderio, line 44: @sidmohanty11 we could be more specific here
```typescript
if (Object.prototype.hasOwnProperty.call(obj, key)) {
  if (key === 'meta') {
    clonedObj.meta = {
      breakpoints: {
        ...obj['meta'].breakpoints,
      },
    };
  } else {
    clonedObj[key] = deepCloneWithConditions(obj[key]);
  }
}
```

clyde-builderio: This would ensure that `deepCloneWithConditions()` is not called for key as `meta` and your intention to add ONLY breakpoints is also fulfilled

.changeset/nervous-rabbits-sing.md
```diff
@@ -0,0 +1,5 @@
+---
+"@builder.io/react": patch
```
clyde-builderio, line 2: is it possible that for existing users using Symbols, the breakpoints were not being passed down and they were happy with the outcome.

But with a `patch` update, the breakpoints would suddenly start to be passed down and they would get a different outcome?

packages/react/src/components/builder-block.component.tsx
sidmohanty11: updated

.changeset/nervous-rabbits-sing.md
sidmohanty11: good question, but i think while users are editing a symbol content the breakpoints work (because it's a symbol content and we dont omit meta) so i suppose they expect it to work but never verified?

packages/react/src/components/builder-block.component.tsx
steve8708: do you have a use case where people need to use `meta` in this flow? aka is this causing issues? 

tl;dr iirc the reason was the meta object can be large and should not be used in rendering, so cloning it can be expensive and pointless

sidmohanty11: it strips the meta from symbol.content here, we need the meta.breakpoints info for users using custom breakpoints:
https://github.com/BuilderIO/builder/blob/695390cba04a90eae25f263d37d43cce4f352d49/packages/react/src/components/builder-block.component.tsx#L424

steve8708: one suggestion then - copy it as-is and don't clone

so

```
clonedObj.meta = obj.meta
```

and not actually traverse and copy property by property that object (for perf reasons)

sidmohanty11: sure updated

PR #4139, opened 2025-09-05
Smartling plugin updates

murrayadam:
## Description

- Users should be able to filter locales and bulk select in translation job
- Users should be able to directly go to content in Smartling from Builder content entry
- Users should be able to send unpublished content entry content to Smartling
- Users should be able to add content entries to new Smartling jobs without having to duplicate the entry
- Users should be able to specify a default project that is preselected in translation jobs
- Private token is obfuscated in plugin settings
- Auto-authorization settings should be respected from plugin settings

### DIFF
--- a/plugins/smartling/src/plugin.tsx
+++ b/plugins/smartling/src/plugin.tsx
@@ -178,8 +178,16 @@ Builder.register('plugin', {
 // Create API instance for plugin use
 const api = new SmartlingApi();
 
+let isInitialized = false;
+
 // Initialize plugin functionality
 const initializeSmartlingPlugin = async () => {
+
+  if (isInitialized) {
+    return;
+  }
+  isInitialized = true;
+  
   // Wait for API to initialize - this should always happen
   await api.loaded;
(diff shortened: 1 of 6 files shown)


plugins/smartling/src/plugin.tsx
```diff
@@ -30,86 +27,197 @@ import stringify from 'fast-json-stable-stringify';
...
+  
+  // Check if job actually exists
+  const jobExists = await api.checkTranslationJobExists(translationJobId);
+  
+  // Update cache
+  jobExistenceCache.set(translationJobId, { exists: jobExists, timestamp: now });
+  
+  // If job doesn't exist, clean up metadata
+  if (!jobExists) {
+    await api.cleanupOrphanedTranslationMetadata(content);
```
anaghav2023: @murrayadam this has introduced some bugs. It is cleaning the meta data, which is required in the API. After I comment this, it works fine.

sanyamkamat (approved):

PR #4146, opened 2025-09-17
[plugin][smartling]:ENG-10295 Ability to add content entries to existing job

anaghav2023:
## Description
1. Adding content entries to existing jobs was not working because the content status was still published.
2. View translation job button was not showing earlier due to the cache checks which don't work.

### DIFF
--- a/plugins/smartling/src/plugin.tsx
+++ b/plugins/smartling/src/plugin.tsx
@@ -664,22 +664,10 @@ const initializeSmartlingPlugin = async () => {
           return false;
         }
         
-        // Show if content has a translation job ID (regardless of status)
         if (!translationJobId) {
           return false;
         }
-        
-        
-        // Trigger background validation and cleanup if needed
-        isContentInActiveTranslationJob(content, api).catch(() => {});
-        
-        // Check cache for immediate result
-        const cached = jobExistenceCache.get(translationJobId);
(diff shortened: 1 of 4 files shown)


yash-builder (approved):

plugins/smartling/src/plugin.tsx
```diff
@@ -664,22 +664,10 @@ const initializeSmartlingPlugin = async () => {
...
         }
         
-        // Show if content has a translation job ID (regardless of status)
         if (!translationJobId) {
           return false;
         }
-        
-        
-        // Trigger background validation and cleanup if needed
-        isContentInActiveTranslationJob(content, api).catch(() => {});
```
sanyamkamat, line 674: what was this code doing?

anaghav2023: 1. This code was to show the "View Translation Job" button in the content actions. Now this button takes us to the data model content entry that got created for the Job. But this function was checking whether an active translation job exists in smartling. So as soon as the job got completed, this button would disappear. So I removed the check, as users want to have a link between content entry and builder job.
2. The thing to think about here is that 1 content entry could be added to multiple jobs, so we need a better UI to show the list of jobs for a content entry. Right now it takes us to the most recent job.
3. Also, this function was buggy, it was clearing the meta data which lines I have commented.

sanyamkamat (approved):

PR #4148, opened 2025-09-18
Return sku code in the content with Commercelayer Plugin

akakream:
## Description

The commercelayer plugin only returns the sku id when the content is fetched with the Content API. This is not very useful as the id of the SKU is not very useful in Commercelayer.
Ideally, one should be able to get the SKU code from the Content API.

This PR adds the SKU code in the response of the Content API for the Content Type Commercelayer product.

_Screenshot_


akakream:
thank you @paprikaf

PR #4152, opened 2025-09-24
fix[angular-gen2]: only use the edge runtime when the edge build is imported

sidmohanty11:
## Description

As we are skipping hydration due to limitations with how we generate dynamic components we have to run eval twice - once on the server and once on the browser. The edge build was working and edge-runtime was being used but on the browser load it was fallbacking to use the browser runtime which was leading to the unsafe-eval errors. This PR fixes the issue by using edge-runtime both in the server and the browser - we will only use the edge build when users import it using:
```ts
import { Content, type BuilderContent } from "@builder.io/sdk-angular/bundle/edge";
```

Jira
https://builder-io.atlassian.net/browse/ENG-10322

Screenshot
image

### DIFF
--- a/packages/sdks/output/angular/scripts/multi-build.mjs
+++ b/packages/sdks/output/angular/scripts/multi-build.mjs
@@ -62,8 +62,12 @@ const transformFile = (filePath, replaceValue, revert = false) => {
                 hasReplacedImport = true;
               }
             },
-            // When building EDGE bundle, force choose-eval to always use the edge evaluator
-            // and skip any browser fallback logic.
+            /**
+             * When building EDGE bundle, force choose-eval to always use the edge evaluator
+             * and skip any browser fallback logic.
+             * 
+             * converts chooseBrowserOrServerEval = (args) => evaluator(args)
+             */
             ExportNamedDeclaration(path) {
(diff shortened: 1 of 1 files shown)


packages/sdks/output/angular/scripts/multi-build.mjs
```diff
@@ -62,6 +62,26 @@ const transformFile = (filePath, replaceValue, revert = false) => {
...
+
+              // Replace the entire initializer with a simple arrow function
+              first.init = babel.types.arrowFunctionExpression(
+                [babel.types.identifier('args')],
+                babel.types.callExpression(
+                  babel.types.identifier('evaluator'),
+                  [babel.types.identifier('args')]
+                )
+              );
+            },
```
midhunadarvin: Could you add a comment for understanding this transformation easier. From what i understand, we will transform this export : 
```
export const chooseBrowserOrServerEval = (args: ExecutorArgs) =>
  isBrowser() || shouldForceBrowserRuntimeInNode({ shouldLogWarning: true })
    ? runInBrowser(args)
    : evaluator(args);
```
to 
```
export const chooseBrowserOrServerEval = (args) => evaluator(args) 
```

Is this correct ?

midhunadarvin (approved):

PR #4153, opened 2025-09-24
fix: conversion not being tracked properly with builder action

midhunadarvin:
## Description

Gen 1 : 
- expose `contentId` on the builder class 
- update the `trackConversion` method to figure out `contentId` and `variantionId` if they are undefined from the SDK value and cookies.

Gen 2: 
- add track and trackConversion methods to the builder globals object.
- add `BuilderGlobalContext` class to share context variables between components and helper functions in the gen2 sdks

Issue : 
https://www.loom.com/share/6bcecbfc1b284331a867f162223e4c20

**Loom - Gen 1 fix**
https://www.loom.com/share/3712e28c436b4d9eac9bd3e6a2a68325

**Loom - Gen 2 fix**
https://www.loom.com/share/fc3b894372d244d8b0c2a14e67ed93ec

---

> [!NOTE]
> Ensure conversions capture contentId and variationId via instance/global context and cookies, adding tracking APIs, inlined context setup, and comprehensive tests.
> 
> - **Core (Gen1)**:
>   - Expose `Builder.contentId` (reactively) and set it in React components on content load/inline render.
>   - Enhance `Builder.trackConversion` to derive `contentId` from instance and `variationId` from test cookie; ignore variant when equal to content.
>   - Add extensive unit tests for conversion parameter handling and edge cases.
> - **SDKs (Gen2)**:
>   - Add global context singleton (`initializeGlobalBuilderContext`) and inject setup script from `Content` to store `apiKey`, `apiHost`, and `contentId`.
>   - Extend `builder` globals with `track` and `trackConversion`, using global context and `getTestCookie`.
>   - Expose/init variant/helper scripts to window (incl. global context initializer) for A/B and personalization flows.
> - **E2E/Specs**:
>   - New conversion tracking pages/specs for basic, variant (cookie-driven), symbol, and section scenarios.
>   - Export test helpers; register new routes/content in specs.
> - **Tooling/Docs**:
>   - Add `watch:sdk` script; document alternative Yarn link workflow in SDK develop docs.
> - **Changeset**: patch releases for affected `@builder.io/*` packages.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2ff9c7f6c16447b787719391381b54d5df8ccb64. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/sdks/src/functions/content-variants.ts
+++ b/packages/sdks/src/functions/content-variants.ts
@@ -1,4 +1,4 @@
-import { getCookieSync, setCookie } from '../helpers/cookie.js';
+import { getCookieSync } from '../helpers/cookie.js';
 
 export const testCookiePrefix = 'builder.tests';
 
@@ -7,56 +7,4 @@ export function getTestCookie(name: string) {
     name: `${testCookiePrefix}.${name}`,
     canTrack: true,
   });
-}
-
-function parseUrlParams(url: string): Map<string, string> {
-  const result = new Map<string, string>();
-
-  try {
-    const urlObj = new URL(url);
-    const params = urlObj.searchParams;
-
-    for (const [key, value] of params) {
-      result.set(key, value);
-    }
-  } catch (error) {
-    console.debug('Error parsing URL parameters:', error);
-  }
-
-  return result;
-}
-
-export function setTestCookie(contentId: string, variationId: string) {
-  // 30 days from now
-  const future = new Date();
-  future.setDate(future.getDate() + 30);
-
-  // Use the native setCookie function directly
-  if (typeof window !== 'undefined') {
-    setCookie({
-      name: `${testCookiePrefix}.${contentId}`,
-      value: variationId,
-      expires: future,
-      canTrack: true,
-    });
-  }
-}
-
-export function setTestsFromUrl() {
-  if (typeof window === 'undefined') return;
-
-  try {
-    // Use native URL object to parse current page URL
-    const params = parseUrlParams(window.location.href);
-
-    // Look for parameters that start with 'builder.tests.'
-    for (const [key, value] of params) {
-      if (key.startsWith(`${testCookiePrefix}.`)) {
-        const testKey = key.replace(`${testCookiePrefix}.`, '');
-        setTestCookie(testKey, value);
-      }
-    }
-  } catch (e) {
-    console.debug('Error parsing tests from URL', e);
-  }
 }


packages/sdks/src/functions/cookie.ts
```diff
@@ -0,0 +1,43 @@
+import { isBrowser } from './is-browser.js';
```
sidmohanty11: we already export these helpers from here: https://github.com/sidmohanty11/builder/blob/main/packages/sdks/src/helpers/cookie.ts/#L1-L112

packages/sdks/src/components/content/content.lite.tsx
```diff
@@ -139,6 +142,16 @@ export default function ContentComponent(props: ContentProps) {
...
 
+    setGlobalBuilderContext({
+      apiKey: props.apiKey,
+      apiHost: props.apiHost,
+      contentId: builderContextSignal.value.content?.id,
+    });
+
+    if (isBrowser()) {
+      setTestsFromUrl();
+    }
```
sidmohanty11: why do we need this?

midhunadarvin: This is based on the functionality in the gen 1 react sdk : https://github.com/BuilderIO/builder/blob/6267e52bdb314b2df774bd34504c9d083c4ad7c7/packages/core/src/builder.class.ts#L1836 

When we click on show live preview in the builder app, it takes us to the preview URL with the following query param :
```
?builder.tests.8a20ce9ce2684905a237e98cdcaccd1a=8a20ce9ce2684905a237e98cdcaccd1a
```

In the gen 1 sdk, this query param is processed and set to the `builder.tests` cookie ( which shows the variant content ). In gen 2 sdk, this query param was not processed and clicking on live preview shows the default variant itself. 

Note: There is another bug with SSR here because when the live preview loads for the first time, the cookie is set, but the default variant has already been rendered. On reload of the page, the correct variant is rendered.

packages/sdks/src/functions/cookie.ts
midhunadarvin: updated

packages/sdks/src/components/content/content.lite.tsx
sidmohanty11: from the sound of it, seems like it should be handled here: https://github.com/sidmohanty11/builder/blob/main/packages/sdks/src/components/content-variants/content-variants.lite.tsx/#L95 before hydration, on the first browser load

packages/sdks/src/functions/evaluate/helpers.ts
```diff
@@ -52,6 +69,56 @@ export const getBuilderGlobals = (): BuilderGlobals => ({
...
+    const builderContext = getGlobalBuilderContext();
+
+    if (!useContentId && builderContext?.contentId) {
+      useContentId = builderContext.contentId;
+    }
+
+    let useVariationId = variationId;
+    if (!useVariationId && useContentId) {
+      useVariationId = getTestCookie(useContentId) || undefined;
+    }
```
sidmohanty11, line 108: with all these fallbacks, i am thinking if we should add some e2e tests to make sure the content id will be correct:
- we can call track inside section models
- we can call track inside symbols (should pick the parent contentId)

midhunadarvin: Added the e2e tests

packages/sdks/src/components/content/content.lite.tsx
midhunadarvin: updated

packages/sdks/src/functions/content-variants.ts
```diff
@@ -0,0 +1,62 @@
...
+    // Look for parameters that start with 'builder.tests.'
+    for (const [key, value] of params) {
+      if (key.startsWith(`${testCookiePrefix}.`)) {
+        const testKey = key.replace(`${testCookiePrefix}.`, '');
+        setTestCookie(testKey, value);
+      }
+    }
+  } catch (e) {
+    console.debug('Error parsing tests from URL', e);
+  }
```
sidmohanty11: all of these functions are not used anymore right?

midhunadarvin: yes, not used anymore. Thanks for catching. removed the unused functions

packages/sdks/src/components/content-variants/inlined-fns.ts
```diff
@@ -3,6 +3,108 @@
...
+          } catch (e) {
+            return {};
+          }
+        })()
+    : window;
+
+  if ((globalObject as any).GlobalBuilderContext) {
+    // if already exists, don't re-initialize
+    return;
+  }
```
sidmohanty11, line 50: As discussed on the call, I recommend avoiding a new global context, as maintaining it would create unpredictable problems across SDKs.

- SSR frameworks could behave in unexpected ways (two different global contexts, also cookie isn't present on server)
- This will be an extra addition to the customer’s HTML as we're injecting inlined scripts
- It’s extra code that we have to maintain.

The simpler solution is to expose just `track` and `trackConversion` on the Gen2 SDK globals obj and update every `evaluate` call to include apiKey and contentId.

PR #4155, opened 2025-09-29
fix[react-gen1]: handle `fetchPriority` vs `fetchpriority` in different react versions

sidmohanty11:
## Description

Handles fetch priority casing based on React versions

> In a previous PR, https://github.com/vercel/next.js/pull/47302, detection for fetchPriority assumed that https://github.com/facebook/react/pull/25927 would land in react@18.3.0 because that was the react@canary version at the time. However, it didn't land in react@18.3.0 and it is expected to land in react@19.0.0 due to the breaking change.
This means that users upgrading to react@18.3.0 will see a warning.

Inspiration from:
https://github.com/facebook/react/issues/28946
https://github.com/vercel/next.js/pull/65235
https://github.com/ascorbic/unpic-img/pull/644

_Screenshot_

### DIFF
--- a/packages/react/src/blocks/Image.tsx
+++ b/packages/react/src/blocks/Image.tsx
@@ -263,10 +263,14 @@ class ImageComponent extends React.Component<any, { imageLoaded: boolean; load: 
     return this.props.image || this.props.src;
   }
 
+  get loadEagerly() {
+    return this.props.builderBlock?.id.startsWith('builder-pixel-') || this.props.highPriority;
+  }
+
   get fetchPriorityProp() {
     const propNameToUse = isNewReact ? 'fetchPriority' : 'fetchpriority';
     return {
-      [propNameToUse]: this.props.highPriority ? 'high' : 'auto',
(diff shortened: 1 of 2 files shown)


packages/react/src/blocks/Image.tsx
```diff
@@ -261,6 +263,13 @@ class ImageComponent extends React.Component<any, { imageLoaded: boolean; load:
     return this.props.image || this.props.src;
   }
 
+  get fetchPriorityProp() {
+    const propNameToUse = isNewReact ? 'fetchPriority' : 'fetchpriority';
+    return {
+      [propNameToUse]: this.props.highPriority ? 'high' : 'auto',
```
sanyamkamat: The previous logic for this value was 
`const eagerLoad = isPixel || this.props.highPriority;`

Do we need to add `isPixel` check here?

sidmohanty11: thanks for the catch, i've updated it

PR #4003, opened 2025-04-03
wip: angular sdk revamp with signals

sidmohanty11:
## Description

Add a short description of what changes you made, why you made them, and any other context that you think might be helpful for someone to better understand what is contained in this pull request. This sort of information is useful for people reviewing the code, as well as anyone from the future trying to understand why changes were made or why a bug started happening.

_Screenshot_
If relevant, add a screenshot or two of the changes you made.

PR #4532, opened 2026-04-20
fix[gen2][react]: ENG-12210 custom component children passed as Fragment instead of array

floating-dynamo:
## Description
- Removed a redundant inner `<Show>` guard in `component-ref.lite.tsx` that caused the React SDKs to wrap children in a `React.Fragment`. 
- There is already an outer `<Show>` guard which is already there to gate rendering based on the `props.componentRef`.
- This broke the React.Children.count in custom components with `canHaveChildren: true`.

_Screenshot_
### Issue
Loom - https://www.loom.com/share/7db1876f7eb84959aa311ef2a9b166d4

### Fix
image

---

> [!NOTE]
> **Low Risk**
> Small, localized rendering change plus version bumps; main risk is subtle regressions in child rendering order/guards for custom components.
> 
> **Overview**
> Fixes custom component `children` rendering in `component-ref` by removing an inner redundant `<Show>` around the `<For>` loop that was causing children to be wrapped in a `React.Fragment`, breaking `React.Children.count` and iteration.
> 
> Adds a small `blockChildrenToRender` computed in state so the `<For>` receives `[]` when `componentRef` is absent, while keeping the existing outer render guard. Includes a patch changeset for `@builder.io/sdk-react`, `@builder.io/sdk-react-nextjs`, and `@builder.io/sdk-react-native`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit afdf1e79082aa9b7121fb13e337734316f6bea8f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/sdks/src/components/block/components/component-ref/component-ref.lite.tsx
+++ b/packages/sdks/src/components/block/components/component-ref/component-ref.lite.tsx
@@ -51,19 +51,17 @@ export default function ComponentRef(props: ComponentProps) {
           contextValue: props.context.value,
         })}
       >
-        <Show when={props.componentRef}>
-          <For each={props.blockChildren}>
-            {(child) => (
-              <Block
-                key={child.id}
-                block={child}
-                context={props.context}
-                registeredComponents={props.registeredComponents}
-                linkComponent={props.linkComponent}
-              />
-            )}
-          </For>
-        </Show>
+        <For each={props.blockChildren}>
+          {(child) => (
+            <Block
+              key={child.id}
+              block={child}
+              context={props.context}
+              registeredComponents={props.registeredComponents}
+              linkComponent={props.linkComponent}
+            />
+          )}
+        </For>
       </state.Wrapper>
     </Show>
   );
--- a/packages/sdks/src/components/block/components/component-ref/component-ref.lite.tsx
+++ b/packages/sdks/src/components/block/components/component-ref/component-ref.lite.tsx
@@ -35,6 +35,9 @@ export default function ComponentRef(props: ComponentProps) {
           default: InteractiveElement,
         })
       : props.componentRef,
+    get blockChildrenToRender() {
+      return props.componentRef ? props.blockChildren : [];
+    },
   });
 
   return (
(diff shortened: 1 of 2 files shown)


AishwaryaParab:
@floating-dynamo question here: what was breaking for JTI exactly and why was it breaking? How did the inner Show wrapper affect them basically?

floating-dynamo:
> @floating-dynamo question here: what was breaking for JTI exactly and why was it breaking? How did the inner Show wrapper affect them basically?

Inshort this is what was happening

**Expected behaviour:**

```js
children = [element, element, element]
React.Children.count(children) === 3
isFragment(children) === false
```

**v5.1.1 behavior (actual):**

```js
children = <Fragment>[element, element, element]</Fragment>
React.Children.count(children) === 1
isFragment(children) === true
```

packages/sdks/src/components/block/components/component-ref/component-ref.lite.tsx
```diff
@@ -51,19 +54,17 @@ export default function ComponentRef(props: ComponentProps) {
...
-                key={child.id}
-                block={child}
-                context={props.context}
-                registeredComponents={props.registeredComponents}
-                linkComponent={props.linkComponent}
-              />
-            )}
-          </For>
-        </Show>
+        <For each={state.blockChildrenToRender}>
```
midhunadarvin, line 57: ❔**question:** ‏
Do we know why the `Show` component was added ? Should we remove the wrapper `Show` component based on some options or parameter ?

floating-dynamo: Looks like the change was made here - https://github.com/BuilderIO/builder/pull/4003 it was done for some angular gen2 sdk related change

AishwaryaParab: maybe check if there can be any regressions if we remove this Show wrapper or no.

floating-dynamo: I compared the angular output between main and my branch there is nothing that will hamper CSS - [Slack thread](https://builder-internal.slack.com/archives/C07P1H7496H/p1776769220323279)

AishwaryaParab (approved):
LGTM

PR #4651, opened 2026-06-09
fix[utils][smartling]: ENG-12274 smartling plugin skips nested fields for translation

AishwaryaParab:
## Description

When a data model field has the **Localize** toggle enabled, Builder stores the entire array as a single **LocalizedValue** object. The Smartling plugin was not sending the individual string fields inside that array for translation. So, only the parent field was recorded, but as a non-string value that Smartling cannot translate.

Example data structure that was broken:
```
"items": {
  "@type": "@builder.io/core:LocalizedValue",
  "Default": [
    { "question": "What does an Excel expert do?", "answer": "An Excel expert..." }
  ]
}
```
`question` and `answer` were dropped and nothing reached Smartling.

Fix:
Updated getTranslateableFields in @builder.io/utils to recursively walk into array/object payloads of a LocalizedValue and extract individual string leaves as separate translation entries:
- metadata.faqs#items#0#question → "What does an Excel expert do?"
- metadata.faqs#items#0#answer → "An Excel expert..."

Also updated applyTranslation to correctly write translated values back into the right position in the locale-specific copy of the array.

Two new unit tests added:
- Verifies extraction produces the correct flat string keys for nested array fields
- Verifies apply-back correctly populates the target locale slot while preserving Default

_Screenshot_
https://www.loom.com/share/e0e1bad23bac408d8ee17619a60bac07

---

> [!NOTE]
> **Medium Risk**
> Changes how localized content metadata is read and written for translation workflows; behavior is covered by new tests but incorrect edge cases could corrupt locale-specific data shapes.
> 
> **Overview**
> Fixes **Smartling** skipping translatable copy when a localized custom field’s value is an **array or object** (e.g. FAQ `items` with `question` / `answer` strings) instead of a single string.
> 
> **`getTranslateableFields`** no longer uses `traverse` on metadata custom fields. It walks **`LocalizedValue`** boundaries with **`extractCustomField`** and **`extractNestedStrings`**, emitting one string entry per leaf (e.g. `metadata.faqs#items#0#question`) instead of a single non-string parent key. Nested object fields like **`metadata.seo`** now split into separate keys (snapshot updated).
> 
> **`applyTranslation`** gains optional **`sourceLocaleId`**, matches compound translation keys under a **`LocalizedValue`**, deep-clones the source locale payload, patches leaves (keeping nested **`LocalizedValue`** objects intact), and stores the result on the target locale. New unit tests cover FAQ extraction/apply-back, double-localized sub-fields, and sourcing from **`en-US`** when it differs from **`Default`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 13ebef5e378ed4ef178c67ef0efaca1dc37f04ec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/utils/src/__snapshots__/translation-helpers.test.ts.snap
+++ b/packages/utils/src/__snapshots__/translation-helpers.test.ts.snap
@@ -393,16 +393,13 @@ Object {
     "instructions": "Button with plain text",
     "value": "Cute Baby",
   },
-  "metadata.seo": Object {
+  "metadata.seo#menuItems#0#menuName": Object {
     "instructions": "Visit https://builder.io/fiddle/... for more details",
-    "value": Object {
-      "menuItems": Array [
-        Object {
-          "menuName": "en menu name",
-        },
-      ],
-      "name": "en name in subfield",
-    },
+    "value": "en menu name",
+  },
+  "metadata.seo#name": Object {
+    "instructions": "Visit https://builder.io/fiddle/... for more details",
+    "value": "en name in subfield",
   },
   "metadata.title": Object {
     "instructions": "Visit https://builder.io/fiddle/... for more details",
--- a/packages/utils/src/translation-helpers.test.ts
+++ b/packages/utils/src/translation-helpers.test.ts
@@ -969,18 +969,6 @@ test('applyTranslation for symbol with localized array containing nested localiz
   expect(symbolData.uspList.Default[1].headline.Default).toBe('Organise your items');
 });
 
-// ---------------------------------------------------------------------------
-// Customer bug: nested fields inside a LocalizedValue array (custom metadata)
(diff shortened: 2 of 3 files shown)


anaghav2023 (commented):
@AishwaryaParab changes look good. I think this is tested on a data model content entry. Better to test this on a normal page model content entry as well

anaghav2023 (approved):

AishwaryaParab:
> changes look good. I think this is tested on a data model content entry. Better to test this on a normal page model content entry as well

I tested this with a page model. The nested fields are getting translated correctly.

floating-dynamo (approved):

PR #4694, opened 2026-06-24
fix[gen2][abTests]: ENG-12786 stop injecting duplicate A/B test init scripts on pages with multiple Content components

AishwaryaParab:
## Description

The Gen 2 React SDK injects the full `window.builderIoAbTest / window.builderIoRenderContent` definition as an inline <script> once per <Content> component, even on pages that have no A/B tests. On pages with several Content regions (header, body, footer…) this ships multiple identical ~4KB scripts. 

JTI saw 11 copies per page.

**Fix:**
- Only emit the init script when a Content actually renders A/B variants (gate it behind shouldRenderVariants). Pages with no A/B tests now emit it zero times.
- Make the script idempotent `(if (!window.builderIoAbTest) { … })` so repeated definitions never re-run and self-removing on hydration targets, matching how the SDK's existing A/B scripts already behave.

Re-added the e2e assertion that the revert removed, counting init scripts in the SSR response (the script self-removes from the live DOM, so we check the response body). Asserts the realistic post-fix counts.

**Note:**
This eliminates the duplication for the reported case and reduces A/B pages to one init per A/B-tested Content (defined once at runtime). Collapsing to a single tag per page regardless of A/B count requires cross-component coordination and is being tracked as a separate PR, since that touches the same hydration-sensitive path that caused the prior revert.

**JIRA Ticket:**
https://builder-io.atlassian.net/browse/ENG-12786

_Screenshot_
Clip of the issue:
https://clips.agent-native.com/r/9yBuG521T6eu

Clip of the fix:
https://clips.agent-native.com/r/JtAevmePWpKT

---

> [!NOTE]
> **Medium Risk**
> Changes inline script placement and hydration-target self-removal on the A/B SSR path, which is historically sensitive, though the fix deliberately avoids reintroducing the prior DOM-mutation regression.
> 
> **Overview**
> **Stops shipping duplicate `builderio-init-variants-fns` inline scripts** when a page has multiple `<Content>` regions. The `window.builderIoAbTest` / `window.builderIoRenderContent` init block is **moved out of the always-on path** and **only rendered when `shouldRenderVariants` is true**, so non–A/B content no longer injects it.
> 
> The init script string is now wrapped in an **idempotent IIFE** (assign globals only if missing) and on React/React Native hydration targets **removes itself via `document.currentScript`** after running, avoiding repeat definitions and the client-side DOM tweaks that previously caused hydration issues.
> 
> **E2E:** SSR A/B tests count `builderio-init-variants-fns` in the **raw navigation response** (not the live DOM) and expect **one** script for pages with a single A/B-tested content; Gen1 React and non-SSR frameworks skip the assertion.
> 
> Patch changeset across Gen2 SDK packages documents the behavior change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 03df10b9db3f6a46ee812e33bd9dfa25b3769025. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/sdks/src/components/content-variants/helpers.ts
+++ b/packages/sdks/src/components/content-variants/helpers.ts
@@ -70,8 +70,20 @@ const isAngularSDK = TARGET === 'angular';
 const isHydrationTarget = getIsHydrationTarget(TARGET);
 
 export const getInitVariantsFnsScriptString = () => `
-  window.${UPDATE_COOKIES_AND_STYLES_SCRIPT_NAME} = ${UPDATE_COOKIES_AND_STYLES_SCRIPT}
-  window.${UPDATE_VARIANT_VISIBILITY_SCRIPT_FN_NAME} = ${UPDATE_VARIANT_VISIBILITY_SCRIPT}
+  (function() {
+    if (!window.${UPDATE_COOKIES_AND_STYLES_SCRIPT_NAME}) {
+      window.${UPDATE_COOKIES_AND_STYLES_SCRIPT_NAME} = ${UPDATE_COOKIES_AND_STYLES_SCRIPT};
+    }
+    if (!window.${UPDATE_VARIANT_VISIBILITY_SCRIPT_FN_NAME}) {
+      window.${UPDATE_VARIANT_VISIBILITY_SCRIPT_FN_NAME} = ${UPDATE_VARIANT_VISIBILITY_SCRIPT};
+    }${
+      isHydrationTarget
+        ? `
+    var thisScriptEl = document.currentScript;
+    if (thisScriptEl) { thisScriptEl.remove(); }`
+        : ''
+    }
+  })();
   `;
 
 export const getUpdateCookieAndStylesScript = (
--- a/.changeset/ab-test-init-script-dedupe.md
+++ b/.changeset/ab-test-init-script-dedupe.md
@@ -1,0 +1,12 @@
+---
+"@builder.io/sdk-angular": patch
+"@builder.io/sdk-react-nextjs": patch
+"@builder.io/sdk-qwik": patch
+"@builder.io/sdk-react": patch
+"@builder.io/sdk-react-native": patch
+"@builder.io/sdk-solid": patch
+"@builder.io/sdk-svelte": patch
+"@builder.io/sdk-vue": patch
+---
+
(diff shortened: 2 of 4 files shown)


floating-dynamo:
Hey @AishwaryaParab , did the dev release version with this change, help the customer fix the issue?

packages/sdks/src/components/content-variants/helpers.ts
```diff
@@ -70,8 +70,20 @@ const isAngularSDK = TARGET === 'angular';
 const isHydrationTarget = getIsHydrationTarget(TARGET);
 
 export const getInitVariantsFnsScriptString = () => `
-  window.${UPDATE_COOKIES_AND_STYLES_SCRIPT_NAME} = ${UPDATE_COOKIES_AND_STYLES_SCRIPT}
-  window.${UPDATE_VARIANT_VISIBILITY_SCRIPT_FN_NAME} = ${UPDATE_VARIANT_VISIBILITY_SCRIPT}
+  (function() {
+    if (!window.${UPDATE_COOKIES_AND_STYLES_SCRIPT_NAME}) {
```
floating-dynamo, line 74: What are we doing here exactly?

AishwaryaParab: This script defines the two globals `window.builderIoAbTest` and `window.builderIoRenderContent` that the A/B scripts call during SSR (before hydration) to show the right variant without flicker.

The changes here:
`if (!window.…) guard:` When multiple A/B-tested Content components emit this script, only the first one actually defines the functions. The rest are no-ops instead of re-defining them.

**Self-removal on React**:  The script removes its own <script> tag (same pattern the existing cookie/visibility scripts already use). This keeps the server DOM matching the client so we don't get the hydration mismatch that caused the 5.2.2 revert. The functions stay on window after the tag is removed.

AishwaryaParab:
> Hey @AishwaryaParab , did the dev release version with this change, help the customer fix the issue?

The dev released version was given to the customer on 25th Jun but the customer has not got back to us yet. I discussed with @sanyamkamat regarding this and we decided to go ahead with the prod release for this task.

floating-dynamo (approved):

PR #4760, opened 2026-08-03
feat: ENG-12701 add editor-controlled image sizes with Gen 2 responsive loading parity

floating-dynamo:
## Description
- Enabled the existing Image 'sizes' field in the editor
- Updated Gen2 SDKs to apply it to both the image and its source tag.
- Also added support for sizes = auto for the lazy loaded images for gen2.

https://builder-io.atlassian.net/browse/ENG-12701

**Screenshot/Clip**
Clip - https://clips.agent-native.com/share/hbtCPViNlE8T?ref=clip_share

---

> [!NOTE]
> **Low Risk**
> Changes are limited to Image markup and editor field visibility; they may shift which CDN widths browsers pick but do not touch auth, data, or core app logic.
> 
> **Overview**
> **Exposes the Image block’s `sizes` option in the visual editor** (advanced field with helper text) across Gen 1 React and Gen 2 SDK component metadata, so authors can set responsive `sizes` without hiding it behind `hideFromUI`.
> 
> **Gen 2 Image rendering** now centralizes `sizes` in `sizesToUse`: use the configured `sizes` when set; otherwise lazy images get `sizes="auto"` and **high-priority images omit `sizes`**. The same value is applied to both the `` and the WebP `<source>`, fixing cases where `srcset` selection ignored `sizes` on the picture source.
> 
> **Tests:** new Playwright coverage for explicit `sizes`, lazy `auto` (Gen 1: `auto, 100vw` vs Gen 2: `auto`), and high-priority behavior; fixture content was adjusted so only the first image carries explicit `sizes` for those scenarios. Patch changeset bumps all affected SDK packages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8403f77ced504a08a781721a9b7eacc20766ac30. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/react/src/blocks/Image.tsx
+++ b/packages/react/src/blocks/Image.tsx
@@ -593,7 +593,9 @@ export const Image = withBuilder(ImageComponent, {
     {
       name: 'sizes',
       type: 'string',
-      hideFromUI: true,
+      advanced: true,
+      helperText:
+        'The sizes attribute for responsive images, e.g. "(max-width: 600px) 100vw, 50vw"',
     },
     {
       name: 'srcset',
--- a/.changeset/real-poets-beam.md
+++ b/.changeset/real-poets-beam.md
@@ -1,0 +1,12 @@
+---
+"@builder.io/react": patch
+"@builder.io/sdk-angular": patch
+"@builder.io/sdk-react-nextjs": patch
+"@builder.io/sdk-qwik": patch
+"@builder.io/sdk-react": patch
+"@builder.io/sdk-solid": patch
+"@builder.io/sdk-svelte": patch
+"@builder.io/sdk-vue": patch
+---
+
+Exposed the existing Image `sizes` field and fixed responsive source selection in Gen 2 SDKs.
--- a/packages/sdks-tests/src/e2e-tests/blocks.spec.ts
+++ b/packages/sdks-tests/src/e2e-tests/blocks.spec.ts
@@ -137,6 +137,28 @@ test.describe('Blocks', () => {
       }
     });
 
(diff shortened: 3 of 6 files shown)


packages/react/src/blocks/Image.tsx
```diff
@@ -593,7 +593,9 @@ export const Image = withBuilder(ImageComponent, {
     {
       name: 'sizes',
       type: 'string',
-      hideFromUI: true,
+      advanced: true,
```
AishwaryaParab, line 596: @floating-dynamo earlier hideFromUI was true but now we have a setting controlling this. This is a change for sizes, so do we have to update any docs for this?

AishwaryaParab:
@floating-dynamo do gen1 sdks need this change? if not, why not?

packages/react/src/blocks/Image.tsx
floating-dynamo: I don’t think an SDK docs update is required because the `sizes` property already existed and was supported. This change only exposes it in the editor behind a space setting, and the field includes helper text with an example.

floating-dynamo:
> @floating-dynamo do gen1 sdks need this change? if not, why not?

Gen 1 React needs the change and is updated in this PR. The other Gen 1 SDKs do not have their own Image component editor schema, they render Builder-generated HTML so there is no separate `sizes` field or `hideFromUI` setting to update.

AishwaryaParab (approved):

PR #4786, opened 2026-08-07
fix[gen2][scripts]: ENG-13137 stop injecting personalization scripts on pages with multiple Content components 

AishwaryaParab:
## Description
On SSR pages, the personalization inline scripts were injected **once per `<Content>` component**, even on pages that use no personalization at all.

A page with a header, a page body and a footer shipped three identical copies of `window.builderIoPersonalization`, `window.filterWithCustomTargeting` and `window.updateVisibilityStylesScript` — roughly **4KB of duplicated HTML per Content**.

The 5.2.4 fix (#4691) solved this for the A/B testing equivalent (`window.builderIoAbTest`) but left the three personalization globals untouched.

**Root cause:**

In `content-variants.lite.tsx` the script was emitted on a single condition: "am I a top-level Content?":
`<Show when={!props.isNestedRender && TARGET !== 'reactNative'}>`
It never checked whether the content actually contained a Variant Container. The A/B equivalent, by contrast, is gated on `shouldRenderVariants`, so it only appears when variants really render.

**Fix:**
- **Emit only when needed**. New `hasPersonalizationContainer()` helper checks whether the content actually contains a Variant Container, and the script is gated on that.
- **Make the definitions idempotent**. Each global is now assigned only `if (!window.X)`, matching what 5.2.4 did for `builderIoAbTest`.

https://builder-io.atlassian.net/browse/ENG-13137

**Screenshot/Clip**
https://clips.agent-native.com/share/CJBhbNylaNH1?ref=clip_share

### DIFF
--- a/packages/sdks/src/blocks/personalization-container/helpers.ts
+++ b/packages/sdks/src/blocks/personalization-container/helpers.ts
@@ -2,6 +2,7 @@ import { TARGET } from '../../constants/target.js';
 import { isBrowser } from '../../functions/is-browser.js';
 import { isEditing } from '../../functions/is-editing.js';
 import type { BuilderBlock } from '../../types/builder-block.js';
+import type { BuilderContent } from '../../types/builder-content.js';
 import type { Target } from '../../types/targets.js';
 import {
   FILTER_WITH_CUSTOM_TARGETING_SCRIPT,
@@ -15,6 +16,8 @@ export const DEFAULT_INDEX = 'default';
 const FILTER_WITH_CUSTOM_TARGETING_SCRIPT_FN_NAME = 'filterWithCustomTargeting';
 const BUILDER_IO_PERSONALIZATION_SCRIPT_FN_NAME = 'builderIoPersonalization';
 const UPDATE_VARIANT_VISIBILITY_SCRIPT_FN_NAME = 'updateVisibilityStylesScript';
+
+const PERSONALIZATION_CONTAINER_COMPONENT_NAME = 'PersonalizationContainer';
 
 export type UserAttributes = {
   date?: string | Date;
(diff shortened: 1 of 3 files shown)


packages/sdks/src/blocks/personalization-container/helpers.ts
```diff
@@ -147,11 +150,46 @@ export function getBlocksToRender({
...
+export const hasPersonalizationContainer = (
+  content: BuilderContent | null | undefined
+) => containsPersonalizationContainer(content);
+
 export const getInitPersonalizationVariantsFnsScriptString = () => {
   return `
-  window.${FILTER_WITH_CUSTOM_TARGETING_SCRIPT_FN_NAME} = ${FILTER_WITH_CUSTOM_TARGETING_SCRIPT}
-  window.${BUILDER_IO_PERSONALIZATION_SCRIPT_FN_NAME} = ${PERSONALIZATION_SCRIPT}
-  window.${UPDATE_VARIANT_VISIBILITY_SCRIPT_FN_NAME} = ${UPDATE_VISIBILITY_STYLES_SCRIPT}
+  (function() {
```
floating-dynamo, line 182: Why was this change made?

floating-dynamo (approved):

packages/sdks/src/blocks/personalization-container/helpers.ts
AishwaryaParab: Every Content block used to inject these personalization scripts in SSR HTML even though personalization wasn't setup for the page. This was done because during SSR, the HTML contains all variants and the job of these scripts is to choose the winner and only show the winner on server side. On client side, everything works but on server side, we reply on these scripts. 

However, we noticed that these scripts were also injected when there were no variants set up for the page which increased the size too. That's why we're removing these scripts if there are no variants.

PR #4789, opened 2026-08-10
feat(sdk): add opt-in debug logging (e.g. to capture x-request-id)

davilima6:
## Purpose

Delivery failures are hard to investigate on Builder's side without a request ID to look up. The SDK already receives an `x-request-id` response header on Content API responses, but `res.json()` discards headers before consumers ever see them, so there is currently no supported way to capture that ID from `@builder.io/sdk` itself.

## Approach and changes

No existing response shapes, return values, or default behavior were changed. With `logLevel` left at its default (`'silent'`), this is a no-op.

- [x] added an opt-in `Builder.logLevel` static setting (`'silent' | 'debug'`, defaults to `'silent'`) that gates a `console.debug` line for every Content API fetch, including the request URL, method, response status, and the `x-request-id` response header when present.
- [x] under Node.js, `Builder.logLevel` defaults from the `BUILDER_SDK_LOG_LEVEL` environment variable (e.g. `BUILDER_SDK_LOG_LEVEL=debug`), so it can be turned on for a deployment without a code change. In browser/edge environments without `process.env`, it falls back to `'silent'` and can still be set programmatically at runtime via `Builder.logLevel = 'debug'`.
- [x] wired the new logging into the two places that consume a fetch `Response` before headers would otherwise be lost: `makeFetchApiCall` (used by `get`/`getAll`/`flushGetContentQueue`) and the `requestUrl` helper kept for the Angular SDK.
- [x] New logic lives in `packages/core/src/functions/log-level.function.ts` so it can be unit tested independently and reads headers defensively (never throws if a `Headers`-like polyfill doesn't implement `.get()`).

## Testing instructions

1. Set `BUILDER_SDK_LOG_LEVEL=debug` in a Node.js process using the SDK and confirm a `console.debug` line appears for each content fetch, including the `x-request-id` value when the API returns one.
1. With `Builder.logLevel` left unset (default), confirm no additional console output appears compared to the current published behavior.

### DIFF
--- a/.changeset/gentle-cameras-behave.md
+++ b/.changeset/gentle-cameras-behave.md
@@ -1,0 +1,5 @@
+---
+"@builder.io/sdk": minor
+---
+
+Added an opt-in `Builder.logLevel` setting (defaulting from the `BUILDER_SDK_LOG_LEVEL` environment variable under Node.js) that logs the request URL, response status, and `x-request-id` response header for every content fetch via `console.debug`. This makes it possible to correlate a specific SDK call with a Builder-side request when reporting intermittent API issues, without changing any existing response shape or default behavior.
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -39,6 +39,26 @@ builder
 builder.canTrack = false;
 ```
 
+## Debugging Content API requests
+
+If you're seeing intermittent issues with Content API responses and need to report them to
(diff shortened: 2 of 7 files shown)


davilima6:
@sanyamkamat , what do you think of this feature and approach? Could we go ahead with it?

PR #4792, opened 2026-08-10
feat: add group to component metadata to support custom component groups

murrayadam:
## Description

Adds an optional `group?: string` field to the SDK's exported component metadata types so TypeScript integrations can pass `group` to `Builder.registerComponent` without tripping excess-property checking.

**Context:** The internal editor PR ([builder-internal#14460](https://github.com/BuilderIO/builder-internal/pull/14460)) adds support for grouping custom components into their own accordion sections in the insert menu (e.g. `group: 'Steps'`, `group: 'Layout'`). The editor reads the value at runtime, but the public SDK types didn't declare the field — so TypeScript users would see:

> Object literal may only specify known properties, and `'group'` does not exist in type `'Component'`.

**Changes:**
- `packages/core/src/builder.class.ts` — added `group?: string` to the `Component` interface (Gen 1, `@builder.io/react`)
- `packages/sdks/src/types/components.ts` — added `group?: string` to the `ComponentInfo` interface (Gen 2 SDKs)

Both are typed as optional strings with JSDoc examples matching the editor's expected values. No runtime behavior change — this is a types-only addition.

**Follow-up:** SDK versions will need to be bumped/published so consumers pick up the new type before the editor feature ships publicly.

---

> [!NOTE]
> **Low Risk**
> Types, docs, and tests only; registration behavior is unchanged aside from documenting and locking through `group` on existing spread-based paths.
> 
> **Overview**
> Adds an optional **`group?: string`** field to component metadata types so TypeScript users can pass `group` in `Builder.registerComponent` without excess-property errors, aligning with the editor’s insert-menu accordion grouping (same `group` → shared section; omitted/blank → default “Custom Components”).
> 
> The field is declared on Gen 1 **`Component`** in `builder.class.ts` and Gen 2 **`ComponentInfo`** in `packages/sdks/src/types/components.ts`, with matching JSDoc. Tests assert compile-time typing, that **`prepareComponentSpecToSend`** keeps `group`, and that **`registerComponent`** stores it on **`Builder.components`**. A changeset patches the published SDK packages; no new runtime registration logic in this diff.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5e282fc7693b8dbc8130991b903c1208a5ebeb0c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sanyamkamat:
@murrayadam we need changeset for this PR

packages/core/src/builder.class.ts
sanyamkamat, line 1: We could add tests too, so that we throw failures if someone accidentally removes this property

sanyamkamat (approved):

PR #4821, opened 2026-08-21
perf(sdk): cache processedBlock on Svelte; single-pass localization

PuruVJ:
Was chasing TBT on a big page and `processedBlock` kept coming up. On Svelte it's recomputed ~25-30 times per block per render. The cache already exists (`_processedBlock`), it's just switched off for Svelte: the read/write/invalidate `useTarget` branches are empty and `svelte` sits in `IS_SDK_WITHOUT_CACHED_PROCESSED_BLOCK`. Let it fall through to the `default` branch so it computes once. The `onUpdate` invalidation runs for Svelte now too, so it still recomputes once per render, just not 25 times.

While in there: `resolveLocalizedValues` walked `component.options` twice (contains-check, then extract), both through the generic `traverse` that allocates a closure plus an `Object.entries` array per node. Made it one in-place pass. Output is byte-identical (checked across densities and locales, including a missing locale).

## Numbers

1,021-block page, ~127k props. tinybench, output asserted byte-identical before timing, `structuredClone` excluded.

| processedBlock reads / block / render | before | after |
| --- | --- | --- |
| 14 (direct sites, floor) | 83 ms | 93x |
| 20 | 119 ms | 133x |
| 25 (realistic) | 149 ms | 166x |
| 30 | 179 ms | 200x |

The "after" path only computes once, so the speedup tracks the read count. Localization on its own is ~2x and byte-identical. The cache is the multiplier.

## Other things

- Kept it Svelte-only. vue/angular/qwik/solid still have empty branches; didn't touch what I can't test. The same one-line flip works for them.
- The localization walk still descends `responsiveStyles`, which is CSS-only and never holds a LocalizedValue. Skipping that one key is another ~30%. Left it out to keep this byte-identical, easy follow-up.
- The cache was presumably off for a reason once. It reads correct now (invalidation fires every update), but the Svelte hydration tests are the real proof I'm not bringing back a stale-render bug.

Short version: it was doing the same work 25 times and walking the tree twice for localization. Now it does the work once and walks once.

---

> [!NOTE]
> **Medium Risk**
> Touches Svelte block processing cache and in-place localization of content options. Wrong cache invalidation or walk behavior could stale-render or drop localized fields.
> 
> **Overview**
> Cuts Svelte TBT by letting `processedBlock` use the existing `_processedBlock` cache (and `onUpdate` invalidation) instead of recomputing ~25 times per block per render. Vue/Angular/Qwik/Solid stay uncached.
> 
> Also rewrites `resolveLocalizedValues` to one in-place walk of `component.options` instead of a contains-check plus extract via `traverse`. Same output, including the missing-locale warning.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 158a958cc7b6a7ec9017f6deb988901573ae3d44. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/sdks/src/components/block/block.lite.tsx
+++ b/packages/sdks/src/components/block/block.lite.tsx
@@ -74,7 +74,6 @@ export default function Block(props: BlockProps) {
     _processedBlock: { value: null as BuilderBlock | null, update: false },
     get processedBlock(): BuilderBlock {
       useTarget({
-        svelte: () => {},
         vue: () => {},
         angular: () => {},
         qwik: () => {},
@@ -102,7 +101,6 @@ export default function Block(props: BlockProps) {
           });
 
       useTarget({
-        svelte: () => {},
         vue: () => {},
         angular: () => {},
         qwik: () => {},
@@ -222,7 +220,6 @@ export default function Block(props: BlockProps) {
    */
   onUpdate(() => {
     useTarget({
-      svelte: () => {},
       vue: () => {},
       angular: () => {},
       qwik: () => {},
--- a/packages/sdks/src/functions/extract-localized-values.ts
+++ b/packages/sdks/src/functions/extract-localized-values.ts
@@ -1,58 +1,55 @@
 import type { BuilderBlock } from '../types/builder-block.js';
-import { traverse } from './traverse.js';
 
-function isLocalizedField(value: any) {
-  return (
-    value &&
-    typeof value === 'object' &&
-    value['@type'] === '@builder.io/core:LocalizedValue'
-  );
+const LOCALIZED_TYPE = '@builder.io/core:LocalizedValue';
+
+function resolveChild(child: any, locale: string, state: { seen: boolean }): any {
+  if (child === null || typeof child !== 'object') {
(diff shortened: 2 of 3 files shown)


kylefowler:
A few comments from the code review bots.

**"Byte-identical" is true for JSON content, not for post-binding values.** `resolveLocalizedValues` runs *after* `evaluateBindings` in `get-processed-block.ts:104-112`, and `set(copied, binding, value)` can put arbitrary runtime JS into `component.options`. Against that input the rewrite is not equivalent to the old `traverse()`: it lost the `WeakSet` cycle guard (stack overflow instead of silent skip), and `for...in` + unconditional write-back touches inherited enumerable keys and throws on frozen objects or getter-only accessors where `Object.entries()` only read. Both are cheap to close — thread a `WeakSet`, switch to `Object.keys`, and only assign when `resolveChild` returns something different. The existing `extract-localized-values.test.ts` covers only plain-JSON shapes, so it wouldn't catch either.

PR #4844, opened 2026-09-04
fix[abTests][gen1]: ENG-13175 A/B Variation Hydration Mismatch on First Server Render

AishwaryaParab:
## Description

When you preview a non-default A/B variation, the page briefly shows the default variation before switching to the one you asked for. This causes a layout shift and a React hydration error. Reloading fixes it, so it only happens on the first visit.

**Root Cause:**
The editor's preview link carries the chosen variation in the URL as `builder.tests.<contentId>=<variationId>`, but nothing reads it early enough:

- The server doesn't look at it, so the HTML ships all variations.
- The inlined variants script only checks the cookie. On a first visit there is no cookie, so it picks a variation at random, usually the default.
- The SDK then writes the URL's variation into the cookie, and React hydrates with that value. It no longer matches the DOM, so React re-renders and you see the switch.

**Fix:**
Both places that choose a variation now check the URL parameter first, then fall back to the cookie, then to the random assignment as before:

- the inlined SSR script, so the correct variation is on screen at first paint
- `VariantsProvider`'s browser branch, so hydration agrees with the DOM

The URL value is validated against the content's known variation ids, and is read at runtime in the browser, so nothing untrusted is written into the script. 

https://builder-io.atlassian.net/browse/ENG-13175

**Screenshot/Clip**
Bug: https://clips.agent-native.com/r/lqL0Ngm5BL7m
Fix: https://clips.agent-native.com/r/ByaNuJ0mjqwo

---

> [!NOTE]
> **Medium Risk**
> Changes first-paint A/B selection and cookie writes in a widely used React path; behavior is constrained by validating URL ids against known variations.
> 
> **Overview**
> Fixes a **flash of the default A/B variation** (and hydration mismatch) when previewing a variation via `builder.tests.<contentId>` on first visit.
> 
> **Variant selection precedence** is now aligned in two places: the **inlined SSR variants script** and **`VariantsProvider`’s browser path** both read `builder.tests.<contentId>` / `builder_tests_<contentId>` first (validated against known variation ids), then cookie, then random assignment. A valid URL choice is also written to the cookie in the inlined script so later logic stays consistent.
> 
> Adds **Jest coverage** for inlined script syntax, URL vs cookie vs random behavior in jsdom, and browser render matching the URL param when no cookie is set. Ships as a **patch** to `@builder.io/react` via changeset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab6cac2b9e1fa649445e2fd7929419392fc2b062. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/react/src/components/variants-provider.component.tsx
+++ b/packages/react/src/components/variants-provider.component.tsx
@@ -18,6 +18,28 @@ function getData(content: BuilderContentVariation) {
   return newData;
 }
 
+// Mirrors the precedence used by the inlined variants script below, so that the DOM
+// the script produces before hydration matches what React renders while hydrating.
+function getVariantIdFromUrl(contentId: string) {
+  const search = (Builder.isBrowser && location.search) || '';
+  if (search.indexOf('builder') === -1) {
+    return null;
+  }
+  const names = ['builder.tests.' + contentId, 'builder_tests_' + contentId];
+  const entries = (search.charAt(0) === '?' ? search.substring(1) : search).split('&');
+  for (const entry of entries) {
+    const parts = entry.split('=');
+    if (names.indexOf(parts[0]) > -1) {
+      try {
+        return decodeURIComponent(parts[1] || '');
+      } catch (err) {
+        return parts[1] || null;
+      }
+    }
+  }
+  return null;
+}
+
 const variantsScript = (variantsString: string, contentId: string) =>
   `
 (function() {
@@ -58,11 +80,34 @@ const variantsScript = (variantsString: string, contentId: string) =>
     }
     return null;
   }
+  function getVariantFromUrl() {
+    var search = location.search || '';
+    if (search.indexOf('builder') === -1) {
+      return null;
+    }
+    var names = ['builder.tests.${contentId}', 'builder_tests_${contentId}'];
+    var entries = (search.charAt(0) === '?' ? search.substring(1) : search).split('&');
+    for (var i = 0; i < entries.length; i++) {
+      var parts = entries[i].split('=');
+      if (names.indexOf(parts[0]) > -1) {
+        try {
+          return decodeURIComponent(parts[1] || '');
+        } catch (err) {
+          return parts[1] || null;
+        }
+      }
+    }
+    return null;
+  }
   var cookieName = 'builder.tests.${contentId}';
   var variantInCookie = getCookie(cookieName);
   var availableIDs = variants.map(function(vr) { return vr.id }).concat('${contentId}');
   var variantId;
-  if (availableIDs.indexOf(variantInCookie) > -1) {
+  var variantInUrl = getVariantFromUrl();
+  if (availableIDs.indexOf(variantInUrl) > -1) {
+    variantId = variantInUrl;
+    setCookie(cookieName, variantId);
+  } else if (availableIDs.indexOf(variantInCookie) > -1) {
     variantId = variantInCookie;
   }
   if (!variantId) {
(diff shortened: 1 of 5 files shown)


packages/react/src/components/variants-provider.component.tsx
```diff
@@ -141,7 +186,11 @@ export const VariantsProvider = ({ initialContent, children, nonce }: VariantsPr
 
   const cookieName = `builder.tests.${initialContent.id}`;
 
-  let variantId = builder.getCookie(cookieName);
+  const variantIdFromUrl = getVariantIdFromUrl(initialContent.id!);
+
+  let variantId = allVariants.some(item => item.id === variantIdFromUrl)
```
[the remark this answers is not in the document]
AishwaryaParab: Not needed. `setTestsFromUrl()` runs in the Builder constructor and already writes every `builder.tests.<contentId>` URL param to the cookie before any render, so a later render without the param still reads the URL-selected value. The script writes it too only because it may run when the bundle never loads.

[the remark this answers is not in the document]
AishwaryaParab: Already aligned. `canTrack` is derived from `browserTrackingDisabled`, which is exactly `window.builderNoTrack || !navigator.cookieEnabled`, and `VariantsProvider` returns the default content on `!canTrack` before the URL check is reached. So with cookies disabled this code path never runs.

packages/react/src/components/variants-provider.component.tsx
```diff
@@ -18,6 +18,28 @@ function getData(content: BuilderContentVariation) {
   return newData;
 }
 
+// Mirrors the precedence used by the inlined variants script below, so that the DOM
+// the script produces before hydration matches what React renders while hydrating.
+function getVariantIdFromUrl(contentId: string) {
+  const search = (Builder.isBrowser && location.search) || '';
```
floating-dynamo, line 24: Why are we also checking `Builder.isBrowser` here ?

packages/react/src/components/variants-provider.component.tsx
```diff
@@ -58,11 +80,34 @@ const variantsScript = (variantsString: string, contentId: string) =>
     }
     return null;
   }
+  function getVariantFromUrl() {
```
floating-dynamo, line 83: How is this different from the one defined above?

floating-dynamo: Is it cause it should be again a part of this variant script in order to exectue?

floating-dynamo:
@AishwaryaParab , also does this change only affect Gen1 React SDKs?

packages/react/src/components/variants-provider.component.tsx
AishwaryaParab: We're following the same pattern here: https://github.com/BuilderIO/builder/pull/4844/changes/BASE..ab6cac2b9e1fa649445e2fd7929419392fc2b062#diff-2e09201826383ce79c7a54fa5e5c9f96ec4cbbbfba124f015f171f97c2146e4cL146

A bare `location.search` would throw a `ReferenceError`. This guard keeps it safe.

AishwaryaParab:
> @AishwaryaParab , also does this change only affect Gen1 React SDKs?

yes

packages/react/src/components/variants-provider.component.tsx
floating-dynamo: And here the builder.isBrowser check is not needed?

AishwaryaParab: This one lives inside the inlined script string, which the browser runs from the HTML before any bundle loads so it can't reference module-scope functions and has to be self-contained. Same reason setCookie/getCookie are duplicated inside that string today. The two copies deliberately use identical precedence (URL, then cookie, then random) so the pre-hydration DOM matches what React renders.

AishwaryaParab: yes, not needed because that function is inside the inlined script string, which only runs as a <script> in the browser (and is only emitted in the Builder.isServer branch).

floating-dynamo (approved):

PR #4848, opened 2026-09-07
Fix CustomCode scriptsClientOnly not re-running scripts after hydration

Manish-Builder-io:
## Description

Fixes a regression where `CustomCode` blocks with `scriptsClientOnly: true` fail to restore SSR-stripped `<script>` tags after hydration.

**Root cause:** `shouldComponentUpdate` in `CustomCodeComponent` only compared `nextProps.code !== this.props.code`. This meant the `setState({ hydrated: true })` call made in `componentDidMount` (which is meant to trigger a one-time re-render to re-insert the scripts that were stripped for SSR) was blocked, since only `state.hydrated` changed and not `props.code`.

**Fix:**
- Added a proper `State` type (`{ hydrated: boolean }`) and used it as the generic parameter on `React.Component<Props, State>` instead of leaving the component untyped for state.
- Updated `shouldComponentUpdate` to also accept `nextState: State` and return `true` when `nextState.hydrated !== this.state.hydrated`, in addition to the existing `code` prop comparison. This preserves the original perf optimization (skipping re-renders when `code` is unchanged) while no longer suppressing the one-time hydration re-render.

**Tests:**
- Added `packages/react/test/custom-code.test.tsx` covering:
  - Scripts are correctly stripped from SSR output when `scriptsClientOnly` is true.
  - Scripts are restored in the DOM and executed after hydration (verifying the fix).
  - `CustomCode` still skips unnecessary re-renders when neither `code` nor hydration state change (verifying no perf regression).
- Updated `packages/react/test/setupTests.ts` to polyfill `TextEncoder`/`TextDecoder` in the jsdom test environment, which `react-dom/server`'s `renderToString` requires but jsdom 19 doesn't provide.
- Added a changeset documenting the patch fix for `@builder.io/react`.

**Link to JIRA ticket:**
https://builder-io.atlassian.net/browse/SUPP-2847

---

<a href="https://builder.io/app/projects/b29d4bc79bfe480fb3f0/proto-opening-bmzklsoi"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cdn.builder.io/api/v1/image/assets%2FYJIGb4i01jvw0SRdL5Bt%2F226fa21c49ce4f95a5aba53aa594fe7a">Edit in Builder</picture></a>&nbsp;&nbsp;<a href="https://b29d4bc79bfe480fb3f0-proto-opening-bmzklsoi.projects.builder.my/"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cdn.builder.io/api/v1/image/assets%2FYJIGb4i01jvw0SRdL5Bt%2Fe530b1333b5b4cedac9c41b8573c8268">Preview</picture></a>

---

To clone this PR locally use the [Github CLI](https://cli.github.com/) with command `gh pr checkout 4848`

You can tag me at @builderio for anything you want me to fix or change

---

> [!NOTE]
> **Low Risk**
> Narrow lifecycle fix in CustomCode with regression tests; behavior change is restoring intended client-only script execution after hydration.
> 
> **Overview**
> Fixes a regression where **Custom Code** blocks with `scriptsClientOnly` stripped `<script>` tags on SSR but never put them back or executed them on the client.
> 
> `shouldComponentUpdate` only compared `props.code`, so the one-time `setState({ hydrated: true })` in `componentDidMount` did not trigger a re-render. The change types component state as `{ hydrated: boolean }` and allows updates when `hydrated` changes, while still skipping re-renders when only unrelated parent props change.
> 
> Adds jsdom tests for SSR script stripping, post-hydration script restore/execution, and a render-count guard. Test setup polyfills `TextEncoder`/`TextDecoder` for `renderToString`. Patch changeset for `@builder.io/react`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 274496ed6dab1d76d1547bc3981d63dd75aed78b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/shaggy-moles-argue.md
+++ b/.changeset/shaggy-moles-argue.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/react': patch
+---
+
+Fix: Custom Code blocks with `scriptsClientOnly` now re-insert and run their `<script>` tags after hydration. `shouldComponentUpdate` was only comparing `props.code`, which suppressed the one-time re-render triggered by `componentDidMount`.
--- a/packages/react/src/blocks/CustomCode.tsx
+++ b/packages/react/src/blocks/CustomCode.tsx
@@ -10,6 +10,10 @@ interface Props {
   scriptsClientOnly?: boolean;
 }
 
+interface State {
+  hydrated: boolean;
+}
+
 // TODO: settings context to pass this down. do in shopify-specific generated code
 const globalReplaceNodes = ({} as { [key: string]: Node[] }) || null;
 
@@ -48,7 +52,7 @@ if (Builder.isBrowser && globalReplaceNodes) {
   }
 }
 
-class CustomCodeComponent extends React.Component<Props> {
+class CustomCodeComponent extends React.Component<Props, State> {
   elementRef: Element | null = null;
   originalRef: Node | Element | null = null;
 
@@ -57,7 +61,7 @@ class CustomCodeComponent extends React.Component<Props> {
 
   firstLoad = true;
   replaceNodes = false;
-  state = {
+  state: State = {
     hydrated: false,
   };
 
@@ -90,8 +94,9 @@ class CustomCodeComponent extends React.Component<Props> {
     }
   }
 
-  shouldComponentUpdate(nextProps: Readonly<Props>): boolean {
-    return nextProps.code !== this.props.code;
+  shouldComponentUpdate(nextProps: Readonly<Props>, nextState: State): boolean {
+    // `hydrated` flips once, in componentDidMount, to re-render the scripts stripped for SSR parity.
+    return nextProps.code !== this.props.code || nextState.hydrated !== this.state.hydrated;
   }
 
   get noReactRender() {
--- a/packages/react/test/setupTests.ts
+++ b/packages/react/test/setupTests.ts
@@ -1,3 +1,11 @@
+import { TextDecoder, TextEncoder } from 'util';
+
+// jsdom 19 ships without TextEncoder/TextDecoder, which react-dom/server needs on modern Node.
+if (typeof globalThis.TextEncoder === 'undefined') {
+  globalThis.TextEncoder = TextEncoder as typeof globalThis.TextEncoder;
+  globalThis.TextDecoder = TextDecoder as typeof globalThis.TextDecoder;
+}
+
 beforeEach(() => {
   jest.spyOn(global.Math, 'random').mockReturnValue(0.123456789);
 });
--- a/packages/react/src/blocks/custom-code.test.tsx
+++ b/packages/react/src/blocks/custom-code.test.tsx
@@ -4,20 +4,15 @@
 
 import * as React from 'react';
 import { renderToString } from 'react-dom/server';
-import { createRoot, hydrateRoot } from 'react-dom/client';
-import { act } from '@testing-library/react';
+import { act, render } from '@testing-library/react';
 import type { BuilderElement } from '@builder.io/sdk';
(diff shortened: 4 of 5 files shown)


Manish-Builder-io:
@builderio-bot Review all unresolved PR comments and latest review feedback.

Be skeptical and verify each suggestion against the code and requirements before making changes.

For every comment:

* Fix it if you agree and reply with what changed.
* If you disagree, reply with a concise rationale.
* Resolve the comment once addressed.
* Leave it open only if human input is required.

Process every comment and finish with a summary of fixed, rejected, and blocked items.

packages/react/test/setupTests.ts
sanyamkamat, line 1: why was this needed?

.changeset/shaggy-moles-argue.md
```diff
@@ -0,0 +1,5 @@
+---
```
sanyamkamat, line 1: Will do a dev release first

Manish-Builder-io: @sanyamkamat can we do the dev release?

Manish-Builder-io:
@builderio-bot Re-investigate the Custom Code `scriptsClientOnly: true` hydration regression from the ground up using the reported real Next.js App Router reproduction, and do not assume the proposed `componentDidMount` change is correct. Confirm why `state.hydrated` remains `false`, why `originalRef` is `null` in the real webpack/Next.js runtime, and how the module-scope `globalReplaceNodes` / `document.querySelectorAll(".builder-custom-code")` initialization interacts with React hydration and chunk evaluation order. Compare the real runtime behavior with the existing test and determine why the test passes while the production scenario fails. Trace the complete SSR, `hydrateRoot`, React hydration mismatch, `componentDidMount`, `setState`, re-render, and `findAndRunScripts()` flow, and identify the minimal reliable lifecycle change that ensures an SSR-stripped Custom Code block is restored and its script executes after hydration without depending on `originalRef` being captured. If the investigation confirms that `isHydrating`/`originalRef` is incorrectly gating the required state transition, implement the smallest targeted fix, preferably one that works in real bundled React/Next.js environments while preserving the intended behavior for non-hydration renders and existing `replaceNodes` logic. Add a regression test that actually exercises the problematic hydration sequence as closely as the repository can support, and strengthen the test if the current test setup masks the production issue. Validate that the fix restores the `<script>` after hydration and executes it, while ensuring existing Custom Code behavior is unchanged. Do not make unrelated refactors, and if the investigation disproves the proposed direction, report the confirmed root cause and implement only the evidence-backed fix.

packages/react/src/blocks/CustomCode.tsx
```diff
@@ -105,7 +105,11 @@ class CustomCodeComponent extends React.Component<Props, State> {
   }
 
   get isHydrating() {
-    return !isShopify && this.originalRef;
+    // `originalRef` is only captured when the SSR'd markup was already in the document when this
+    // module was evaluated. Bundlers load the SDK chunk asynchronously, so that ordering is not
+    // guaranteed. `scriptsClientOnly` always strips scripts on the server, so the first client
+    // render has to match that regardless of whether the node was captured.
+    return !isShopify && Boolean(this.originalRef || this.props.scriptsClientOnly);
```
sanyamkamat, line 112: This change kinda makes sense. I will do another dev release for this one tomorrow

Manish-Builder-io: @sanyamkamat Could you please do a dev release today for testing.

PR #4852, opened 2026-09-10
fix[smartling][utils]: ENG-13890 exclude URL, image and link fields from translation jobs

AishwaryaParab:
## Description

Image links, button URLs and page links from Builder content were all showing up as strings for Smartling to translate.

**Root Cause:**
This is a regression from #4651 (June). That PR taught the extractor to look inside a localized list or object and send each string it finds as its own translation unit, which was the right fix for text fields that were previously being skipped.

The catch is that the extractor has no idea what type a field is. So once it starts walking a payload, it picks up every string in there. `imageUrl`, `link` and `screenImage` are strings, so off they went.

We already patched one version of this in #4826, where the same walk was sending dropdown values like "White" and "Left". That fix relied on the editor recording which fields are non-translatable, which only helps for localized list fields on blocks that carry that metadata. URLs on symbols and on model fields go through different code paths and were never covered.

**Fix:**
Added one small check: a string is not translatable if it has no whitespace and starts with /, http://, https:// or cdn.builder.io/. That check now runs everywhere a string becomes a translation unit (7 places: symbol inputs, model fields, the two nested-string branches, the localized-leaf branch, and the block flat-string case).

The whitespace part matters. "Visit https://sumup.com for more" is a real sentence and still gets translated. Only a bare URL on its own gets skipped.

Skipping a field means Smartling never returns a value for it, and the SDK resolves a missing locale to undefined rather than falling back to Default, so the link would just vanish on the translated page. To prevent that, the source URL is now copied into the target locale for any field we skip. Three changes for that: the block path already had this logic behind a condition that was too narrow, so the condition was widened; the symbol and model-field paths had none, so each got a small block.

Both new blocks are gated on the same URL check, so they only ever restore a value we deliberately skipped. A field that's still waiting on a pending job is left alone, so English never leaks into a translated locale.

https://builder-io.atlassian.net/browse/ENG-13890

**Screenshot/Clip**
https://clips.agent-native.com/r/uFQLIO0vOcIM

### DIFF
--- a/packages/utils/src/translation-helpers.ts
+++ b/packages/utils/src/translation-helpers.ts
@@ -14,8 +14,8 @@ export type TranslateableFields = {
   };
 };
 
-// A bare URL or asset reference is a routing/config value, never a translation unit.
-// Whitespace means prose ("Visit https://x.com for more"), which stays translatable.
+// A bare url or asset is routing config, never copy. Whitespace means prose
+// ("Visit https://x.com for more"), which stays translatable.
 function isNonTranslatableValue(value: string) {
   const trimmed = value.trim();
   if (!trimmed || /\s/.test(trimmed)) {
@@ -24,8 +24,7 @@ function isNonTranslatableValue(value: string) {
   const lower = trimmed.toLowerCase();
   return (
     lower.startsWith('/') ||
-    // './x' and '../x' are unambiguous relative links. A bare 'foo/bar' is not: it is
-    // shape-identical to real copy like 'and/or', 'Yes/No' or '24/7', so it stays in.
+    // './x' and '../x' are clearly links. Bare 'foo/bar' is not: it looks like 'and/or'.
     lower.startsWith('./') ||
     lower.startsWith('../') ||
     lower.startsWith('http://') ||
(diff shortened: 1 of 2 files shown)


packages/utils/src/translation-helpers.ts
```diff
@@ -890,7 +924,7 @@ export function applyTranslation(
           if (!compoundKeys.length) {
             // Everything came back excluded, or there was never anything to send. Both differ
             // from the provider not having answered yet, which must leave the locale alone.
-            if (matchingKeys.length || (excludedPaths && !hasTranslatableLeaf())) {
+            if (matchingKeys.length || !hasTranslatableLeaf()) {
```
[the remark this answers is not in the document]
AishwaryaParab: Not valid.

packages/utils/src/translation-helpers.ts
```diff
@@ -14,6 +14,33 @@ export type TranslateableFields = {
   };
 };
 
+// A bare URL or asset reference is a routing/config value, never a translation unit.
+// Whitespace means prose ("Visit https://x.com for more"), which stays translatable.
+function isNonTranslatableValue(value: string) {
```
lihuelg, line 19: does this mean url is not translatable, but a text containing an url is?
Couldn't it be solved using a new URL()?

AishwaryaParab: does this mean url is not translatable, but a text containing an url is? -> that's correct
I did explore `new URL()` but SumUp also uses relative paths like `/en-gb/pricing`, `./checkout`, `cdn.builder.io/api/v1/image/x `and just `#`, so, `new URL()` doesn't really cover these

lihuelg (approved):

